PR #1, opened 2019-01-09
[WIP] Begin building command structure

juanvallejo:
Adds top-level command, and `info` command.
Adds vendored deps.

Work in progress... Opening now to begin collecting feedback.

cc @deads2k


pkg/cmd/info.go
```diff
@@ -22,16 +32,27 @@ var (
...
+	baseDir string
+	// whether or not to allow writes to an existing and populated base directory
+	overwrite bool
+
 	genericclioptions.IOStreams
 }
 
 func NewInfoOptions(streams genericclioptions.IOStreams) *InfoOptions {
 	return &InfoOptions{
+		printFlags:  genericclioptions.NewPrintFlags("gathered").WithDefaultOutput("yaml"),
```
juanvallejo: @deads2k I should probably set a `WithTypeSetter()` param here, right? This might save me the trouble of manually calling `SetGroupVersionKind()` on each resource I want to print later. Not sure what scheme to use, however.

pkg/cmd/info.go
```diff
@@ -285,3 +310,119 @@ func retrieveConfigResourceNames(discoveryClient discovery.CachedDiscoveryInterf
...
+	defer dest.Close()
+
+	if err := o.printer.PrintObj(ns, dest); err != nil {
+		return err
+	}
+
+	resourcesToStore := map[string]runtime.Object{}
+
+	// collect resource information for namespace
+	pods, err := o.kubeClient.CoreV1().Pods(namespace).List(metav1.ListOptions{})
```
juanvallejo: @deads2k I'm thinking I should probably use a dynamic client, right? That way I don't have to worry about setting GVK data for each list of resources

pkg/cmd/info.go
deads2k: > @deads2k I should probably set a `WithTypeSetter()` param here, right? This might save me the trouble of manually calling `SetGroupVersionKind()` on each resource I want to print later. Not sure what scheme to use, however.

a scheme that matches the client you plan to use.

pkg/cmd/info.go
deads2k: > @deads2k I'm thinking I should probably use a dynamic client, right? That way I don't have to worry about setting GVK data for each list of resources

there is benefit to having real types that you can navigate. You're going to finding refs here.

deads2k:
@juanvallejo get the bare minimum "produce something" and we'll merge that

PR #2, opened 2019-01-10
Implement per pod data gathering

juanvallejo:
Adds per-pod data collection such as metrics, logs, etc.
Work in progress...

cc @deads2k


juanvallejo:
@deads2k I believe the tool should now work as advertised. Only thing I am not collecting are events.
Some tests are in order as well.

juanvallejo:
TODO:

- [ ] Gather per-namespace events
- [ ] Gather operand namespace info

PR #4, opened 2019-01-11
Additional improvements

juanvallejo:
cc @deads2k


juanvallejo:
Adds /version data gathering
Adds operand namespace data gathering
Adds `operator.openshift.io` resource gathering

TODO
- [ ] switch from exec to port-forwarding to collect container /metrics, /healthz, etc.

PR #12, opened 2019-01-15
[WIP] prove plugins

deads2k:
proof that plugins can be a thing

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,2 +1,3 @@
 all:
 	mkdir -p ./bin && go build -o ./bin/openshift-must-gather main.go
+	cp plugins/cee-awesomeness/* bin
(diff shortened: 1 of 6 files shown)


deads2k:
A very rough and tumble proof that we *can* do it.  Not proof that it is the only option available.

@sferich888 @juanvallejo

sferich888:
@deads2k I wonder if we should explore doing this as a command tree? 

The default (openshift-must-gather) command with no options simply runs (all plugins/commands); however, if you supply a plugin/subcommand (version, inspect/collect <object>, etc) you can just get specific data from that plugin/subcommand (data group).

I like the ability to write plugins / in any language - to extend as we need; however that may be harder to maintain in this repo.

deads2k:
I think we're going to go with a simpler scripting approach

PR #15, opened 2019-01-15
use container port

deads2k:
always use the first container port.  People just have to work out their ordering.

/assign @juanvallejo

### DIFF
--- a/pkg/cmd/inspect.go
+++ b/pkg/cmd/inspect.go
@@ -71,6 +71,7 @@ func NewInspectOptions(streams genericclioptions.IOStreams) *InspectOptions {
 	return &InspectOptions{
 		printFlags:  genericclioptions.NewPrintFlags("gathered").WithDefaultOutput("yaml").WithTypeSetter(scheme.Scheme),
 		configFlags: genericclioptions.NewConfigFlags(),
+		overwrite:   true,
 		IOStreams:   streams,
 	}
 }
@@ -99,7 +100,6 @@ func NewCmdInspect(parentName string, streams genericclioptions.IOStreams) *cobr
 	}
 
 	cmd.Flags().StringVar(&o.baseDir, "base-dir", "must-gather", "Root directory used for storing all gathered cluster operator data. Defaults to $(PWD)/must-gather")
-	cmd.Flags().BoolVar(&o.overwrite, "overwrite", false, "If true, allow this command to write to an existing location with previous data present")
 
 	o.printFlags.AddFlags(cmd)
 	return cmd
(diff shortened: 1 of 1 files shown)


pkg/cmd/inspect.go
```diff
@@ -99,7 +100,6 @@ func NewCmdInspect(parentName string, streams genericclioptions.IOStreams) *cobr
 	}
 
 	cmd.Flags().StringVar(&o.baseDir, "base-dir", "must-gather", "Root directory used for storing all gathered cluster operator data. Defaults to $(PWD)/must-gather")
-	cmd.Flags().BoolVar(&o.overwrite, "overwrite", false, "If true, allow this command to write to an existing location with previous data present")
```
juanvallejo: do we not want this option to prevent accidental mutation of existing data?

pkg/cmd/inspect.go
```diff
@@ -599,7 +573,28 @@ func (o *InspectOptions) gatherPodData(destDir, namespace string, pod *corev1.Po
...
+	if len(container.Ports) == 0 {
+		log.Printf("        Skipping container endpoint collection for pod %q container %q: No ports\n", pod.Name, container.Name)
+		return nil
+	}
+	port := &util.RemoteContainerPort{
+		Protocol: "https",
+		Port:     container.Ports[0].ContainerPort,
+	}
+
+	if err := o.gatherContainerEndpoints(path.Join(destDir, "/"+container.Name), pod, &container, port); err != nil {
```
juanvallejo: Something I noticed that worries me slightly is that a number of containers may have `dns` port, for example, as their first port (with metrics last). Meaning as it currently stands, we'll always miss collecting metrics for those.

pkg/cmd/inspect.go
```diff
@@ -599,7 +573,28 @@ func (o *InspectOptions) gatherPodData(destDir, namespace string, pod *corev1.Po
 	return nil
 }
 
-func (o *InspectOptions) gatherContainerData(destDir string, pod *corev1.Pod, container *corev1.Container, metricsPort *util.RemoteContainerPort) error {
+func (o *InspectOptions) gatherContainerInfo(destDir string, pod *corev1.Pod, container corev1.Container) error {
+	if err := o.gatherContainerAllLogs(path.Join(destDir, "/"+container.Name), pod, &container); err != nil {
+		return err
+	}
+
+	if len(container.Ports) == 0 {
```
juanvallejo: Also, a few operator pods, for example, don't have anything under `container.Ports` as it currently stands

juanvallejo: An immediate example I can think of is the `openshift-kube-apiserver-operator` pod

PR #16, opened 2019-01-15
creating a vesion command to get clusterversion

sferich888:
@juanvallejo  can you take a look and let me know what you think.

### DIFF
--- a/pkg/cmd/version.go
+++ b/pkg/cmd/version.go
@@ -1,0 +1,209 @@
+package cmd
+
+import (
+	"fmt"
+	"io/ioutil"
+	"log"
+	"os"
+	"path"
+
+	"github.com/spf13/cobra"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/util/errors"
+	"k8s.io/cli-runtime/pkg/genericclioptions"
+	"k8s.io/cli-runtime/pkg/genericclioptions/resource"
+	"k8s.io/client-go/discovery"
+	"k8s.io/client-go/dynamic"
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/kubernetes/scheme"
+	"k8s.io/client-go/rest"
+
+	configv1 "github.com/openshift/api/config/v1"
+	"github.com/openshift/must-gather/pkg/util"
+)
+
+var (
+	versionExample = `
+	# Collect debugging data for the "openshift-apiserver-operator"
+	%[1]s version
+`
+)
+
+type VersionOptions struct {
+	printFlags  *genericclioptions.PrintFlags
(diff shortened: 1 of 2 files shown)


juanvallejo:
cc @deads2k

pkg/cmd/version.go
```diff
@@ -0,0 +1,209 @@
...
+	if len(o.baseDir) == 0 {
+		return fmt.Errorf("--base-dir must not be empty")
+	}
+	return nil
+}
+
+func (o *VersionOptions) Run() error {
+	// next, ensure we're able to proceed writing data to specified destination
+	if err := o.ensureDirectoryViable(o.baseDir, o.overwrite); err != nil {
+		log.Printf("Failed to setup %q", o.baseDir)
```
juanvallejo: I think we can depend on the error message here. Logging seems redundant

pkg/cmd/version.go
```diff
@@ -0,0 +1,209 @@
...
+			if err := o.Run(); err != nil {
+				return err
+			}
+
+			return nil
+		},
+	}
+
+	cmd.Flags().StringVar(&o.baseDir, "base-dir", "must-gather", "Root directory used for storing all gathered cluster operator data. Defaults to $(PWD)/must-gather")
+	cmd.Flags().BoolVar(&o.overwrite, "overwrite", false, "If true, allow this command to write to an existing location with previous data present")
```
juanvallejo: You can safely remove this flag. Also, default `o.overwrite` to true under `NewVersionOptions`

pkg/cmd/version.go
```diff
@@ -0,0 +1,209 @@
...
+		return err
+	}
+
+	o.builder = resource.NewBuilder(o.configFlags)
+	return nil
+}
+
+func (o *VersionOptions) Validate() error {
+	if len(o.args) >= 1 {
+		return fmt.Errorf("this option takes not arguments")
```
juanvallejo: `len(o.args) > 0` reads a bit better imo. Also, how about rephrasing to `this command does not take arguments`

pkg/cmd/version.go
```diff
@@ -0,0 +1,209 @@
...
+	clusterVersion, err := o.dynamicClient.Resource(configv1.SchemeGroupVersion.WithResource("clusterversions")).List(metav1.ListOptions{})
+	if err != nil {
+		errs = append(errs, err)
+	}
+
+	objToPrint := runtime.Object(clusterVersion)
+
+	filename := fmt.Sprintf("%s.yaml", "clusterversion")
+
+	if err := o.fileWriter.WriteFromResource(path.Join(o.baseDir, "/"+filename), objToPrint); err != nil {
```
juanvallejo: maybe consider putting these under `must-gather/resources/config.openshift.io/clusterversions.yaml`

juanvallejo: This is the current location where we place them with the `inspect` command.

sferich888:
https://github.com/openshift/must-gather/pull/22 already collects this.

PR #19, opened 2019-01-16
iterate and save non-namespace related objs

juanvallejo:
@deads2k does this patch make sense? If so, we'll also need to store a resource version as a new field in configv1.ObjectReference

### DIFF
--- a/pkg/cmd/inspect.go
+++ b/pkg/cmd/inspect.go
@@ -201,13 +201,16 @@ func (o *InspectOptions) Run() error {
 			continue
 		}
 
-		namespaces, err := obtainClusterOperatorNamespaces(info.Object)
+		namespaces, otherObjs, err := obtainClusterOperatorRelatedObjects(info.Object)
 		if err != nil {
 			errs = append(errs, err)
 			continue
 		}
 
 		// save operator data for each clusteroperator namespace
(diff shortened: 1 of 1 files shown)


pkg/cmd/inspect.go
```diff
@@ -219,6 +222,19 @@ func (o *InspectOptions) Run() error {
...
+		if len(otherObjs) == 0 {
+			log.Printf("unable to find any non-namespace related objects to clusteroperator/%s. Will not perform further data collection...\n", info.Name)
+		}
+		//for _, relatedObj := range otherObjs {
+		//	if len(relatedObj.Namespace) == 0 {
+		//		o.gatherClusterOperatorRelatedObjectData(path.Join(o.baseDir, "/cluster-scoped-resources", "/"+relatedObj.Group, "/"+relatedObj.Resource), relatedObj)
+		//		continue
+		//	}
+		//
+		//	o.gatherClusterOperatorRelatedObjectData(path.Join(o.baseDir, "/namespaces", "/"+relatedObj.Namespace, "/"+relatedObj.Resource), relatedObj)
```
juanvallejo: we'll need to store Version for every configv1.ObjectReference in order to fetch and store the resource using a dynamic client (if this patch makes sense)

PR #22, opened 2019-01-18
restructure inspect to handle polymorphic resources

juanvallejo:
cc @deads2k

### DIFF
--- a/cmd/must-gather.go
+++ b/cmd/must-gather.go
@@ -8,19 +8,16 @@ import (
 
 	"k8s.io/cli-runtime/pkg/genericclioptions"
 
-	mustgather "github.com/openshift/must-gather/pkg/cmd"
+	"github.com/openshift/must-gather/pkg/cmd/inspect"
 )
 
 type MustGatherOptions struct {
-	configFlags *genericclioptions.ConfigFlags
-
 	genericclioptions.IOStreams
 }
 
 func NewMustGatherOptions(streams genericclioptions.IOStreams) *MustGatherOptions {
 	return &MustGatherOptions{
-		configFlags: genericclioptions.NewConfigFlags(),
-		IOStreams:   streams,
+		IOStreams: streams,
 	}
 }
 
@@ -46,7 +43,7 @@ func NewCmdMustGather(streams genericclioptions.IOStreams) *cobra.Command {
 		},
 	}
 
-	cmd.AddCommand(mustgather.NewCmdInspect("openshift-must-gather", streams))
+	cmd.AddCommand(inspect.NewCmdInspect("openshift-must-gather", streams))
 	return cmd
 }
 
--- a/cmd/must-gather.go
+++ b/cmd/must-gather.go
@@ -43,7 +43,7 @@ func NewCmdMustGather(streams genericclioptions.IOStreams) *cobra.Command {
 		},
 	}
 
-	cmd.AddCommand(inspect.NewCmdInspect("openshift-must-gather", streams))
+	cmd.AddCommand(inspect.NewCmdInspect(streams))
 	return cmd
 }
 
--- a/pkg/cmd/inspect/inspect.go
+++ b/pkg/cmd/inspect/inspect.go
@@ -1,8 +1,6 @@
 package inspect
 
 import (
-	"bytes"
-	"encoding/json"
 	"fmt"
 	"io/ioutil"
 	"log"
@@ -12,11 +10,7 @@ import (
 
 	"github.com/spf13/cobra"
 
-	appsv1 "k8s.io/api/apps/v1"
-	corev1 "k8s.io/api/core/v1"
-	kapierrs "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/apimachinery/pkg/util/errors"
@@ -30,7 +24,6 @@ import (
 	"k8s.io/client-go/rest"
 
 	configv1 "github.com/openshift/api/config/v1"
-	"github.com/openshift/api/route"
 
 	"github.com/openshift/must-gather/pkg/util"
 )
@@ -77,13 +70,13 @@ func NewInspectOptions(streams genericclioptions.IOStreams) *InspectOptions {
 	}
 }
 
-func NewCmdInspect(parentName string, streams genericclioptions.IOStreams) *cobra.Command {
+func NewCmdInspect(streams genericclioptions.IOStreams) *cobra.Command {
 	o := NewInspectOptions(streams)
 
 	cmd := &cobra.Command{
 		Use:          "inspect <operator> [flags]",
 		Short:        "Collect debugging data for a given cluster operator",
-		Example:      fmt.Sprintf(inspectExample, parentName),
+		Example:      fmt.Sprintf(inspectExample, os.Args[0]),
 		SilenceUsage: true,
 		RunE: func(c *cobra.Command, args []string) error {
 			if err := o.Complete(c, args); err != nil {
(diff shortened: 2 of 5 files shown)


pkg/cmd/inspect.go
```diff
@@ -155,6 +155,93 @@ func (o *InspectOptions) Validate() error {
 	return nil
 }
 
+func objectReferenceToResourceInfo(clientGetter genericclioptions.RESTClientGetter, ref *configv1.ObjectReference) (*resource.Info, error) {
```
deads2k: I think it's time to make packages for commands.

pkg/cmd/inspect.go
```diff
@@ -155,6 +155,93 @@ func (o *InspectOptions) Validate() error {
...
+	infos, err := b.Do().Infos()
+	if err != nil {
+		return nil, err
+	}
+
+	return infos[0], nil
+}
+
+func infoToContextKey(info *resource.Info) string {
+	return fmt.Sprintf("%s/%s.%s/%s/%s", info.Namespace, info.ResourceMapping().GroupVersionKind.Group, info.ResourceVersion, info.ResourceMapping().Resource.Resource, info.Name)
```
deads2k: no resourceversion in this key

pkg/cmd/inspect.go
```diff
@@ -155,6 +155,93 @@ func (o *InspectOptions) Validate() error {
...
+
+	infos, err := b.Do().Infos()
+	if err != nil {
+		return nil, err
+	}
+
+	return infos[0], nil
+}
+
+func infoToContextKey(info *resource.Info) string {
```
deads2k: doc the meaning of this key

pkg/cmd/inspect.go
```diff
@@ -155,6 +155,93 @@ func (o *InspectOptions) Validate() error {
...
+	return fmt.Sprintf("%s/%s.%s/%s/%s", info.Namespace, info.ResourceMapping().GroupVersionKind.Group, info.ResourceVersion, info.ResourceMapping().Resource.Resource, info.Name)
+}
+
+// collectPolymorphicResourceInfo receives an object to gather debugging data for, and a context to keep track of
+// already-seen objects when following related-object reference chains.
+func (o *InspectOptions) collectPolymorphicResourceInfo(info *resource.Info, context map[string]bool) ([]string, error) {
```
deads2k: make context a struct that contains a "visited" string set

deads2k: `InspectResource`

pkg/cmd/inspect.go
```diff
@@ -155,6 +155,93 @@ func (o *InspectOptions) Validate() error {
...
+	context[infoToContextKey(info)] = true
+
+	unstr, ok := info.Object.(*unstructured.Unstructured)
+	if !ok {
+		return nil, fmt.Errorf("unexpected type. Expecting %q but got %T", "*unstructured.Unstructured", info.Object)
+	}
+
+	skippedNamespaces := []string{}
+
+	switch info.ResourceMapping().Resource.Resource {
```
deads2k: no, switch on teh groupresource

deads2k: switch on the actual info.Object()'s groupresource

pkg/cmd/inspect.go
```diff
@@ -155,6 +155,93 @@ func (o *InspectOptions) Validate() error {
...
+	case "clusteroperators":
+		// save clusteroperator resources to disk
+		if err := o.gatherClusterOperatorResource(path.Join(o.baseDir, "/clusteroperator"), unstr); err != nil {
```
deads2k: `base/cluster-scoped/config.openshift.io(group)/clusteroperators(resource)/<name>.yaml` for individuals and `base/cluster-scoped/config.openshift.io/clusteroperators(resource).yaml` for lists I think

deads2k: move the logic from this case into it's own function I think

pkg/cmd/inspect.go
deads2k: structure like polymorphic subresources where it's a bare method not attached to the options struct

pkg/cmd/inspect.go
deads2k: have a variant `objectRefToContextKey` so you don't have to retrieve things you've already seen.

pkg/cmd/inspect.go
```diff
@@ -155,6 +155,93 @@ func (o *InspectOptions) Validate() error {
...
+
+		// obtain associated objects for the current clusteroperator resources
+		relatedObjReferences, err := obtainClusterOperatorRelatedObjects(unstr)
+		if err != nil {
+			return nil, err
+		}
+
+		errs := []error{}
+		for _, relatedRef := range relatedObjReferences {
+			relatedInfo, err := objectReferenceToResourceInfo(o.configFlags, relatedRef)
```
deads2k: intercept already visited items here too

pkg/cmd/inspect.go
```diff
@@ -155,6 +155,93 @@ func (o *InspectOptions) Validate() error {
...
+		}
+
+		if len(errs) > 0 {
+			return skippedNamespaces, errors.NewAggregate(errs)
+		}
+	case "namespaces":
+		if err := o.gatherNamespaceData(path.Join(o.baseDir, "namespaces", info.Name), info.Name); err != nil {
+			if kapierrs.IsNotFound(err) {
+				skippedNamespaces = append(skippedNamespaces, info.Name)
+				return skippedNamespaces, err
```
deads2k: why?  We've already visited it, so there's no reason for a special case, right?

juanvallejo: This is for cases where the namespace is not found at all, so that we can list all of them (if any) at the end

juanvallejo:
@deads2k thanks, comments addressed

pkg/cmd/inspect/inspect.go
```diff
@@ -167,102 +167,134 @@ func (o *InspectOptions) Run() error {
...
 	errs := []error{}
 	if err := o.gatherConfigResourceData(path.Join(o.baseDir, "/cluster-scoped-resources/config.openshift.io")); err != nil {
 		errs = append(errs, err)
 	}
 
-	// gather operator.openshift.io resource data
+	// then, gather operator.openshift.io resource data
 	if err := o.gatherOperatorResourceData(path.Join(o.baseDir, "/cluster-scoped-resources/operator.openshift.io")); err != nil {
 		errs = append(errs, err)
 	}
```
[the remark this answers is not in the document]
davemulford: That makes sense to me. Having _sane_ defaults is always a good idea.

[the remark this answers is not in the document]
juanvallejo: Agree, cc @deads2k

pkg/cmd/inspect/inspect.go
```diff
@@ -167,102 +167,134 @@ func (o *InspectOptions) Run() error {
...
 		}
+	}
 
-		namespaces, err := obtainClusterOperatorNamespaces(info.Object)
-		if err != nil {
-			errs = append(errs, err)
-			continue
+	if len(skippedNamespaces) > 0 {
+		for _, namespace := range skippedNamespaces {
+			log.Printf("Data collection skipped namespace %q. Unable to find namespace...\n", namespace)
```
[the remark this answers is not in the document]
juanvallejo: yes, makes sense

cmd/must-gather.go
```diff
@@ -46,7 +43,7 @@ func NewCmdMustGather(streams genericclioptions.IOStreams) *cobra.Command {
 		},
 	}
 
-	cmd.AddCommand(mustgather.NewCmdInspect("openshift-must-gather", streams))
+	cmd.AddCommand(inspect.NewCmdInspect("openshift-must-gather", streams))
```
deads2k: nit: passing a parent doesn't add us much value, does it?

pkg/cmd/inspect/inspect.go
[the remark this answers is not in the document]
deads2k: > Agree, cc @deads2k

I'm ok always adding those when collecting anything.

pkg/cmd/inspect/inspect.go
```diff
@@ -167,102 +167,138 @@ func (o *InspectOptions) Run() error {
...
+		return errors.NewAggregate(allErrs)
+	}
+
+	log.Printf("Finished successfully with no errors.\n")
+	return nil
+}
+
+// InspectResource receives an object to gather debugging data for, and a context to keep track of
+// already-seen objects when following related-object reference chains.
+func InspectResource(info *resource.Info, context *resourceContext, o *InspectOptions) ([]string, error) {
```
deads2k: separate file please

pkg/cmd/inspect/inspect.go
```diff
@@ -167,102 +167,138 @@ func (o *InspectOptions) Run() error {
...
+	if !ok {
+		return nil, fmt.Errorf("unexpected type. Expecting %q but got %T", "*unstructured.Unstructured", info.Object)
+	}
+
+	notFoundNamespaces := []string{}
+
+	switch info.ResourceMapping().Resource.GroupResource() {
+	case configv1.GroupVersion.WithResource("clusteroperators").GroupResource():
+		// save clusteroperator resources to disk
+		if err := o.gatherClusterOperatorResource(path.Join(o.baseDir, "/cluster-scoped-resources", "/"+unstr.GroupVersionKind().Group, "/clusteroperators"), unstr); err != nil {
```
deads2k: make this idiot proof.  Only pass the basedir, the rest is computed to always be the same.  Constant for `cluster-scoped-resources` please

pkg/cmd/inspect/inspect.go
```diff
@@ -167,102 +167,138 @@ func (o *InspectOptions) Run() error {
...
-	}
 
-	if len(skippedNamespaces) > 0 {
-		for _, namespace := range skippedNamespaces {
-			log.Printf("Data collection skipped namespace %q. Unable to find namespace...\n", namespace)
+		if len(errs) > 0 {
+			return notFoundNamespaces, errors.NewAggregate(errs)
+		}
+	case corev1.SchemeGroupVersion.WithResource("namespaces").GroupResource():
+		if err := o.gatherNamespaceData(path.Join(o.baseDir, "namespaces", info.Name), info.Name); err != nil {
```
deads2k: same thing about paths here.  Also, namespaces are cluster scoped for this purpose. neat, right?

Make a special group folder called `core` for `""`

pkg/cmd/inspect/inspect.go
```diff
@@ -167,102 +167,138 @@ func (o *InspectOptions) Run() error {
...
+			if kapierrs.IsNotFound(err) {
+				notFoundNamespaces = append(notFoundNamespaces, info.Name)
+				return notFoundNamespaces, err
+			}
+		}
+	default:
+		// save the current object to disk
+		filename := fmt.Sprintf("%s.yaml", unstr.GetName())
+		objPath := path.Join(o.baseDir, "/namespaces", "/"+unstr.GetNamespace(), "/"+info.ResourceMapping().Resource.Resource)
```
deads2k: choose paths based on the presence of namespace

deads2k:
let's go for it

PR #24, opened 2019-01-18
Manage deps through glide

juanvallejo:
cc @deads2k @soltysh


glide.yaml
```diff
@@ -0,0 +1,26 @@
+package: github.com/openshift/must-gather
+import:
+- package: github.com/spf13/cobra
```
soltysh: These should come from k8s

glide.yaml
```diff
@@ -0,0 +1,26 @@
...
+  repo:    git@github.com:spf13/cobra
+  version: 7547e83b2d85fd1893c7d76916f67689d761fecb
+- package: github.com/spf13/pflag
+  repo:    git@github.com:spf13/pflag
+  version: 24fa6976df40757dce6aea913e7b81ade90530e1
+- package: k8s.io/client-go
```
soltysh: pin to a version similarly to what we do in origin, iow. `kubernetes-1.11.1` for now, or you can even go with `kubernetes-1.12.4` which is what we'll get with the rebase

glide.yaml
```diff
@@ -0,0 +1,26 @@
...
+- package: k8s.io/cli-runtime
+  repo:    https://github.com/kubernetes/cli-runtime
+  version: 5702a69c455d3a129efbc2e86f8de61eb1af25a5
+- package: k8s.io/api
+  repo:    https://github.com/kubernetes/api
+  version: 173ce66c1e39d1d0f56e0b3347ff2988068aecd0
+- package: k8s.io/apimachinery
+  repo:    https://github.com/kubernetes/apimachinery
+  version: 001e837a013905514689c0e17c7065db5638974e
+- package: k8s.io/openshift/api
```
soltysh: For now openshift we always set those to `master`

glide.yaml
```diff
@@ -0,0 +1,26 @@
...
+  repo:    https://github.com/openshift/api
+  version: c3ef7090a436a4ce1334e8a0401083b072dc7ab8
+- package: k8s.io/xeonx/timeago
```
soltysh: This is weird, where did you get it from? Why the repo is different than package?

Makefile
```diff
@@ -2,3 +2,6 @@ all:
 	mkdir -p ./bin && \
 		go build -o bin/openshift-must-gather cmd/must-gather.go && \
 		go build -o bin/openshift-dev-helpers dev-helpers/cmd/dev-helpers.go
+
+update-deps:
```
soltysh: This will eventually have to go as soon as we switch to @tnozicka's build system

glide.yaml
juanvallejo: oops, typo on my part. We use this in the utils/events.go file to format event times

Makefile
juanvallejo: yep, agree

juanvallejo:
@soltysh thanks for the review, comments addressed

PR #27, opened 2019-01-22
events: add option to turn on absolute time

mfojtik:


pkg/cmd/events.go
```diff
@@ -30,6 +30,7 @@ type EventsOptions struct {
 
 	componentName  string
 	listComponents bool
+	absoluteTime   bool
```
mfojtik: @soltysh is there some standard option for this in kubectl?

PR #28, opened 2019-01-22
dev-helper: fix logical bug in audit verb parsing

mfojtik:


mfojtik:
just logical bug, self-lgtm.

PR #32, opened 2019-01-25
Update gitignore and setup OWNERS

soltysh:
/assign @juanvallejo


OWNERS
```diff
@@ -0,0 +1,13 @@
+reviewers:
+  - deads2k
+  - soltysh
+  - tnozicka
+  - mfojtik
+  - juanvallejo
```
sferich888: Should I, @RickJWagner and others from CEE be on this list?

soltysh: We'll get you there, for now this list is sufficient.

juanvallejo:
hm, few ci failures, but I doubt any are related to this pull :)

.gitignore
```diff
@@ -12,3 +12,6 @@
 *.out
 
 bin/*
```
sferich888: We don't create this directory anymore (we can remove this).

soltysh: Done.

PR #33, opened 2019-01-25
Gssdefaultsflag

RickJWagner:
Add flag to indicate Red Hat Support suggested defaults to help filter the amount of data gathered.

### DIFF
--- a/pkg/cmd/inspect/inspect.go
+++ b/pkg/cmd/inspect/inspect.go
@@ -6,6 +6,7 @@ import (
 	"log"
 	"os"
 	"path"
+	"strconv"
 	"strings"
 
 	"github.com/spf13/cobra"
@@ -57,6 +58,8 @@ type InspectOptions struct {
 	baseDir string
 	// whether or not to allow writes to an existing and populated base directory
 	overwrite bool
+	// flag to indicate GSS defaults should be used in deciding what to gather
+	gssdefaults bool
 
 	genericclioptions.IOStreams
 }
@@ -66,6 +69,7 @@ func NewInspectOptions(streams genericclioptions.IOStreams) *InspectOptions {
 		printFlags:  genericclioptions.NewPrintFlags("gathered").WithDefaultOutput("yaml").WithTypeSetter(scheme.Scheme),
 		configFlags: genericclioptions.NewConfigFlags(),
 		overwrite:   true,
+		gssdefaults: false,
 		IOStreams:   streams,
 	}
 }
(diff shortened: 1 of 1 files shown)


sferich888:
@RickJWagner currently this code has no logic (it simply adds the flag), 

What items will we want to capture by default? 

- clusteroperators
- machine***

RickJWagner:
Hi @sferich888,
You're right, this first commit is just to put the flag in place.  Future commits will add the business logic to gather information helpful on middleware support cases.

Some of the things we'll be going after (especially things like heap dumps, thread dumps) might require some pretty fine-grained manipulation on the target machine.  These probably won't be quick to implement.  (Maybe these shouldn't be in the 'defaults', though.  Feel free to let me know.)  I suppose something like application server logs are closer to the true 'defaults' we're after.

So....   I thought I'd try to sneak in the flag as a first commit.  It's almost certain we'll use it, no matter what the resource retrieval implementation looks like.  And since the codebase is pretty fast moving, I figured it might be helpful-- less stuff that might have to be rebased if I take too long building the rest of the logic.

But I'm fine holding off to go for the 'all in one fell swoop' commit, if that's best. 

Thanks,
Rick

sferich888:
> Some of the things we'll be going after (especially things like heap dumps, thread dumps) might require some pretty fine-grained manipulation on the target machine. These probably won't be quick to implement. (Maybe these shouldn't be in the 'defaults', though. Feel free to let me know.) I suppose something like application server logs are closer to the true 'defaults' we're after.

Yea I don't think we want to do these types of things as a default, I believe the goal of the defaults is to do things like:

- https://access.redhat.com/solutions/3799561
- Collect a lists of nodes, machines, machinesets, machineconfigs
    - https://github.com/derekwaynecarr/openshift-the-easy-way/blob/master/docs/compute/01-managing-compute.md
- tell us if an auto-scaler is deployed
    - `oc get clusterautoscaler.autoscaling.openshift.io`

> So.... I thought I'd try to sneak in the flag as a first commit. It's almost certain we'll use it, no matter what the resource retrieval implementation looks like. 

That's fine, I just think we need to start defining what the default objects are.

juanvallejo:
Once we have a list of default resources, we could slightly tweak the resource builder so that we can just pass a slice of `resourceKind/resourceName` strings to it

juanvallejo:
Would also allow users to specify multiple resources via more than one arg in the command line

juanvallejo:
> Exactly, where/how do we do that?

https://github.com/openshift/must-gather/pull/34

PR #35, opened 2019-01-29
limit config/operator.openshift.io data gathering

juanvallejo:
Prevents config.openshift.io and operator.openshift.io data
collection, unless explicitly dealing with a clusteroperator resource

cc @sferich888 @deads2k

### DIFF
--- a/pkg/cmd/inspect/resource.go
+++ b/pkg/cmd/inspect/resource.go
@@ -27,6 +27,9 @@ import (
 const (
 	clusterScopedResourcesDirname = "cluster-scoped-resources"
 	namespaceResourcesDirname     = "namespaces"
+
+	configResourceDataKey   = "/cluster-scoped-resources/config.openshift.io"
+	operatorResourceDataKey = "/cluster-scoped-resources/operator.openshift.io"
 )
 
 // InspectResource receives an object to gather debugging data for, and a context to keep track of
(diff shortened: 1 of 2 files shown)


pkg/cmd/inspect/resource.go
```diff
@@ -27,6 +27,9 @@ import (
 const (
 	clusterScopedResourcesDirname = "cluster-scoped-resources"
 	namespaceResourcesDirname     = "namespaces"
+
+	configResourceDataKey   = "/cluster-scoped-resources/config.openshift.io"
+	operatorResourceDataKey = "/cluster-scoped-resources/operator.openshift.io"
```
sferich888: Should we not string build here? 
This will keep clusterScopedResourcesDirname from getting out of sync with the DataKey's

pkg/cmd/inspect/resource.go
```diff
@@ -45,6 +48,17 @@ func InspectResource(info *resource.Info, context *resourceContext, o *InspectOp
 
 	switch info.ResourceMapping().Resource.GroupResource() {
 	case configv1.GroupVersion.WithResource("clusteroperators").GroupResource():
+		// first, gather config.openshift.io resource data
+		errs := []error{}
+		if err := o.gatherConfigResourceData(path.Join(o.baseDir, "/cluster-scoped-resources/config.openshift.io"), context); err != nil {
+			errs = append(errs, err)
+		}
```
sferich888: Will we possibly want this collected outside of clusteroperators? 
What other top-level objects do we have a need to collect config.openshift.io for? 

If we add this in with other major objects, will it be skipped because of `ctx.visited.Has`?

pkg/cmd/inspect/resource.go
juanvallejo: I don't think we mind if they get out of sync. Primary purpose of the key is to create a unique identifier to keep track of what we've collected in the resourceContext

pkg/cmd/inspect/resource.go
juanvallejo: We could collect it for namespaces as well, if that makes sense.

> If we add this in with other major objects, will it be skipped because of ctx.visited.Has?

It will always be collected once per *resourceContext.

juanvallejo: No, with this patch, we are only collecting config.openshift.io and operator.openshift.io when a user specifies a `clusteroperator` among the resource/name pairs they wish to collect via the cli.

If we also wish to gather config.openshift.io and operator.openshift.io data when, say, only a namespace is specified by the user, we can copy this block into the `namespaces` switch case as well

PR #36, opened 2019-01-29
Creating a default input when --default is supplied

sferich888:
Completes https://github.com/openshift/must-gather/pull/33 with minor changes

### DIFF
--- a/pkg/cmd/inspect/inspect.go
+++ b/pkg/cmd/inspect/inspect.go
@@ -38,6 +38,12 @@ var (
 `
 )
 
+var (
+	defaults = []string{
+		"clusteroperator",
+	}
+)
+
 type InspectOptions struct {
 	printFlags  *genericclioptions.PrintFlags
 	configFlags *genericclioptions.ConfigFlags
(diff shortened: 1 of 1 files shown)


pkg/cmd/inspect/inspect.go
```diff
@@ -38,6 +38,12 @@ var (
 `
 )
 
+var (
+	defaults = []string{
```
juanvallejo: nit: move up with `inspectExample`

pkg/cmd/inspect/inspect.go
```diff
@@ -105,6 +115,10 @@ func NewCmdInspect(streams genericclioptions.IOStreams) *cobra.Command {
 func (o *InspectOptions) Complete(cmd *cobra.Command, args []string) error {
 	o.args = args
 
+	if o.defaults == true {
```
juanvallejo: nit: `== true` is redundant here

pkg/cmd/inspect/inspect.go
```diff
@@ -105,6 +115,10 @@ func NewCmdInspect(streams genericclioptions.IOStreams) *cobra.Command {
...
+		o.args = append(o.args, defaults...)
```
juanvallejo: I guess we should not worry much about `--defaults` adding duplicate resources since we keep track of collected data via the *resourceContext. However this does mean that if a user does something like `must-gather clusteroperators --defaults`, we'd be fetching clusteroperators twice

PR #38, opened 2019-01-30
adding deployment configs to default collection of namespace objects

sferich888:
Also modifying how errors are returned so that we can report on, what resources we are not gathering.

### DIFF
--- a/pkg/cmd/inspect/resource.go
+++ b/pkg/cmd/inspect/resource.go
@@ -19,6 +19,7 @@ import (
 	"k8s.io/client-go/rest"
 
 	configv1 "github.com/openshift/api/config/v1"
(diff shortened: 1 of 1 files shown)


pkg/cmd/inspect/resource.go
```diff
@@ -151,36 +152,38 @@ func (o *InspectOptions) gatherNamespaceData(baseDir, namespace string) error {
 	}
 
 	ns, err := o.kubeClient.CoreV1().Namespaces().Get(namespace, metav1.GetOptions{})
-	if err != nil {
+	if err != nil {              // If we can't get the namespace we need to exit out
```
juanvallejo: is this here comment necessary?

PR #40, opened 2019-01-30
adding addition default collections on namespaces

sferich888:
and defaults for --defaults

### DIFF
--- a/pkg/cmd/inspect/inspect.go
+++ b/pkg/cmd/inspect/inspect.go
@@ -41,6 +41,8 @@ var (
 var (
 	defaults = []string{
 		"clusteroperator",
+		"nodes",
+		"persistentvolumes",
 	}
 )
 
--- a/pkg/cmd/inspect/resource.go
+++ b/pkg/cmd/inspect/resource.go
@@ -20,6 +20,7 @@ import (
 
 	configv1 "github.com/openshift/api/config/v1"
 	ocpappsv1 "github.com/openshift/api/apps/v1"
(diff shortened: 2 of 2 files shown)


pkg/cmd/inspect/inspect.go
```diff
@@ -41,6 +41,8 @@ var (
 var (
 	defaults = []string{
 		"clusteroperator",
+		"nodes",
+		"persistentvolumes",
```
sferich888: I don't think this will work due to how args are read currently.

juanvallejo: it'll work as long as any args provided by the user aren't resource/name pairs

juanvallejo: cc @deads2k ^

sferich888: We can't assume that no one does`mg inspect --defaults namespace abc` well have to find a way to mitigate this.

juanvallejo: We could think about making  arguments and `--defaults` mutually exclusive

sferich888:
Wasn't that the intent, to begin with?

pkg/cmd/inspect/inspect.go
davemulford: Why would both not work?  Seems like we should be able to gather both the defaults, and whatever else the user adds on?

deads2k:
we decided against this direction.

PR #41, opened 2019-01-31
use resource builder to gather all namespace data

juanvallejo:
cc @deads2k @sferich888

### DIFF
--- a/pkg/cmd/inspect/resource.go
+++ b/pkg/cmd/inspect/resource.go
@@ -9,7 +9,6 @@ import (
 	"path"
 	"strings"
 
-	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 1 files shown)


pkg/cmd/inspect/resource.go
```diff
@@ -166,27 +164,42 @@ func (o *InspectOptions) gatherNamespaceData(baseDir, namespace string) error {
 
 	log.Printf("    Collecting resources for namespace %q...\n", namespace)
 
+	// "all" maps to every resource we wish to collect for the given namespace,
+	// except for "events", "configmaps" and "secrets" so we'll gather those manually.
+	b := resource.NewBuilder(o.configFlags).
+		Unstructured().
+		ResourceTypeOrNameArgs(true, "all").
```
deads2k: try `.All`

deads2k: > try `.All`

oh, nm. that's cool.  does that work?

juanvallejo: it works, collects all resources we expect from a namespace + bc, dc, and routes. (We skip routes down below to collect and redact their info manually)

PR #42, opened 2019-01-31
Fix gofmt error

nak3:
This patch fixes gofmt error.

Please refer to https://github.com/openshift/must-gather/pull/38. It was merged without passing `make verify`.


soltysh:
This was picked in #32.

PR #49, opened 2019-02-06
Add additional ns info.  

RickJWagner:
Add additional ns info.  Formatted so gofmt should pass.

### DIFF
--- a/pkg/cmd/inspect/resource.go
+++ b/pkg/cmd/inspect/resource.go
@@ -179,8 +179,11 @@ func (o *InspectOptions) gatherNamespaceData(baseDir, namespace string) error {
 	}
 
 	resourcesTypesToStore := map[schema.GroupVersionResource]bool{
-		corev1.SchemeGroupVersion.WithResource("events"):     true,
-		corev1.SchemeGroupVersion.WithResource("configmaps"): true,
+		corev1.SchemeGroupVersion.WithResource("events"):          true,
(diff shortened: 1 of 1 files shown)


pkg/cmd/inspect/resource.go
```diff
@@ -179,8 +179,11 @@ func (o *InspectOptions) gatherNamespaceData(baseDir, namespace string) error {
 	}
 
 	resourcesTypesToStore := map[schema.GroupVersionResource]bool{
-		corev1.SchemeGroupVersion.WithResource("events"):     true,
-		corev1.SchemeGroupVersion.WithResource("configmaps"): true,
+		corev1.SchemeGroupVersion.WithResource("events"):          true,
+		corev1.SchemeGroupVersion.WithResource("configmaps"):      true,
+		corev1.SchemeGroupVersion.WithResource("endpoints"):       true,
+		corev1.SchemeGroupVersion.WithResource("podtemplates"):    true,
```
soltysh: These are not actual resources, you might want to have just pods.

deads2k:
Have a look to see what isn't already covered by `all` and you add it here https://github.com/openshift/must-gather/blob/master/pkg/cmd/inspect/namespace.go#L18

RickJWagner:
This PR is no longer relevant.   ResourcesTypesToStore now lives in namespace.go and contains only 'pods', as @soltysh suggested in February.

PR #53, opened 2019-02-07
remove defaults since we have a truly polymorphic command

deads2k:
Given a truly polymorphic, re-entrant helper, there isn't a  need for defaults to be set.  Instead of wiring something like this, users should just indicate all the things they want to collect.

/assign @juanvallejo @soltysh


sferich888:
No, we need a way (this is what defaults does) to collect a large set if things (easy button). 

Storing this in code is better than in a KCSnor document in some way.

PR #54, opened 2019-02-07
create a no-arg command for collecting information related to a control-plane

deads2k:
`openshift-must-gather inspect-control-plane`

@sferich888 @juanvallejo  as an example.

### DIFF
--- a/cmd/openshift-must-gather/main.go
+++ b/cmd/openshift-must-gather/main.go
@@ -3,6 +3,8 @@ package main
 import (
 	"os"
 
+	"github.com/openshift/must-gather/pkg/cmd/inspectcontrolplane"
+
 	"github.com/spf13/cobra"
 	"github.com/spf13/pflag"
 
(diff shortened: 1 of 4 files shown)


soltysh:
> openshift-must-gather inspect-control-plane

I was wondering about it, I'd like to have a few flavors, but I'm not sure we should be writing command for each. So possible solutions might be:
1. `openshift-must-gather inspect <what>` and we'd have a series of sub-commands for every component.
2. `openshift-must-gather inspect --component <what>` and we'd have flag for every component. 
Wdyt?

deads2k:
> 2\. `openshift-must-gather inspect --component <what>` and we'd have flag for every component.

Not this.  I think this command should stay neutral.  

Consider `oc gather-my-cluster` versus `oc gather this/resource`.  The former is strongly opinionated and will result in anger from people who want one thing more or one thing less or in this use-case do something different or handle 403s for these resources specially.  The latter is internally consistent about its behavior and surprisingly hard to have opinions about.  I think that `inspect` is the latter case


/hold

holding while we consider our options.

deads2k:
This proved a possibility, but until we have a need, I think we'll go with a folder of bash scripts.

PR #57, opened 2019-02-07
obtain a serviceaccount token to avoid authorization errors when fetching pod endpoint data

juanvallejo:
cc @deads2k @soltysh 

I too was running into issues in my cluster where my KUBECONFIG cert could not be used to retrieve endpoints served by pods.

Obtaining a token from a serviceaccount with enough permissions was one way I found to mitigate this problem. Open to suggestions.


soltysh:
/close
working on an alternative solution

PR #59, opened 2019-02-12
Update examples in inspect command & gitignore

soltysh:


soltysh:
Simple fix, self-tagging.

PR #63, opened 2019-02-26
allow audit filtering

deads2k:
allow slicing an audit log by resource, namespace, and name.  useful for GC

```
	# find all GC calls to deployments in any apigroup (extensions or apps)
	openshift-dev-helpers audit -f audit.log --user=system:serviceaccount:kube-system:generic-garbage-collector --resource=deployments.*

	# find all failed calls to kube-system and olm namespaces
	openshift-dev-helpers audit -f audit.log --namespace=kube-system --namespace=openshift-operator-lifecycle-manager --failed-only

	# find all GETs against deployments and any resource under config.openshift.io
	openshift-dev-helpers audit -f audit.log --resource=deployments.* --resource=*.config.openshift.io --verb=get

```

/assign @mfojtik


pkg/cmd/audit/audit.go
```diff
@@ -1,7 +1,10 @@
-package cmd
+package audit
 
 import (
 	"fmt"
+	"strings"
+
+	"k8s.io/apimachinery/pkg/runtime/schema"
```
soltysh: Sort these correctly.

deads2k:
Updated description with usage.  This no longer pulls from a URL.  I think the size is too large to reasonably offer that choice.  It handles recursive directories and works on a series of filters applied as a logical AND.

We probably want `-o json` to allow piping and successive constriction, but this was enough to solve my problem.

mfojtik:
❤️ 

/lgtm

PR #65, opened 2019-02-27
Adding the first draft of the CEE Playlist 

sferich888:
Required gathering items/resources that support will need to start looking at a cluster.

### DIFF
--- a/collection-scripts/gather
+++ b/collection-scripts/gather
@@ -1,0 +1,30 @@
+#!/bin/bash
+
+# Resource List
+resources=()
+
+# Operator Resources
+resources+=(clusteroperators)
+
+# Certificate Resources
+resources+=(certificatesigningrequests)
+
+# Machine/Node Resources
+resources+=(nodes machines machineconfigs)
+
+# Namespaces/Project Resources
+resources+=(ns/default ns/openshift ns/kube-system)
+
+# Storage Resources
+resources+=(volumes)
+
+# Networking Resources
+resources+=(clusternetworks hostsubnets)
+
+# Autoscaler Resources
+resources+=(clusterautoscaler machineautoscaler)
+
+# Run the Collection of Resources using must-gather
+for resource in ${resources[@]}; do
+    /usr/bin/openshift-must-gather inspect ${resource}
+done


collection-scripts/gather
```diff
@@ -0,0 +1,28 @@
+#!/bin/bash
+
+# Resource List
+resources=()
+
+# Operator Resources
+resources+=(clusteroperators)
+
+# Certificate Resources
+resources+=(csr)
```
sferich888: @deads2k I assume given the threads you have included me in this is a thing we want to collect.

sferich888: I updated this to be plural thus why its out dated.

sferich888:
@deads2k @mfojtik can one of you give this a review (on the direction we want to take this). 
I feel its ok, but don't want to merge my own PR's with our someone first reviewing it. 

cc: @RickJWagner

collection-scripts/gather
sanchezl:
```suggestion
resources+=(certificatesigningrequests)
```
`csr` is just a short name, so the plural form is `certificatesigningrequests`:
```console
[sanchezl@monolith must-gather]$  ./openshift-must-gather inspect csrs --base-dir CSRS
Error: the server doesn't have a resource type "csrs"
[sanchezl@monolith must-gather]$  ./openshift-must-gather inspect certificatesigningrequests --base-dir CSR
2019/04/01 21:15:22 Finished successfully with no errors.
[sanchezl@monolith must-gather]$ tree CSR/
CSR/
└── cluster-scoped-resources
    └── certificates.k8s.io
        └── certificatesigningrequests
            ├── csr-29q65.yaml
            ├── csr-49n9f.yaml
            └── csr-9qwh4.yaml

3 directories, 3 files
[sanchezl@monolith must-gather]$ 
```

sferich888: Fixed

collection-scripts/gather
```diff
@@ -0,0 +1,30 @@
...
+resources+=(clusteroperators)
+
+# Certificate Resources
+resources+=(certificatesigningrequests)
+
+# Machine/Node Resources
+resources+=(nodes machines machineconfigs)
+
+# Namespaces/Project Resources
+resources+=(ns/default ns/openshift ns/kube-system)
```
mfojtik: why is `ns/openshift` interesting?

RickJWagner:
I've reviewed this PR and find it good.  (I think all of the listed resources have potential value, depending on what needs to be investigated.)
We may want to add to this over time.  (Example:  I've had some recent work that dealt with users/groups/roles, the information wasn't easy to gather.  This might be my own lack of experience, though, so I don't suggest any changes for the time being.)

collection-scripts/gather
sferich888: To pull content from the OpenShift Namespace? 
I'm not sure what this question is trying to ask.

PR #78, opened 2019-04-24
make events more powerful

deads2k:
Update the events dev helper to be more the audit one with more filtering and slicing.

/assign @mfojtik 

TODO, we need sorting, but this is a step forward.


mfojtik:
/lgtm

agree, this is improvement

PR #80, opened 2019-04-25
audit: add top output type

mfojtik:
NOTE: This is super inefficient implementation of `dev-tools -o top`.

The full syntax is:

```
openshift-dev-helpers audit -f FILE -o top --by user
openshift-dev-helpers audit -f FILE -o top --by verb
```

First will display top 10 users making any request. Second will display top 5 requests per verb (GET, CREATE, etc.)


mfojtik:
I think David will have opinions here.

mfojtik:
@deads2k we can improve as we go, this gets us started I think

PR #82, opened 2019-04-25
change must-gather image base FROM base to cli

sanchezl:
Change the base image used when running `make images` locally to better match what gets built during a release build.


deads2k:
passed must-gather

Issue #85, opened 2019-04-26
Error reporting in a collection script cause "oc adm must-gather" to fail


djfjeff:
I created a custom image with a new collection script called gather-cnv (https://github.com/djfjeff/must-gather/blob/gather-cnv-support/collection-scripts/gather-cnv).

Basically, for now, the new collection script only do : `openshift-must-gather inspect ns/kubevirt`

Manually running the inspect command above fetch the information but still produce some error output : 
```
openshift-must-gather inspect ns/kubevirt

2019/04/26 14:02:40 Gathering data for ns/kubevirt...
2019/04/26 14:02:40     Collecting resources for namespace "kubevirt"...
2019/04/26 14:02:40     Gathering pod data for namespace "kubevirt"...
2019/04/26 14:02:40         Gathering data for pod "virt-api-f8d6cd97-5l68b"
2019/04/26 14:02:41         Unable to gather previous container logs: previous terminated container "virt-api" in pod "virt-api-f8d6cd97-5l68b" not found
2019/04/26 14:02:42         Skipping /version info gathering for pod "virt-api-f8d6cd97-5l68b". Endpoint not found...
2019/04/26 14:02:42         Gathering data for pod "virt-api-f8d6cd97-6h4kx"
2019/04/26 14:02:43         Unable to gather previous container logs: previous terminated container "virt-api" in pod "virt-api-f8d6cd97-6h4kx" not found
2019/04/26 14:02:43         Skipping /version info gathering for pod "virt-api-f8d6cd97-6h4kx". Endpoint not found...
2019/04/26 14:02:43         Gathering data for pod "virt-controller-865b95f6c6-gtplf"
2019/04/26 14:02:43         Unable to gather previous container logs: previous terminated container "virt-controller" in pod "virt-controller-865b95f6c6-gtplf" not found
2019/04/26 14:02:44         Gathering data for pod "virt-controller-865b95f6c6-vl6cc"
2019/04/26 14:02:44         Unable to gather previous container logs: previous terminated container "virt-controller" in pod "virt-controller-865b95f6c6-vl6cc" not found
2019/04/26 14:02:44         Gathering data for pod "virt-handler-5zd2z"
2019/04/26 14:02:44         Unable to gather previous container logs: previous terminated container "virt-handler" in pod "virt-handler-5zd2z" not found
2019/04/26 14:02:44         Gathering data for pod "virt-handler-9q9cp"
[...]

PR #87, opened 2019-04-26
Add Istio

jwendell:
As a separated command so that people may opt-in on running it only
if they have Istio installed:

oc adm must-gather -- gather-istio

### DIFF
--- a/collection-scripts/gather-istio
+++ b/collection-scripts/gather-istio
@@ -1,0 +1,15 @@
+#!/bin/bash
+
+# Resource List
+resources=()
+
+# Istio Resources
+resources+=(ns/istio-operator ns/istio-system)
+
+# Custom Resources
+resources+=(CustomResourceDefinition)
+
+# Run the Collection of Resources using must-gather
+for resource in ${resources[@]}; do
+    /usr/bin/openshift-must-gather inspect ${resource}
+done


sferich888:
@jwendell let's talk about this Monday, idk if this is the way we should go.

collection-scripts/gather-istio
```diff
@@ -0,0 +1,15 @@
+#!/bin/bash
```
sferich888: re-name this file `gather_istio` to stay consistent with other commands in the collection scripts repo.

collection-scripts/gather-istio
```diff
@@ -0,0 +1,15 @@
...
+
+# Resource List
+resources=()
+
+# Istio Resources
+resources+=(ns/istio-operator ns/istio-system)
+
+# Custom Resources
+resources+=(CustomResourceDefinition)
```
sferich888: We likely don't want to collect this. We likely want to collect the specific ISTIO CRD and or Active CR. 

We probably want to have this pull `clusteroperator/ISTIO` or something like that.

collection-scripts/gather-istio
```diff
@@ -0,0 +1,15 @@
...
+
+# Run the Collection of Resources using must-gather
+for resource in ${resources[@]}; do
+    /usr/bin/openshift-must-gather inspect ${resource}
+done
```
sferich888: See: https://github.com/openshift/must-gather/issues/85 and add `exit 0` to this script.

PR #89, opened 2019-05-08
gather: Collect machineconfigpools too

cgwalters:
These are also cluster scoped and useful data describing the
state of the system.


sferich888:
@cgwalters we should discuss what the relationship between nodes, machines and machine configs are. 

I see this as a short term fix.

PR #93, opened 2019-05-30
handle pure negation and don't match empty strings

deads2k:
fixes matching for negative users.


sallyom:
confirmed as an example, `./openshift-dev-helpers audit -f audit.log --user='-system:anonymous'` has the result of returning info for all users except system:anonymous.

If/when there's time, we can add some unit tests for this repo : )

PR #94, opened 2019-05-30
Gather extended network data

rcarrillocruz:
https://jira.coreos.com/browse/SDN-428

### DIFF
--- a/collection-scripts/gather
+++ b/collection-scripts/gather
@@ -19,7 +19,7 @@ resources+=(ns/default ns/openshift ns/kube-system)
 resources+=(storageclasses persistentvolumes volumeattachments)
 
 # Networking Resources
-resources+=(network.config.openshift.io networkoperator.openshift.io clusternetworks hostsubnets)
+resources+=(network.config.openshift.io network.operator.openshift.io clusternetworks hostsubnets)
 
 # Autoscaler Resources
 resources+=(clusterautoscaler machineautoscaler)
--- a/collection-scripts/gather
+++ b/collection-scripts/gather
@@ -19,7 +19,7 @@ resources+=(ns/default ns/openshift ns/kube-system)
 resources+=(storageclasses persistentvolumes volumeattachments)
 
 # Networking Resources
-resources+=(clusternetworks hostsubnets)
+resources+=(network.config.openshift.io networkoperator.openshift.io clusternetworks hostsubnets)
 
 # Autoscaler Resources
 resources+=(clusterautoscaler machineautoscaler)
--- a/collection-scripts/gather
+++ b/collection-scripts/gather
@@ -19,7 +19,7 @@ resources+=(ns/default ns/openshift ns/kube-system)
 resources+=(storageclasses persistentvolumes volumeattachments)
 
 # Networking Resources
-resources+=(network.config.openshift.io network.operator.openshift.io clusternetworks hostsubnets)
+resources+=(clusternetworks hostsubnets)
 
 # Autoscaler Resources
 resources+=(clusterautoscaler machineautoscaler)
--- a/collection-scripts/gather_network_logs
+++ b/collection-scripts/gather_network_logs
@@ -4,18 +4,31 @@ NETWORK_LOG_PATH="${BASE_COLLECTION_PATH}/network_logs/"
 
 mkdir -p ${NETWORK_LOG_PATH}/
 
-WORKERS=$(oc get nodes --no-headers -o custom-columns=":metadata.name")
+NODES=$(oc get nodes --no-headers -o custom-columns=":metadata.name")
+PIDS=()
 
-for WORKER in $WORKERS; do
-    SDN_POD=$(oc -n openshift-sdn get pods --no-headers -o custom-columns=":metadata.name" --field-selector spec.nodeName=$WORKER -l app=sdn)
-    oc -n openshift-sdn exec $SDN_POD -- bash -c "iptables-save -c" > ${NETWORK_LOG_PATH}/${WORKER}_iptables
-    oc -n openshift-sdn exec $SDN_POD -- bash -c "ovs-vsctl show" > ${NETWORK_LOG_PATH}/${WORKER}_ovs_dump
-    oc -n openshift-sdn logs $SDN_POD > ${NETWORK_LOG_PATH}/${SDN_POD}_sdn_logs
+for NODE in $NODES; do
+    SDN_POD=$(oc -n openshift-sdn get pods --no-headers -o custom-columns=":metadata.name" --field-selector spec.nodeName=$NODE -l app=sdn)
+    oc -n openshift-sdn exec $SDN_POD -- bash -c "iptables-save -c" > ${NETWORK_LOG_PATH}/${NODE}_iptables &
+    PIDS+=($!)
+    oc -n openshift-sdn exec $SDN_POD -- bash -c "ovs-vsctl show" > ${NETWORK_LOG_PATH}/${NODE}_ovs_dump &
+    PIDS+=($!)
+    oc -n openshift-sdn logs $SDN_POD > ${NETWORK_LOG_PATH}/${SDN_POD}_sdn_log &
+    PIDS+=($!)
+
+    OVS_POD=$(oc -n openshift-sdn get pods --no-headers -o custom-columns=":metadata.name" --field-selector spec.nodeName=$NODE -l app=ovs)
+    oc -n openshift-sdn exec $OVS_POD -- bash -c "cat /var/log/openvswitch/ovs-vswitchd.log" > ${NETWORK_LOG_PATH}/${OVS_POD}_vswitchd_log &
+    PIDS+=($!)
+    oc -n openshift-sdn exec $OVS_POD -- bash -c "cat /var/log/openvswitch/ovsdb-server.log" > ${NETWORK_LOG_PATH}/${OVS_POD}_ovsdb_log &
+    PIDS+=($!)
+
+    CONTROLLER_POD=$(oc -n openshift-sdn get pods --no-headers -o custom-columns=":metadata.name" --field-selector spec.nodeName=$NODE -l app=sdn-controller)
+    if [ -n "$CONTROLLER_POD" ]; then
+       oc -n openshift-sdn logs $CONTROLLER_POD > ${NETWORK_LOG_PATH}/${CONTROLLER_POD}_controller_logs &
+       PIDS+=($!)
+    fi
 done
 
-MASTERS=$(oc get nodes --no-headers -o custom-columns=":metadata.name" -l node-role.kubernetes.io/master)
-
-for MASTER in $MASTERS; do
-    CONTROLLER_POD=$(oc -n openshift-sdn get pods --no-headers -o custom-columns=":metadata.name" --field-selector spec.nodeName=$MASTER -l app=sdn-controller)
-    oc -n openshift-sdn logs $CONTROLLER_POD > ${NETWORK_LOG_PATH}/${CONTROLLER_POD}_controller_logs
-done
+echo "INFO: Waiting for node network log collection to complete ..."
+wait ${PIDS[@]}
+echo "INFO: Node network log collection to complete."
--- a/collection-scripts/gather_network_logs
+++ b/collection-scripts/gather_network_logs
@@ -4,7 +4,7 @@ NETWORK_LOG_PATH="${BASE_COLLECTION_PATH}/network_logs/"
 
 mkdir -p ${NETWORK_LOG_PATH}/
 
-NODES=$(oc get nodes --no-headers -o custom-columns=":metadata.name")
+NODES="${NETWORK_NODES:-$(oc get nodes --no-headers -o custom-columns=':metadata.name')}"
 PIDS=()
 
 for NODE in $NODES; do
--- a/collection-scripts/gather_network_logs
+++ b/collection-scripts/gather_network_logs
@@ -13,20 +13,12 @@ for NODE in $NODES; do
     PIDS+=($!)
     oc -n openshift-sdn exec $SDN_POD -- bash -c "ovs-vsctl show" > ${NETWORK_LOG_PATH}/${NODE}_ovs_dump &
     PIDS+=($!)
-    oc -n openshift-sdn logs $SDN_POD > ${NETWORK_LOG_PATH}/${SDN_POD}_sdn_log &
-    PIDS+=($!)
 
     OVS_POD=$(oc -n openshift-sdn get pods --no-headers -o custom-columns=":metadata.name" --field-selector spec.nodeName=$NODE -l app=ovs)
     oc -n openshift-sdn exec $OVS_POD -- bash -c "cat /var/log/openvswitch/ovs-vswitchd.log" > ${NETWORK_LOG_PATH}/${OVS_POD}_vswitchd_log &
     PIDS+=($!)
     oc -n openshift-sdn exec $OVS_POD -- bash -c "cat /var/log/openvswitch/ovsdb-server.log" > ${NETWORK_LOG_PATH}/${OVS_POD}_ovsdb_log &
     PIDS+=($!)
-
-    CONTROLLER_POD=$(oc -n openshift-sdn get pods --no-headers -o custom-columns=":metadata.name" --field-selector spec.nodeName=$NODE -l app=sdn-controller)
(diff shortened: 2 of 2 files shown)


collection-scripts/gather_network_logs
```diff
@@ -0,0 +1,21 @@
+#!/bin/bash
+BASE_COLLECTION_PATH="must-gather"
+NETWORK_LOG_PATH="${BASE_COLLECTION_PATH}/network_logs/"
+
+mkdir -p ${NETWORK_LOG_PATH}/
+
+WORKERS=$(oc get nodes --no-headers -o custom-columns=":metadata.name")
```
sferich888: This is not restricted to workers (consider re-naming this to nodes). 
Or ad a label selector like you do below (for Masters)

collection-scripts/gather_network_logs
```diff
@@ -0,0 +1,21 @@
...
+    oc -n openshift-sdn exec $SDN_POD -- bash -c "ovs-vsctl show" > ${NETWORK_LOG_PATH}/${WORKER}_ovs_dump
+    oc -n openshift-sdn logs $SDN_POD > ${NETWORK_LOG_PATH}/${SDN_POD}_sdn_logs
+done
+
+MASTERS=$(oc get nodes --no-headers -o custom-columns=":metadata.name" -l node-role.kubernetes.io/master)
+
+for MASTER in $MASTERS; do
+    CONTROLLER_POD=$(oc -n openshift-sdn get pods --no-headers -o custom-columns=":metadata.name" --field-selector spec.nodeName=$MASTER -l app=sdn-controller)
+    oc -n openshift-sdn logs $CONTROLLER_POD > ${NETWORK_LOG_PATH}/${CONTROLLER_POD}_controller_logs
+done
```
sferich888: I believe this is already captured because of the operator<>namespace<>pod<>container<>log relationship when you as for the networking operator.

collection-scripts/gather_network_logs
```diff
@@ -0,0 +1,21 @@
...
+mkdir -p ${NETWORK_LOG_PATH}/
+
+WORKERS=$(oc get nodes --no-headers -o custom-columns=":metadata.name")
+
+for WORKER in $WORKERS; do
+    SDN_POD=$(oc -n openshift-sdn get pods --no-headers -o custom-columns=":metadata.name" --field-selector spec.nodeName=$WORKER -l app=sdn)
+    oc -n openshift-sdn exec $SDN_POD -- bash -c "iptables-save -c" > ${NETWORK_LOG_PATH}/${WORKER}_iptables
+    oc -n openshift-sdn exec $SDN_POD -- bash -c "ovs-vsctl show" > ${NETWORK_LOG_PATH}/${WORKER}_ovs_dump
+    oc -n openshift-sdn logs $SDN_POD > ${NETWORK_LOG_PATH}/${SDN_POD}_sdn_logs
+done
```
sferich888: If a customer has 500 nodes (or more) do we really want to run this on every node? 
Can we thread the collection of these item. See what I do in [gather_service_logs script](https://github.com/openshift/must-gather/blob/master/collection-scripts/gather_service_logs#L15-L28), with the wait command and the `PIDS` array.

collection-scripts/gather_network_logs
```diff
@@ -0,0 +1,21 @@
...
+
+mkdir -p ${NETWORK_LOG_PATH}/
+
+WORKERS=$(oc get nodes --no-headers -o custom-columns=":metadata.name")
+
+for WORKER in $WORKERS; do
+    SDN_POD=$(oc -n openshift-sdn get pods --no-headers -o custom-columns=":metadata.name" --field-selector spec.nodeName=$WORKER -l app=sdn)
+    oc -n openshift-sdn exec $SDN_POD -- bash -c "iptables-save -c" > ${NETWORK_LOG_PATH}/${WORKER}_iptables
+    oc -n openshift-sdn exec $SDN_POD -- bash -c "ovs-vsctl show" > ${NETWORK_LOG_PATH}/${WORKER}_ovs_dump
+    oc -n openshift-sdn logs $SDN_POD > ${NETWORK_LOG_PATH}/${SDN_POD}_sdn_logs
```
sferich888: I believe this is already captured because of the operator<>namespace<>pod<>container<>log relationship when you as for the networking operator.

collection-scripts/gather_network_logs
sferich888: @deads2k I feel like this (the two `exec` commands) should be worked directly into must-gather the binary (over some script like this). Something like what we do with: https://github.com/openshift/must-gather/blob/master/pkg/cmd/inspect/pod.go#L102-L123

collection-scripts/gather
```diff
@@ -19,7 +19,7 @@ resources+=(ns/default ns/openshift ns/kube-system)
 resources+=(storageclasses persistentvolumes volumeattachments)
 
 # Networking Resources
-resources+=(clusternetworks hostsubnets)
+resources+=(network.config.openshift.io networkoperator.openshift.io clusternetworks hostsubnets)
```
sferich888: I believe the networkoperator is already captured by [clusteroperators on line 7](https://github.com/openshift/must-gather/pull/94/files#diff-49627f729628fe028600d501df843ea1R7)

I believe network.config.openshift.io is already captured because of the operator<>config mapping (with the networkoperator)

deads2k: > I believe the networkoperator is already captured by [clusteroperators on line 7](https://github.com/openshift/must-gather/pull/94/files#diff-49627f729628fe028600d501df843ea1R7)

it should be.  If it isn't, the related resources on the clusteroperator resource should be updated to reflect the need.

collection-scripts/gather_network_logs
deads2k: > @deads2k I feel like this (the two `exec` commands) should be worked directly into must-gather the binary (over some script like this). Something like what we do with: https://github.com/openshift/must-gather/blob/master/pkg/cmd/inspect/pod.go#L102-L123

I don't think so.  The inspect command is about what you should do with **any** pod.  These exec calls are abut what you do for a *particular* pod.  It doesn't make sense to collect that information on all logs.

sferich888:
> This is a good start. I'd recommend taking a look at what https://github.com/openshift/openshift-sdn/blob/master/hack/debug.sh gathered too per node to see what else can help debug. @danwinship wrote that originally and it has been useful.

Too much of that script are 'tests'! If we want customers running test to see if networking is working why are we not having the run the test suite? 

`oc run --attach --stdin --tty --image=registry.svc.ci.openshift.org/openshift/origin-v4.0:tests --overrides='{"kind":"Pod","apiVersion":"v1","metadata":{"name":"openshift-tests","creationTimestamp":null,"labels":{"run":"openshift-tests"}},"spec":{"containers":[{"name":"openshift-tests","image":"registry.svc.ci.openshift.org/openshift/origin-v4.0:tests","env":[{"name":"KUBECONFIG","value":"/kubeconfig/kubeconfig"}],"args":["/bin/bash","-c","openshift-tests run openshift/conformance"],"resources":{},"imagePullPolicy":"IfNotPresent","volumeMounts":[{"name":"kubeconfig-volume","mountPath":"/kubeconfig"}]}],"restartPolicy":"Never","dnsPolicy":"ClusterFirst","volumes":[{"configMap":{"name":"kubeconfig"},"name":"kubeconfig-volume"}]}}' --rm --restart=Never openshift-tests`

I am sure we could break out a set of a simple test for this? Or use an existing suite, right?

knobunc:
@sferich888 true.  I was not considering running the tests.  But it also runs many commands to gather useful networking information.

We should make sure we are getting the stuff from these sections:
- https://github.com/openshift/openshift-sdn/blob/master/hack/debug.sh#L316 
- https://github.com/openshift/openshift-sdn/blob/master/hack/debug.sh#L59 
- https://github.com/openshift/openshift-sdn/blob/master/hack/debug.sh#L109

danwinship:
> I'd recommend taking a look at what https://github.com/openshift/openshift-sdn/blob/master/hack/debug.sh gathered too per node to see what else can help debug.

We never *removed* anything from that though, so a lot of the stuff it is gathering is only useful for diagnosing problems that never occur any more.

collection-scripts/gather_network_logs
squeed: Do you have a sample output from must-gather where we can check?

collection-scripts/gather_network_logs
squeed: I would say, yes we do. If we're diagnosing some kind of networking weirdness, it's likely that the client and server will be on different nodes -- and we'll want the networking dump for both of them. That's one of the more common reasons we have to send tickets back to support - because they only know half of the picture.

If there's a way to qualify which nodes should be dumped, that would be useful, but I'm not familiar enough with how the script is intended to be run to make that call.

collection-scripts/gather
squeed: Will this fail if clusternetworks and hostsubnets do not exist? They're only present for openshift-sdn. Other networking providers won't have the resources defined.

collection-scripts/gather_network_logs
squeed: Can you gather the OVS logs too?

collection-scripts/gather_network_logs
squeed: agreed, this should be on all nodes.

collection-scripts/gather_network_logs
```diff
@@ -4,18 +4,31 @@ NETWORK_LOG_PATH="${BASE_COLLECTION_PATH}/network_logs/"
...
+    PIDS+=($!)
 
-for MASTER in $MASTERS; do
-    CONTROLLER_POD=$(oc -n openshift-sdn get pods --no-headers -o custom-columns=":metadata.name" --field-selector spec.nodeName=$MASTER -l app=sdn-controller)
-    oc -n openshift-sdn logs $CONTROLLER_POD > ${NETWORK_LOG_PATH}/${CONTROLLER_POD}_controller_logs
+    CONTROLLER_POD=$(oc -n openshift-sdn get pods --no-headers -o custom-columns=":metadata.name" --field-selector spec.nodeName=$NODE -l app=sdn-controller)
+    if [ -n "$CONTROLLER_POD" ]; then
+       oc -n openshift-sdn logs $CONTROLLER_POD > ${NETWORK_LOG_PATH}/${CONTROLLER_POD}_controller_logs &
+       PIDS+=($!)
+    fi
```
sferich888: https://github.com/openshift/must-gather/pull/94/files/94623a418b7d7b733709245012286a17b67ade26..98cfda1aa47bc9d4cd72113c45e2a29f7c37157d#discussion_r289164403 still applies

collection-scripts/gather
danwinship: There's no need to collect the clusternetwork object. Its contents are completely deterministic given `network.operator.openshift.io` (which is missing a `.` here btw).

> Will this fail if clusternetworks and hostsubnets do not exist? They're only present for openshift-sdn. Other networking providers won't have the resources defined.

This is a problem with most of `gather_network_logs` as well. Probably we should be plugin-agnostic in `gather`, and check what the configured network plugin is in `gather_network_logs`, and collect accordingly.

danwinship:
> It would be really nice if there was a way to focus what nodes the information is gathered for. But if there is no way to accomplish that, we should impose a limit and only gather information from the first N workers. (Perhaps N is 20?)

If we can't guarantee that it's going to include the "relevant" nodes then 20 is not really any better than 1. No one is going to look throught 20 sets of iptables and OVS dumps. They'll look at 1 or 2, and if they don't see anything obviously wrong, they'll ask the customer to get the data from the relevant nodes.

collection-scripts/gather
sferich888: Can / should we add a dpendency in the must-gather binary for the networking operator and the denoted objects here? 

@deads2k

collection-scripts/gather_network_logs
sferich888: https://gcsweb-ci.svc.ci.openshift.org/gcs/origin-ci-test/pr-logs/pull/operator-framework_operator-lifecycle-manager/889/pull-ci-operator-framework-operator-lifecycle-manager-master-e2e-aws/2148/artifacts/e2e-aws/must-gather/

collection-scripts/gather
```diff
@@ -19,7 +19,7 @@ resources+=(ns/default ns/openshift ns/kube-system)
 resources+=(storageclasses persistentvolumes volumeattachments)
 
 # Networking Resources
-resources+=(clusternetworks hostsubnets)
+resources+=(network.config.openshift.io network.operator.openshift.io clusternetworks hostsubnets)
```
sferich888: This is already collected,https://storage.googleapis.com/origin-ci-test/pr-logs/pull/openshift_must-gather/94/pull-ci-openshift-must-gather-master-e2e-aws/45/artifacts/e2e-aws/must-gather/cluster-scoped-resources/config.openshift.io/networks.yaml so we don't need this explicitly.

sferich888: https://gcsweb-ci.svc.ci.openshift.org/gcs/origin-ci-test/pr-logs/pull/openshift_must-gather/94/pull-ci-openshift-must-gather-master-e2e-aws/45/artifacts/e2e-aws/must-gather/cluster-scoped-resources/network.openshift.io/

sferich888: https://storage.googleapis.com/origin-ci-test/pr-logs/pull/openshift_must-gather/94/pull-ci-openshift-must-gather-master-e2e-aws/45/artifacts/e2e-aws/must-gather/cluster-scoped-resources/operator.openshift.io/networks/cluster.yaml

collection-scripts/gather_network_logs
```diff
@@ -0,0 +1,34 @@
+#!/bin/bash
+BASE_COLLECTION_PATH="must-gather"
+NETWORK_LOG_PATH="${BASE_COLLECTION_PATH}/network_logs/"
+
+mkdir -p ${NETWORK_LOG_PATH}/
+
+NODES=$(oc get nodes --no-headers -o custom-columns=":metadata.name")
```
sferich888: Pleas create an input to the script that takes a string of space delimited node names. 

If nothing is supplied (this command can be run); and a warning should be printed to the user that this is not recommended (to run the script without node names) and that it may take a large amount of time (print how many nodes were collection information from)!

collection-scripts/gather_network_logs
sferich888: I'm not seeing this get collected in CI so we should keep this (ignore my original comment).

collection-scripts/gather_network_logs
sferich888: Can we resolve this comment?

collection-scripts/gather
rcarrillocruz: So providing we can't know if a given network object will exist as that depends on the plugin, I suggest we remove the clusternetworks and hostsubnets line from reousrces, and we add logic  in gather_network_logs to gather objects per the plugin used.

collection-scripts/gather_network_logs
rcarrillocruz: makes sense

collection-scripts/gather_network_logs
rcarrillocruz: I don't follow this comment, the link doesn't tell me what you want, can you pls clarify?

sferich888: I thought this was collected by a relationship in must-gather the binary (I'm validating that that is not correct), and that this collection is ok.

collection-scripts/gather
sferich888: We can do that, however I would like to see what you mean by 'logic' (I think we want to meet these scripts VERY simple.

Today the philosophy we use, is to try and collect data, fail quickly and move on (it's better to ask not get what we request (explain why) , and move on).

collection-scripts/gather_network_logs
```diff
@@ -0,0 +1,26 @@
+#!/bin/bash
+BASE_COLLECTION_PATH="must-gather"
+NETWORK_LOG_PATH="${BASE_COLLECTION_PATH}/network_logs/"
+
+mkdir -p ${NETWORK_LOG_PATH}/
+
+NODES="${NETWORK_NODES:-$(oc get nodes --no-headers -o custom-columns=':metadata.name')}"
```
sferich888: NETWORK_NODES needs to be a bash argument (I don't think we can read in environment variables; like this).

sferich888:
/cherrypick release-4.1

@rcarrillocruz  can you please fill a bz and connect it to the generated PR

PR #99, opened 2019-06-11
[WIP]Allow installer pods data to be collected

ravisantoshgudimetla:
Allow installer pods data to be collected in whatever the
state they're in. This allows logs to be collected to be
collected from those installer pods

### DIFF
--- a/pkg/cmd/inspect/pod.go
+++ b/pkg/cmd/inspect/pod.go
@@ -16,8 +16,13 @@ import (
 )
 
 func (o *InspectOptions) gatherPodData(destDir, namespace string, pod *corev1.Pod) error {
-	if pod.Status.Phase != corev1.PodRunning {
-		log.Printf("        Skipping container data collection for pod %q: Pod not running\n", pod.Name)
+	isInstallerPod := false
+	// allow installer pods logs to be collected in what ever state they are in
+	if strings.HasPrefix(pod.Name, "installer-") {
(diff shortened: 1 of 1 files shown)


RickJWagner:
This logic looks incorrect, if the intention is to log for either a running pod OR an installer pod.

if pod.Status.Phase != corev1.PodRunning || !isInstallerPod {

-------------- Try this and see. Only true/true (A running installer pod) will log ------------------------
package main

import "fmt"

func main() {
            installerpod := true
            running := true
            if installerpod != true || running != true {
	    	fmt.Println("not installerpod or not running, won't collect")
            }else{
	    	fmt.Println("installerpod or running, will collect")
            }
    }

ravisantoshgudimetla:
Ya, it was not intention to push this change. I was playing around with must gather and forgot to push the latest changes. I will push them in a bit.

ravisantoshgudimetla:
And by the way, that was not the intention

sferich888:
@ravisantoshgudimetla are you planning on following up with this PR? 
If yes, how are we not going to get this confused, with https://github.com/openshift/installer/blob/master/cmd/openshift-install/gather.go#L156-L162 and what should be collected here? 

Do we need to mirror this logic is some way to get data of the bootstrap system (if its still around) to understand why the install pods were failing? 

@mfojtik what are your thoughs on this one?

PR #100, opened 2019-06-13
Bug 1720321 - must-gather does not collect all non-core CRs in the given namespace

sanchezl:
When empty `Name` and/or `Namespace` values are specified in a `ClusterOperator.spec.relatedObjects` gather everything that matches the non-empty field that are specified.

For example, the following entry in `ClusterOperator.spec.relatedObjects` would collect all `CatalogSources` in the `openshift-marketplace` namespace:
```
status:
  relatedObjects:
  - group:     operators.coreos.com
    namespace: openshift-marketplace
    resource:  catalogsources
```

https://bugzilla.redhat.com/show_bug.cgi?id=1720321

### DIFF
--- a/pkg/cmd/inspect/resource.go
+++ b/pkg/cmd/inspect/resource.go
@@ -69,15 +69,17 @@ func InspectResource(info *resource.Info, context *resourceContext, o *InspectOp
 				continue
 			}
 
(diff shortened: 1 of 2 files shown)


sanchezl:
@soltysh , @tnozicka , @mfojtik , @deads2k : PTAL

jwforres:
@sanchezl is this just waiting on review?

PR #113, opened 2019-07-16
[WIP] Collect metal3 resources

pkliczewski:
As metal3 is part of openshift I am moving collection scripts from github.com/djfjeff/metal3-must-gather to this repo.

### DIFF
--- a/collection-scripts/gather_metal3
+++ b/collection-scripts/gather_metal3
@@ -1,0 +1,19 @@
+#!/bin/bash
+
+# Resource List
+resources=()
+
+# Metal3 related Resources
+resources+=(ns/openshift-machine-api ns/openshift-machine-config-operator)
+
+for i in $(/usr/bin/oc get crd | grep metal3 | awk '{print $1}')
+do
+  resources+=(crd/$i)
+done
+
+# Run the collection of resources using must-gather
+for resource in ${resources[@]}; do
+  /usr/bin/openshift-must-gather inspect ${resource}
+done
+
+exit 0
--- a/collection-scripts/gather
+++ b/collection-scripts/gather
@@ -37,3 +37,6 @@ done
 
 # Gather Service Logs (using a suplamental Script); Scoped to Masters.
 /usr/bin/gather_service_logs master
+
+# Gather Metal3 related resources
+/usr/bin/gather_metal3


collection-scripts/gather_metal3
```diff
@@ -0,0 +1,19 @@
+#!/bin/bash
+
+# Resource List
+resources=()
+
+# Metal3 related Resources
+resources+=(ns/openshift-machine-api ns/openshift-machine-config-operator)
```
sferich888: Don't we already collect this with related resources and cluster operators?

collection-scripts/gather_metal3
```diff
@@ -0,0 +1,19 @@
...
+
+for i in $(/usr/bin/oc get crd | grep metal3 | awk '{print $1}')
```
sferich888: Should this be gathered by related resources?

collection-scripts/gather_metal3
pkliczewski: My understanding was that we need to collect this. @etingof please confirm

collection-scripts/gather_metal3
pkliczewski: My understanding was that we need to collect this. @etingof please confirm

collection-scripts/gather_metal3
etingof: @sferich888's comment makes sense to me. But I'd defer this to @dhellmann to be sure.

russellb:
> I don't think we need this. If the metal3 components are a cluster operator (@eparis suggested to me that it would be), then the base collection script should collect the things your looking for if the operator and its related components collect define proper related resources.

The metal3 related components are all integrated under the machine-api-operator.  If there's operator level configuration, I suspect it would be there (machine-api-operator).  

There's a machine actuator already integration (cluster-api-provider-baremetal) but there's a lot more being integrated here: https://github.com/openshift/machine-api-operator/pull/302

ohadlevy:
just wondering, how would we collect Ironic logs? (or the IPMI call traces etc?)

sferich888:
How would you do this today without must gather?

russellb:
> How would you do this today without must gather?

Grab the logs of the containers running in our bare metal management pod in the `openshift-machine-api` namespace.

I see that the `machine-api-operator` specifies its namespace as a `RelatedObject` in its `ClusterOperator` resource.  It looks like that will automatically collect all custom resources in that namespace, which covers what this patch was doing.

This patch wasn't addressing logs.  Can you give some pointers regarding log collection?

Related - an integration guide in this repo would be helpful.

pkliczewski:
> 
> This patch wasn't addressing logs. Can you give some pointers regarding log collection?
> 
One of the goals of this PR was to trigger discussion whether we are not missing anything. The code was moved from `djfjeff/metal3-must-gather` but there was no assumption that it is complete.

PR #115, opened 2019-07-31
Bug 1723472: must-gather should redact kubectl.kubernetes.io/last-applied-configuration in secrets

sanchezl:
Redact the `kubectl.kubernetes.io/last-applied-configuration` in Secrets to avoid exposing non-public data Secrets after an `oc apply -f` update.


deads2k:
Thanks.  The content of many annotations is essential to debugging behavior in many cases.

/lgtm

/cherrypick release-4.1

PR #118, opened 2019-08-05
Fix machine-api resources

michaelgugino:
Current must-gather script returns no items for
things inside openshift-machine-api namespace.

This commit adds support for those items by specifying
namespace in gather script.

https://jira.coreos.com/browse/CLOUD-558

### DIFF
--- a/collection-scripts/gather
+++ b/collection-scripts/gather
@@ -3,7 +3,10 @@
 # Resource List
 resources=()
 
-# Cluster Version Information 
+# machine-api resources
+machineresources=()
+
+# Cluster Version Information
 resources+=(clusterversion ns/openshift-cluster-version)
 
 # Operator Resources
@@ -12,8 +15,11 @@ resources+=(clusteroperators)
 # Certificate Resources
 resources+=(certificatesigningrequests)
 
-# Machine/Node Resources
-resources+=(nodes machines machineconfigs machineconfigpools)
+# MCO/Node Resources
+resources+=(nodes machineconfigs machineconfigpools)
+
+# machine-api resources
+machineresources+=(machines machinesets)
 
 # Namespaces/Project Resources
 resources+=(ns/default ns/openshift ns/kube-system ns/openshift-etcd)
@@ -25,11 +31,16 @@ resources+=(storageclasses persistentvolumes volumeattachments)
 resources+=(clusternetworks hostsubnets)
 
 # Autoscaler Resources
-resources+=(clusterautoscaler machineautoscaler)
+machineresources+=(clusterautoscaler machineautoscaler)
 
 # Run the Collection of Resources using must-gather
 for resource in ${resources[@]}; do
     /usr/bin/openshift-must-gather inspect ${resource}
+done
+
+# Run the Collection of MachineAPIResources using must-gather
+for resource in ${machineresources[@]}; do
+    /usr/bin/openshift-must-gather inspect ${resource} --namespace openshift-machine-api
 done
 
 # Collect System Audit Logs


collection-scripts/gather
```diff
@@ -25,13 +31,18 @@ resources+=(storageclasses persistentvolumes volumeattachments)
...
 # Autoscaler Resources
-resources+=(clusterautoscaler machineautoscaler)
+machineresources+=(clusterautoscaler machineautoscaler)
 
 # Run the Collection of Resources using must-gather
 for resource in ${resources[@]}; do
     /usr/bin/openshift-must-gather inspect ${resource}
 done
 
+# Run the Collection of MachineAPIResources using must-gather
```
sferich888: Why collect this separate? Why not just included this in **resources**?

collection-scripts/gather
```diff
@@ -12,8 +15,11 @@ resources+=(clusteroperators)
 # Certificate Resources
 resources+=(certificatesigningrequests)
 
-# Machine/Node Resources
-resources+=(nodes machines machineconfigs machineconfigpools)
+# MCO/Node Resources
+resources+=(nodes machineconfigs machineconfigpools)
```
sferich888: Do we have a relation ship between these? 
IE: If we get a node; do we need to know its machineconfigs or machineconfigpools? 

If so - it might be better to put this relationship into to the must-gather binary so every time a node is collected *(or one of the other objects); the proper related objects are collected.

ingvagabund: I can't speak for everyone though I believe `machineconfigs` and `machineconfigpools` are namespace-less. That's why we need to query for machines separately since they live only under `openshift-machine-api` namespace.

ingvagabund: @runcom FYI ^^

ashcrow: See https://github.com/openshift/machine-config-operator/blob/master/docs/MachineConfigServer.md#detailed-design

ingvagabund: @sferich888 I have a feeling it's impossible to know which machine config was used for a given node. Based on posted [1], I don't think node object has any field/label/annotation that can say which endpoint was used. Also, worker nodes have different image in general. Each image knows an endpoint (most likely it's hard-coded in it when it's baked). Due to security reason so you can't pretend to be master when you are worker. Or, I am wrong @runcom @ashcrow ?

[1] https://github.com/openshift/machine-config-operator/blob/master/docs/MachineConfigServer.md#detailed-design

sferich888:
/lgtm

After talking this over with a few folks; this seems like the least bad solution to implement to get us the needed data. I am concerned that we could get back a MASS amount of information ased on the number of nodes; but I don't think that can be avoided here, in a mass collection tool.

eparis:
I believe we want to do the equivalent of:
```
                   {
                        "group": "operators.coreos.com",
                        "name": "",
                        "namespace": "openshift-marketplace",
                        "resource": "OperatorSource"
                    },
```
in related resources instead of trying to do it like this.

collection-scripts/gather
sferich888: According to https://docs.openshift.com/container-platform/4.1/machine_management/applying-autoscaling.html the `clusterautoscaler` is something you would find in the `default` namespace while the `machineautoscaler` is something you find out of the `openshift-machine-api` namespace. 

Should we just do some loop or update the loop above to get things from all namespaces?

sferich888: Let's consider https://github.com/openshift/must-gather/pull/127/files over this PR.

sferich888:
/close
closing in favor of https://github.com/openshift/must-gather/pull/127

PR #120, opened 2019-08-08
Bug 1726197:  must-gather does not account for command usage in liveness probe

sanchezl:
must-gather attempts to gather info from `/healthz`, `/version`, and `/metrics` endpoints when gathering info from a container. Not all containers implement these endpoints, so this PR makes a slightly better attempt at identifying which containers most likely don't contain these endpoints.

### DIFF
--- a/pkg/cmd/inspect/pod.go
+++ b/pkg/cmd/inspect/pod.go
@@ -66,10 +66,20 @@ func (o *InspectOptions) gatherContainerInfo(destDir string, pod *corev1.Pod, co
 		return err
 	}
 
-	if len(container.Ports) == 0 {
+	switch len(container.Ports) {
+	case 0:
 		log.Printf("        Skipping container endpoint collection for pod %q container %q: No ports\n", pod.Name, container.Name)
 		return nil
+	case 1:
+		if container.Ports[0].Name == "grpc" {
+			log.Printf("        Skipping container endpoint collection for pod %q container %q: No suitable ports\n", pod.Name, container.Name)
+			return nil
+		}
+	default:
+		log.Printf("        Skipping container endpoint collection for pod %q container %q: Too many ports\n", pod.Name, container.Name)
+		return nil
 	}
+
 	port := &util.RemoteContainerPort{
 		Protocol: "https",
 		Port:     container.Ports[0].ContainerPort,


pkg/cmd/inspect/pod.go
```diff
@@ -66,10 +66,20 @@ func (o *InspectOptions) gatherContainerInfo(destDir string, pod *corev1.Pod, co
...
+	case 0:
 		log.Printf("        Skipping container endpoint collection for pod %q container %q: No ports\n", pod.Name, container.Name)
 		return nil
+	case 1:
+		if container.Ports[0].Name == "grpc" {
+			log.Printf("        Skipping container endpoint collection for pod %q container %q: No suitable ports\n", pod.Name, container.Name)
+			return nil
+		}
+	default:
+		log.Printf("        Skipping container endpoint collection for pod %q container %q: Too many ports\n", pod.Name, container.Name)
```
soltysh: Hmm.. I wonder if this is backwards incompatible change. Because previously with more than one we'd at least check the first, shouldn't we check all? I'm not saying previous was correct, but I think we should think about checking all containers, you can't assume the first is the only/most important one, right?

soltysh: @sanchezl iirc there's a separate BZ about scraping more than a single port (see 1726197https://bugzilla.redhat.com/show_bug.cgi?id=1741901).

soltysh:
> @soltysh @sanchezl why didn't inspect move into `oc adm`?

Timing

pkg/cmd/inspect/pod.go
sanchezl: > Hmm.. I wonder if this is backwards incompatible change. Because previously with more than one we'd at least check the first, shouldn't we check all?

When tested, this PR does not change the must-gather output except that some empty `healthz` sub directories are no longer created.

sanchezl: > @sanchezl iirc there's a separate BZ about scraping more than a single port (see 1726197https://bugzilla.redhat.com/show_bug.cgi?id=1741901).

I've opened enhancement https://github.com/openshift/enhancements/pull/40 for review.

soltysh: Gotcha, but since you're here already why not fix it to scrape all?

soltysh:
This can be closed now, we need to move this to https://github.com/openshift/oc
@sanchezl would you mind?
/close

PR #127, opened 2019-09-03
Fixing machine collection 

sferich888:
This is to supplement https://github.com/openshift/machine-api-operator/pull/391 
We should deprecate https://github.com/openshift/must-gather/pull/118 in favor of this PR.

### DIFF
--- a/collection-scripts/gather
+++ b/collection-scripts/gather
@@ -13,7 +13,7 @@ resources+=(clusteroperators)
 resources+=(certificatesigningrequests)
 
 # Machine/Node Resources
-resources+=(nodes machines machineconfigs machineconfigpools)
+resources+=(nodes machineconfigs machineconfigpools)
 
 # Namespaces/Project Resources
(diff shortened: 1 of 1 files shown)


sferich888:
@michaelgugino please review

RickJWagner:
It looks like 118 shuffles around the inspection (namespace-wise) of clusterautoscaler, machineautoscaler, machines.  Also added inspection of machinesets.  

If we forego 118 and instead go with 127 and 391, we get:
127 - Removes from inspection clusterautoscaler, machineautoscaler, machines
391 - Adds inspection of machines, machinesets

Net:  127/391 instead of 118 means we don't inspect clusterautoscaler, machineautoscaler (at least not as seen in this group of PRs.)

Is that the intention?  If so, I'll be glad to LGTM

sferich888:
We do inspect the cluster and machine autoscalers with openshift/cluster-autoscaler-operator#117 

In short you have the right summation of all the changes, but is spread across 3-4 prs

PR #129, opened 2019-09-20
generate /must-gather/version file

sanchezl:
Conform to must-gather spec: https://github.com/openshift/enhancements/blame/master/must-gather.md#L113

Sample output:
```
openshift/must-gather
4.2-2019-09-19-151434
```
If the version can't be determined from the image id, the entire image id is provided instead.

### DIFF
--- a/collection-scripts/gather
+++ b/collection-scripts/gather
@@ -1,5 +1,10 @@
 #!/bin/bash
 
+# generate /must-gather/version file
+. version
+echo "openshift/must-gather" > /must-gather/version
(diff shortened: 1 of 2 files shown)


collection-scripts/gather
```diff
@@ -1,5 +1,10 @@
 #!/bin/bash
 
+# generate /must-gather/version file
+. version
+echo "openshift/must-gather" > /must-gather/version
```
sferich888: Is this necessary? You already are putting it in a file by the same name.

PR #130, opened 2019-09-27
migrate dependencies to go.mod

sanchezl:


deads2k:
/lgtm

But you owe a bash script shim to oc adm must-gather

PR #133, opened 2019-10-14
gather host files for debugging and fix typo

rphillips:
Adds node host files to the must gather logs.

/cc @runcom @sjenning @RobertKrawitz

### DIFF
--- a/collection-scripts/gather
+++ b/collection-scripts/gather
@@ -38,5 +38,8 @@ done
 # Collect System Audit Logs
 /usr/bin/gather_audit_logs
 
+# Collect host files
+/usr/bin/gather_host_files
+
 # Gather Service Logs (using a suplamental Script); Scoped to Masters.
 /usr/bin/gather_service_logs master
--- a/collection-scripts/gather_host_files
+++ b/collection-scripts/gather_host_files
@@ -1,0 +1,26 @@
+#!/bin/bash
+BASE_COLLECTION_PATH="/must-gather"
+HOSTFILES_LOG_PATH="${BASE_COLLECTION_PATH}/hostfiles/"
+
+mkdir -p ${HOSTFILES_LOG_PATH}/
+
+if [ $# -eq 0 ]; then
+    echo "WARNING: Collecting host files on ALL nodes in your cluster. This could take a large amount of time." >&2
+fi
+
+NODES="${@:-$(oc get nodes --no-headers -o custom-columns=':metadata.name')}"
+PIDS=()
+
+for NODE in $NODES; do
+    MCD_POD=$(oc -n openshift-machine-config-operator get pods --no-headers -o custom-columns=":metadata.name" --field-selector spec.nodeName=$NODE -l k8s-app=machine-config-daemon)
+    oc -n openshift-machine-config-operator exec $MCD_POD -- bash -c "cat /rootfs/etc/kubernetes/kubelet.conf" > ${HOSTFILES_LOG_PATH}/${NODE}_kubelet.conf &
+    PIDS+=($!)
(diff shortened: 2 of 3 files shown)


sferich888:
Don't we collect this with related resources?

rphillips:
@sferich888 I haven't seen these files in must-gather. What are 'related resources'?

sferich888:
https://github.com/openshift/must-gather/pull/118 (if you follow this to the bottom of the rabbit hole) has an example of 'related resources'.

collection-scripts/gather_host_files
```diff
@@ -0,0 +1,26 @@
...
+for NODE in $NODES; do
+    MCD_POD=$(oc -n openshift-machine-config-operator get pods --no-headers -o custom-columns=":metadata.name" --field-selector spec.nodeName=$NODE -l k8s-app=machine-config-daemon)
+    oc -n openshift-machine-config-operator exec $MCD_POD -- bash -c "cat /rootfs/etc/kubernetes/kubelet.conf" > ${HOSTFILES_LOG_PATH}/${NODE}_kubelet.conf &
+    PIDS+=($!)
+    oc -n openshift-machine-config-operator exec $MCD_POD -- bash -c "cat /rootfs/etc/containers/registries.conf" > ${HOSTFILES_LOG_PATH}/${NODE}_registries.conf &
+    PIDS+=($!)
+    oc -n openshift-machine-config-operator exec $MCD_POD -- bash -c "cat /rootfs/etc/crio/crio.conf" > ${HOSTFILES_LOG_PATH}/${NODE}_crio.conf &
+    PIDS+=($!)
+done
+
```
sferich888: Shouldn't all of these files be in either some machine config, later down my the MCO at bootstrap? 

Do we have to take this path to get to these files?

rphillips: We want these the files from the node to be included in must gather to help debugging. Instead of relying on the files within the MCO ignition config.

sferich888: Can we co locate these with files within the MCO ignition config? 
Should we only pull these if the MCO file and the file on disk are different?

RobertKrawitz: Debugging is going to be easier if the files are always present in the must-gather archive, vs. having to go off and find them somewhere.  Remember, the point here is to gather information if things go wrong, possibly including someone editing them by hand after disabling the CVO.

sferich888: The questions here are more about asking when we collect these files. The way this PR is currently setup on large clusters (with large number of nodes) we could potentially be collected 1000's of files. 

So I'm asking if we should be first be comparing the files on disk with the MCO ignition files and only collection files if we see a difference. This is an expense and possibly time consuming operation to exec into a pod on every node.

I'm also worried that issue like expired certificate signing requests coul block operations like this from working or what's broken about a cluster inhibiting this from working and this is needing to collect it with Ash in the end.

rphillips: Don't the other must gather scripts exec to every node?

collection-scripts/gather
```diff
@@ -38,5 +38,8 @@ done
 # Collect System Audit Logs
 /usr/bin/gather_audit_logs
 
+# Collect host files
+/usr/bin/gather_host_files
```
sferich888, line 42: If were are going to gather files from 'every node' then we shouldn't to this by default - some clusters have LARGE node sets, and this could cause must-gather to run for an indefinite amount of time.

collection-scripts/gather_host_files
sferich888: Other scripts exec into pods as well, yes, however they don't do this on 'every' node in the cluster by default, (normally its a non-default option, and done as part of additional debugging).

sferich888 (changes requested):

PR #88, opened 2019-04-29
threading mg execution and ensuring that we alway exit 0 (true)

sferich888:
@RickJWagner @coreydaley PTAL
@sanchezl  you may want to review as well.

Issue #6, opened 2019-01-12
Errors don't align with reason for failure


sferich888:
Example:
```
$ bin/openshift-must-gather info clusteroperators
Error: the server doesn't have a resource type "clusteroperators"
```
is caused because the user's kubeconfig are not set and/or the active context is not authenticated with the cluster. 

We should provide a better error and/or things to try (like logging in) if something like this happens.

juanvallejo:
Upstream PR: https://github.com/kubernetes/kubernetes/pull/73608

juanvallejo:
We'll have to re-bump dependencies to pick up that fix. It has already made it into the cli-runtime repo: https://github.com/kubernetes/cli-runtime/commit/3d3b57cfe43cffbcef88581facc53b03aa7dcd67

sferich888:
@sanchezl should we keep this open? 
@deads2k what are your thoughts?

PR #136, opened 2019-11-14
Split resources to named and groups

soltysh:
/assign @sferich888

### DIFF
--- a/collection-scripts/gather
+++ b/collection-scripts/gather
@@ -5,37 +5,42 @@
 echo "openshift/must-gather" > /must-gather/version
 version >> /must-gather/version
 
-# Resource List
-resources=()
+# Named resource list, eg. ns/openshift-config
+named_resources=()
+
+# Resource groups list, eg. pods
+group_resources=()
 
 # Cluster Version Information
-resources+=(clusterversion ns/openshift-cluster-version)
+named_resources+=(ns/openshift-cluster-version)
+group_resources+=(clusterversion)
 
 # Operator Resources
-resources+=(clusteroperators)
+group_resources+=(clusteroperators)
 
 # Certificate Resources
-resources+=(certificatesigningrequests)
+group_resources+=(certificatesigningrequests)
 
 # Machine/Node Resources
-resources+=(nodes)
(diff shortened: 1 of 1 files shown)


sferich888:
@mfojtik I think we still want to pus/fix this correct? 
/lgtm but we need it to have merge conflicts fixed.

sferich888:
I think this PR is a way of addressing https://github.com/openshift/must-gather/issues/6  (however its taking an alternative approach, to what is discussed in that issue).

sferich888:
This PR should obsolete https://github.com/openshift/must-gather/pull/88 as well. IE if this PR merges we need to close #88 as its not longer needed.

soltysh:
@sferich888 rebased and ready for merge

soltysh:
/hold
looks like the failure is real

soltysh:
This PR actually found a bug in `oc adm inspect` when you specify multiple resources. Namely `oc adm inspect clusterversions clusteroperators` will fail and it should not. Working on a fix...

soltysh:
Actually, it was my bad, when you're passing multiple group of resources you need to split them with a comma, (eg. `pods,services`), whereas named resources need to be split with a space (et. `pod/a pod/b`). I've updated help for `oc adm inspect` to include examples in https://github.com/openshift/oc/pull/621
This should be now green and ready to go.
/hold cancel

collection-scripts/gather
```diff
@@ -5,37 +5,42 @@
...
-resources+=(networks.operator.openshift.io)
+group_resources+=(networks.operator.openshift.io)
 
 # Run the Collection of Resources using inspect
-for resource in "${resources[@]}"; do
-    oc adm inspect --dest-dir must-gather ${resource}
-done
+# running accross all-namespaces for the few "Autoscaler" resouces.
+oc adm inspect --dest-dir must-gather "${named_resources[@]}"
+group_resources_text=$(IFS=, ; echo "${group_resources[*]}")
```
sferich888, line 42: Why do we do this for grouped resources but not named resources? 
Having this here in the conversation would be good for posterity.

PR #140, opened 2020-01-15
network: actually call network gather script

dcbw:
Is this right?

@rcarrillocruz @danwinship @knobunc @wking

### DIFF
--- a/collection-scripts/gather
+++ b/collection-scripts/gather
@@ -38,6 +38,9 @@ for resource in "${resources[@]}"; do
     oc adm inspect --dest-dir must-gather ${resource}
 done
 
+# Gather network Logs (using a supplemental script)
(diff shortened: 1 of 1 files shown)


danwinship:
I thought that the network log script was considered too heavyweight and so it was skipped by default but there was supposed to be some way to run it manually?

wking:
> ... but there was supposed to be some way to run it manually?

With `oc adm must-gather gather_network_logs` or some such?

dcbw:
@danwinship so how are we supposed to run it in CI? Just be on the ball and run it manually?

danwinship:
I don't know. I just remember that `gather_network_logs` gets a bunch of logs on every node, and it was declared that we can't do that in every must-gather run, because in some clusters that would result in way too much data being generated.

PR #142, opened 2020-02-07
Gather Kuryr specific data

MaysaMacedo:
This commit ensures that all Kuryr related data
is gathered.

### DIFF
--- a/collection-scripts/gather_network_logs
+++ b/collection-scripts/gather_network_logs
@@ -1,91 +1,127 @@
 #!/bin/bash
 BASE_COLLECTION_PATH="/must-gather"
-NETWORK_LOG_PATH="${BASE_COLLECTION_PATH}/network_logs/"
+NETWORK_LOG_PATH="${BASE_COLLECTION_PATH}/network_logs"
 
 mkdir -p ${NETWORK_LOG_PATH}/
+
+function gather_openshiftsdn_nodes_data {
+  for NODE in $NODES; do
+      SDN_POD=$(oc -n openshift-sdn get pods --no-headers -o custom-columns=":metadata.name" --field-selector spec.nodeName=$NODE -l app=sdn)
+      oc -n openshift-sdn exec $SDN_POD -- bash -c "iptables-save -c" > ${NETWORK_LOG_PATH}/${NODE}_iptables &
+      PIDS+=($!)
+      oc -n openshift-sdn exec $SDN_POD -- bash -c "ovs-vsctl show" > ${NETWORK_LOG_PATH}/${NODE}_ovs_dump &
+      PIDS+=($!)
+
+      OVS_POD=$(oc -n openshift-sdn get pods --no-headers -o custom-columns=":metadata.name" --field-selector spec.nodeName=$NODE -l app=ovs)
+      oc -n openshift-sdn exec $OVS_POD -- bash -c "cat /var/log/openvswitch/ovs-vswitchd.log" > ${NETWORK_LOG_PATH}/${OVS_POD}_vswitchd_log &
+      PIDS+=($!)
+      oc -n openshift-sdn exec $OVS_POD -- bash -c "cat /var/log/openvswitch/ovsdb-server.log" > ${NETWORK_LOG_PATH}/${OVS_POD}_ovsdb_log &
+      PIDS+=($!)
+  done
+}
+
+function gather_kuryr_nodes_data {
+  for NODE in $NODES; do
+      CNI_POD=$(oc -n openshift-kuryr get pods --no-headers -o custom-columns=":metadata.name" --field-selector spec.nodeName=$NODE -l app=kuryr-cni)
+      oc exec -n openshift-kuryr $CNI_POD -- bash -c 'for pid in $(find /host_proc/[1-9]*/ns/net | cut -d/ -f3); do\
+          echo "Namespace: $pid";\
+          nsenter --net=/host_proc/$pid/ns/net ip -d addr;\
+          done' > ${NETWORK_LOG_PATH}/${CNI_POD}_interfaces & PIDS+=($!)
+  done
+}
+
+function gather_ovn_kubernetes_nodes_data {
+  for NODE in $NODES; do
+      OVNKUBE_NODE_POD=$(oc -n openshift-ovn-kubernetes get pods --no-headers -o custom-columns=":metadata.name" --field-selector spec.nodeName=$NODE -l app=ovnkube-node)
+      oc -n openshift-ovn-kubernetes exec -c ovnkube-node $OVNKUBE_NODE_POD -- bash -c "iptables-save -c" > ${NETWORK_LOG_PATH}/${OVNKUBE_NODE_POD}_iptables &
+      PIDS+=($!)
+      oc -n openshift-ovn-kubernetes exec -c ovnkube-node $OVNKUBE_NODE_POD -- bash -c "ip addr" > ${NETWORK_LOG_PATH}/${OVNKUBE_NODE_POD}_ip_addr &
+      PIDS+=($!)
+      oc -n openshift-ovn-kubernetes exec -c ovnkube-node $OVNKUBE_NODE_POD -- bash -c "ip route" > ${NETWORK_LOG_PATH}/${OVNKUBE_NODE_POD}_ip_route &
+      PIDS+=($!)
+      oc -n openshift-ovn-kubernetes exec -c ovnkube-node $OVNKUBE_NODE_POD -- bash -c "ip -6 route" > ${NETWORK_LOG_PATH}/${OVNKUBE_NODE_POD}_ip_6_route &
+      PIDS+=($!)
+      oc -n openshift-ovn-kubernetes exec -c ovnkube-node $OVNKUBE_NODE_POD -- bash -c \
+      "ovs-ofctl dump-ports-desc br-int" > ${NETWORK_LOG_PATH}/${OVNKUBE_NODE_POD}_ovs_ofctl_dump_ports_br_int &
+      PIDS+=($!)
+      oc -n openshift-ovn-kubernetes exec -c ovnkube-node $OVNKUBE_NODE_POD -- bash -c "ovs-ofctl dump-flows br-int" > \
+      ${NETWORK_LOG_PATH}/${OVNKUBE_NODE_POD}_ovs_ofctl_dump_flows_br_int &
+      PIDS+=($!)
+      oc -n openshift-ovn-kubernetes exec -c ovnkube-node $OVNKUBE_NODE_POD -- bash -c \
+      "ovs-ofctl dump-ports-desc br-local" > ${NETWORK_LOG_PATH}/${OVNKUBE_NODE_POD}_ovs_ofctl_dump_ports_br_local &
+      PIDS+=($!)
+      oc -n openshift-ovn-kubernetes exec -c ovnkube-node $OVNKUBE_NODE_POD -- bash -c "ovs-ofctl dump-flows br-local" > \
+      ${NETWORK_LOG_PATH}/${OVNKUBE_NODE_POD}_ovs_ofctl_dump_flows_br_local &
+      PIDS+=($!)
+  done
+}
+
+function gather_ovn_kubernetes_master_data {
+  OVNKUBE_MASTER_IPS=($(oc -n openshift-ovn-kubernetes get pods -l app=ovnkube-master -o=jsonpath='{.items[*].status.podIP}'))
+  OVNKUBE_MASTER_POD=$(oc -n openshift-ovn-kubernetes get pods --no-headers -o custom-columns=':metadata.name' -l app=ovnkube-master | tail -1)
+
+  oc -n openshift-ovn-kubernetes exec -c ovnkube-master $OVNKUBE_MASTER_POD -- bash -c \
+  "ovn-nbctl --db=ssl:${OVNKUBE_MASTER_IPS[0]}:9641,ssl:${OVNKUBE_MASTER_IPS[1]}:9641,ssl:${OVNKUBE_MASTER_IPS[2]}:9641 \
+  -p /ovn-cert/tls.key -c /ovn-cert/tls.crt -C /ovn-ca/ca-bundle.crt show" > \
+  ${NETWORK_LOG_PATH}/${OVNKUBE_MASTER_POD}_ovn_nbctl_show &
+  PIDS+=($!)
+  oc -n openshift-ovn-kubernetes exec -c ovnkube-master $OVNKUBE_MASTER_POD -- bash -c \
+  "ovn-nbctl --db=ssl:${OVNKUBE_MASTER_IPS[0]}:9641,ssl:${OVNKUBE_MASTER_IPS[1]}:9641,ssl:${OVNKUBE_MASTER_IPS[2]}:9641 \
+  -p /ovn-cert/tls.key -c /ovn-cert/tls.crt -C /ovn-ca/ca-bundle.crt list Logical_Switch_Port" > \
+  ${NETWORK_LOG_PATH}/${OVNKUBE_MASTER_POD}_ovn_nbctl_list_lsp &
+  PIDS+=($!)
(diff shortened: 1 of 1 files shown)


collection-scripts/gather_network_logs
```diff
@@ -26,6 +26,38 @@ if [ "$NETWORK_TYPE" == "openshiftsdn" ]; then
...
+      ${NETWORK_LOG_PATH}/${CONTROLLER_POD}_log &
+      PIDS+=($!)
+      oc -n openshift-kuryr logs $CONTROLLER_POD --previous > \
+      ${NETWORK_LOG_PATH}/${CONTROLLER_POD}_previous_log &
+      PIDS+=($!)
+
+      NODES="${@:-$(oc get nodes --no-headers -o custom-columns=':metadata.name')}"
+      for NODE in $NODES; do
+          CNI_POD=$(oc -n openshift-kuryr get pods --no-headers -o custom-columns=":metadata.name" --field-selector spec.nodeName=$NODE -l app=kuryr-cni)
+          oc -n openshift-kuryr logs $CNI_POD > ${NETWORK_LOG_PATH}/${CNI_POD}_log &
```
luis5tb: do we want the previous one here too?

MaysaMacedo: For sure. Thanks!

collection-scripts/gather_network_logs
```diff
@@ -26,6 +26,46 @@ if [ "$NETWORK_TYPE" == "openshiftsdn" ]; then
         oc -n openshift-sdn exec $OVS_POD -- bash -c "cat /var/log/openvswitch/ovsdb-server.log" > ${NETWORK_LOG_PATH}/${OVS_POD}_ovsdb_log &
         PIDS+=($!)
     done
+elif [ "$NETWORK_TYPE" == "kuryr" ];
```
gryf: `elif [ "$NETWORK_TYPE" == "kuryr" ]; then` ?

collection-scripts/gather_network_logs
```diff
@@ -26,6 +26,46 @@ if [ "$NETWORK_TYPE" == "openshiftsdn" ]; then
         oc -n openshift-sdn exec $OVS_POD -- bash -c "cat /var/log/openvswitch/ovsdb-server.log" > ${NETWORK_LOG_PATH}/${OVS_POD}_ovsdb_log &
         PIDS+=($!)
     done
+elif [ "$NETWORK_TYPE" == "kuryr" ]; then
+      CONTROLLER_POD=$(oc -n openshift-kuryr get pods --no-headers -o custom-columns=':metadata.name' -l app=kuryr-controller | tail -1)
```
dulek: Instead of ` --no-headers -o custom-columns=':metadata.name'` you can do `-o name`, but it'll get you `pod/` part too. Which still works with `oc logs` so it might be fine. No pushing here, whatever you prefer.

collection-scripts/gather_network_logs
```diff
@@ -26,6 +26,46 @@ if [ "$NETWORK_TYPE" == "openshiftsdn" ]; then
...
+          RESTART_COUNT="${@:-$(oc get po $CNI_POD -n openshift-kuryr -o jsonpath={.status.containerStatuses[*].restartCount})}"
+          if [ "$RESTART_COUNT" -ne "0" ]; then
+              oc -n openshift-kuryr logs $CNI_POD --previous > ${NETWORK_LOG_PATH}/${CNI_POD}_previous_log &
+              PIDS+=($!)
+          fi
+          oc -n openshift-kuryr exec $CNI_POD -- bash -c "ip -d addr" > ${NETWORK_LOG_PATH}/${CNI_POD}_interfaces &
+          PIDS+=($!)
+      done
+
+      oc get pods -A -o wide > ${NETWORK_LOG_PATH}/get_pods & PIDS+=($!)
```
dulek: Do we want Kuryrnets too?

collection-scripts/gather_network_logs
```diff
@@ -26,6 +26,46 @@ if [ "$NETWORK_TYPE" == "openshiftsdn" ]; then
...
+      for NODE in $NODES; do
+          CNI_POD=$(oc -n openshift-kuryr get pods --no-headers -o custom-columns=":metadata.name" --field-selector spec.nodeName=$NODE -l app=kuryr-cni)
+          oc -n openshift-kuryr logs $CNI_POD > ${NETWORK_LOG_PATH}/${CNI_POD}_log &
+          PIDS+=($!)
+          RESTART_COUNT="${@:-$(oc get po $CNI_POD -n openshift-kuryr -o jsonpath={.status.containerStatuses[*].restartCount})}"
+          if [ "$RESTART_COUNT" -ne "0" ]; then
+              oc -n openshift-kuryr logs $CNI_POD --previous > ${NETWORK_LOG_PATH}/${CNI_POD}_previous_log &
+              PIDS+=($!)
+          fi
+          oc -n openshift-kuryr exec $CNI_POD -- bash -c "ip -d addr" > ${NETWORK_LOG_PATH}/${CNI_POD}_interfaces &
```
dulek: Here we also need that info from inside the pod namespaces. Easiest way would be to just do `ip netns` and for each `echo $netns && ip netns exec $netns ip -d addr`.

collection-scripts/gather_network_logs
MaysaMacedo: Thanks! Your suggestion is indeed more clean, but it would break the creation of the log file.

collection-scripts/gather_network_logs
MaysaMacedo: done.

collection-scripts/gather_network_logs
MaysaMacedo: `ip netns` excepts that the namespaces are under` /var/run/netns/ `, which for Kuryr case that is not True, as we don't link them from` /proc/$pid/ns/net`, so the result would be empty. We could also retrieve that info by inspecting the container and `nsenter` on it inside the host, but it's not ideal.

collection-scripts/gather_network_logs
```diff
@@ -26,6 +26,52 @@ if [ "$NETWORK_TYPE" == "openshiftsdn" ]; then
...
+              nsenter --net=/host_proc/$pid/ns/net ip -d addr;\
+              done' > ${NETWORK_LOG_PATH}/${CNI_POD}_interfaces & PIDS+=($!)
+      done
+
+      oc get pods -A -o wide > ${NETWORK_LOG_PATH}/get_pods & PIDS+=($!)
+      openstack port list > ${NETWORK_LOG_PATH}/openstack_port_list & PIDS+=($!)
+
+      oc get kuryrnets -A > ${NETWORK_LOG_PATH}/get_kuryrnets & PIDS+=($!)
+      openstack network list > ${NETWORK_LOG_PATH}/openstack_network_list & PIDS+=($!)
+      openstack subnet list > ${NETWORK_LOG_PATH}/openstack_subnet_list & PIDS+=($!)
```
luis5tb: can you also save the router info (to see if the subnets are connected to it)

luis5tb:
LGTM, just a note about couple of extra information that we may want to grab, router info and quota info

dulek:
Luis comments makes sense, besides that LGTM.

collection-scripts/gather_network_logs
```diff
@@ -26,6 +26,59 @@ if [ "$NETWORK_TYPE" == "openshiftsdn" ]; then
...
+          CNI_POD=$(oc -n openshift-kuryr get pods --no-headers -o custom-columns=":metadata.name" --field-selector spec.nodeName=$NODE -l app=kuryr-cni)
+          oc -n openshift-kuryr logs $CNI_POD > ${NETWORK_LOG_PATH}/${CNI_POD}_log &
+          PIDS+=($!)
+          RESTART_COUNT="${@:-$(oc get po $CNI_POD -n openshift-kuryr -o jsonpath={.status.containerStatuses[*].restartCount})}"
+          if [ "$RESTART_COUNT" -ne "0" ]; then
+              oc -n openshift-kuryr logs $CNI_POD --previous > ${NETWORK_LOG_PATH}/${CNI_POD}_previous_log &
+              PIDS+=($!)
+          fi
+          oc exec -n openshift-kuryr $CNI_POD -- bash -c 'for pid in $(find /host_proc/[1-9]*/ns/net | cut -d/ -f3); do\
+              nsenter --net=/host_proc/$pid/ns/net ip -d addr;\
```
dulek: Ah, it's worth to print the namespace name so we can find the corresponding request in kuryr-daemon logs.

collection-scripts/gather_network_logs
```diff
@@ -26,6 +26,60 @@ if [ "$NETWORK_TYPE" == "openshiftsdn" ]; then
...
+      ${NETWORK_LOG_PATH}/${CONTROLLER_POD}_log &
+      PIDS+=($!)
+
+      CONTROLLER_RESTART_COUNT="${@:-$(oc get po $CONTROLLER_POD -n openshift-kuryr -o jsonpath={.status.containerStatuses[*].restartCount})}"
+      if [ "$CONTROLLER_RESTART_COUNT" -ne "0" ]; then
+          oc -n openshift-kuryr logs $CONTROLLER_POD --previous > ${NETWORK_LOG_PATH}/${CONTROLLER_POD}_previous_log &
+          PIDS+=($!)
+      fi
+
+      NODES="${@:-$(oc get nodes --no-headers -o custom-columns=':metadata.name')}"
```
sferich888: We use this concept in several places; should we try and functionalize it? 
It seems like we're repeating bash code here (where it's possible we will get out of alignment in the various places where we want to restrict collection only to specific nodes). Should we look at handling this differently, more genericly?

collection-scripts/gather_network_logs
```diff
@@ -26,6 +26,60 @@ if [ "$NETWORK_TYPE" == "openshiftsdn" ]; then
...
+              oc -n openshift-kuryr logs $CNI_POD --previous > ${NETWORK_LOG_PATH}/${CNI_POD}_previous_log &
+              PIDS+=($!)
+          fi
+          oc exec -n openshift-kuryr $CNI_POD -- bash -c 'for pid in $(find /host_proc/[1-9]*/ns/net | cut -d/ -f3); do\
+              echo "Namespace: $pid";\
+              nsenter --net=/host_proc/$pid/ns/net ip -d addr;\
+              done' > ${NETWORK_LOG_PATH}/${CNI_POD}_interfaces & PIDS+=($!)
+      done
+
+      oc get pods -A -o wide > ${NETWORK_LOG_PATH}/get_pods & PIDS+=($!)
```
sferich888: We will likely already collect this output with other must-gather runs; I am not sure why we would re-collect the pods list here.

collection-scripts/gather_network_logs
```diff
@@ -26,6 +26,60 @@ if [ "$NETWORK_TYPE" == "openshiftsdn" ]; then
...
+      openstack port list > ${NETWORK_LOG_PATH}/openstack_port_list & PIDS+=($!)
```
sferich888: do these `openstack` commands work in the pod? 
What updates are we doing to the default cli-image that provides this command?

collection-scripts/gather_network_logs
MaysaMacedo: When Kuryr-Kubernetes is used we only collect it here. It could be useful to verify if the pods are  stuck on `containerCreating`

collection-scripts/gather_network_logs
MaysaMacedo: the openstack command does not work inside the containers. We shouldn't assume the cloud operator has it enabled. So, I should probably add a note about how to install it.

sferich888: I think we need to add the tools/package dependency to our image if we plan to use this command.

collection-scripts/gather_network_logs
sferich888: Let's let the other 'collectors' collect this, and if you want this you simply need to supply `gather; gather_network_logs` as the 'command' to must-gather call; thus getting you all this information in one archive.

collection-scripts/gather_network_logs
sferich888: @dcbw and @knobunc your teams may want to look at what is being done here; just to make sure we're not possibly introducing a bug/feature regression.

collection-scripts/gather_network_logs
MaysaMacedo: Seems the `gather` script does list pods running on namespaces created by the user, only the OpenShift ones created by the installer. As it lists the current and previous logs of the Kuryr pods, I'm removing the logs we're fetching from those pods.

MaysaMacedo:
/hold

Working on removing the need for openstackcli package

MaysaMacedo:
/hold cancel
OpenStack cli not needed anymore.

collection-scripts/gather_network_logs
```diff
@@ -1,91 +1,125 @@
...
+  PIDS+=($!)
+}
+
+function gather_kuryr_data {
+  CONTROLLER_POD=$(oc -n openshift-kuryr get pods --no-headers -o custom-columns=":metadata.name" -l app=kuryr-controller)
+  oc -n openshift-kuryr exec $CONTROLLER_POD -- bash -c \
+  'kuryr-gather-openstack-data --config-dir /etc/kuryr' > get_openstack_data &
+  PIDS+=($!)
+  oc get pods -A -o wide --show-labels > ${NETWORK_LOG_PATH}/get_pods & PIDS+=($!)
+  oc get kuryrnets -A > ${NETWORK_LOG_PATH}/get_kuryrnets & PIDS+=($!)
```
dulek: Had we lost kuryrnetpolicy somewhere?

MaysaMacedo:
> @MaysaMacedo Looks good. I like the functions. PTAL PR 145. I have made some changes in the base capture. Maybe we could combine these PRs?

@pecameron Thanks for taking a look! Sure, we can combine. How would you prefer to combine them?

collection-scripts/gather_network_logs
pecameron: @sferich888 could you take a look at PR145? We have been working in the same space.

pecameron:
@MaysaMacedo I don't have a preference, do you? We could merge one and edit the other. I tested by bringing up ovn and sdn clusters and running the script (PR145) so they gather expected data. I have been looking into how they get used. This is my first time working in this repo, so I am still trying to understand how it all works.

MaysaMacedo:
@pecameron No preferences either. I guess we can combine your work into this PR. What do you think?

dcbw:
@pecameron @MaysaMacedo slight prference for merging this one and then Phil updating his to match. I do like having separate functions for each network plugin type.

MaysaMacedo:
@sferich888 This PR is ready for another look, as it was checked by folks from other SDN teams.

collection-scripts/gather_network_logs
```diff
@@ -1,91 +1,127 @@
...
+  oc -n openshift-ovn-kubernetes exec -c ovnkube-master $OVNKUBE_MASTER_POD -- bash -c \
+  "ovn-sbctl --db=ssl:${OVNKUBE_MASTER_IPS[0]}:9642,ssl:${OVNKUBE_MASTER_IPS[1]}:9642,ssl:${OVNKUBE_MASTER_IPS[2]}:9642 \
+  -p /ovn-cert/tls.key -c /ovn-cert/tls.crt -C /ovn-ca/ca-bundle.crt show" > \
+  ${NETWORK_LOG_PATH}/${OVNKUBE_MASTER_POD}_ovn_sbctl_show &
+  PIDS+=($!)
+}
+
+function gather_kuryr_data {
+  CONTROLLER_POD=$(oc -n openshift-kuryr get pods --no-headers -o custom-columns=":metadata.name" -l app=kuryr-controller)
+  oc -n openshift-kuryr exec $CONTROLLER_POD -- bash -c \
```
sferich888: If your having networking issues is this risky?

MaysaMacedo: Thanks for the review!

Good point. But, to avoid having to rely on the `openstackcli` being available on the host this was the only alternative. Also, we tried to follow the same pattern other SDNs are doing (executing commands from the network pods).

PR #150, opened 2020-04-14
Bug 1829113: Add openshift-kni-infra namespace to gather

cybertron:
openshift-kni-infra is the namespace where baremetal service pods
run. Currently it is not collected by must-gather which can make it
difficult to debug baremetal deployment issues. This adds the
namespace to the list of namespaces from which to collect logs.

### DIFF
--- a/collection-scripts/gather
+++ b/collection-scripts/gather
@@ -21,7 +21,7 @@ resources+=(certificatesigningrequests)
 resources+=(nodes machineconfigs machineconfigpools)
 
 # Namespaces/Project Resources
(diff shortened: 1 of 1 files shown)


sferich888:
/lgtm

I'm not sure why this is needed but it's fine to add, generally we want to add resources (types) so that if they exist in other places we get them in those namespaces too.

cybertron:
These are static pods deployed on each node by machine-config-operator. Is there a better way to include that sort of thing in must-gather?

runcom:
@cybertron please update the title here to reference https://bugzilla.redhat.com/show_bug.cgi?id=1829113

PR #154, opened 2020-05-18
Gather etcd information from must-gather

palonsoro:
One of the most important things to check while getting information from a cluster is to check the cluster etcd health. This cannot be achieved by just getting some pod logs or kubernetes resources, but needs its own set of commands to be run in etcd pods.

In this pull request, I am adding a `gather_etcd` script to gather that information and I also made it to be run by default from main `gather` script.

Having this information on every must-gather would be of great help for support.

I'll be happy to help with any change or improvement required on this. And thanks in advance.

### DIFF
--- a/collection-scripts/gather
+++ b/collection-scripts/gather
@@ -43,3 +43,6 @@ done
 
 # Gather Service Logs (using a suplamental Script); Scoped to Masters.
 /usr/bin/gather_service_logs master
+
+# Gather etcd information
+/usr/bin/gather_etcd
--- a/collection-scripts/gather
+++ b/collection-scripts/gather
@@ -38,11 +38,11 @@ for resource in "${resources[@]}"; do
     oc adm inspect --dest-dir must-gather ${resource}
 done
 
+# Gather etcd information
+/usr/bin/gather_etcd
+
 # Collect System Audit Logs
 /usr/bin/gather_audit_logs
 
 # Gather Service Logs (using a suplamental Script); Scoped to Masters.
 /usr/bin/gather_service_logs master
-
-# Gather etcd information
-/usr/bin/gather_etcd
--- a/collection-scripts/gather_etcd
+++ b/collection-scripts/gather_etcd
@@ -3,15 +3,15 @@
 BASE_COLLECTION_PATH="/must-gather"
 ETCD_LOG_PATH="${BASE_COLLECTION_PATH}/etcd_info"
 
-ETCD_CONTAINER='etcdctl'
+ETCDCTL_CONTAINER='etcdctl'
 
 RUNNING_ETCD_POD=$(oc get pod -n openshift-etcd --no-headers|grep Running| grep -o -m 1 '\S*etcd\S*')
 
 # We cannot rely on ETCDCTL_ENDPOINTS on container because it may contain bootstrap VM
-ETCDCTL_ENDPOINTS=$(oc exec ${RUNNING_ETCD_POD}  -n openshift-etcd -c ${ETCD_CONTAINER} -- /bin/sh -c "etcdctl member list" | awk -F', ' '{printf "%s%s",sep,$5; sep=","}')
+ETCDCTL_ENDPOINTS=$(oc exec ${RUNNING_ETCD_POD}  -n openshift-etcd -c ${ETCDCTL_CONTAINER} -- /bin/sh -c "etcdctl member list" | awk -F', ' '{printf "%s%s",sep,$5; sep=","}')
 
 function ocp4etcdctl {
-    oc -n openshift-etcd exec -c ${ETCD_CONTAINER} ${RUNNING_ETCD_POD} -- env ETCDCTL_ENDPOINTS=${ETCDCTL_ENDPOINTS} etcdctl $@
+    oc -n openshift-etcd exec -c ${ETCDCTL_CONTAINER} ${RUNNING_ETCD_POD} -- env ETCDCTL_ENDPOINTS=${ETCDCTL_ENDPOINTS} etcdctl $@
 }
 
 if [[ -z $RUNNING_ETCD_POD ]];then
@@ -19,7 +19,7 @@ if [[ -z $RUNNING_ETCD_POD ]];then
     exit 1
 fi
 
-echo "Getting information from pod \"${RUNNING_ETCD_POD}\", container \"${ETCD_CONTAINER}\""
+echo "Getting information from pod \"${RUNNING_ETCD_POD}\", container \"${ETCDCTL_CONTAINER}\""
 echo "Using endpoints: ${ETCDCTL_ENDPOINTS}"
 
 mkdir -p ${ETCD_LOG_PATH}
--- a/collection-scripts/gather_etcd
+++ b/collection-scripts/gather_etcd
@@ -8,10 +8,10 @@ ETCDCTL_CONTAINER='etcdctl'
 RUNNING_ETCD_POD=$(oc get pod -n openshift-etcd --no-headers|grep Running| grep -o -m 1 '\S*etcd\S*')
 
 # We cannot rely on ETCDCTL_ENDPOINTS on container because it may contain bootstrap VM
-ETCDCTL_ENDPOINTS=$(oc exec ${RUNNING_ETCD_POD}  -n openshift-etcd -c ${ETCDCTL_CONTAINER} -- /bin/sh -c "etcdctl member list" | awk -F', ' '{printf "%s%s",sep,$5; sep=","}')
+ETCDCTL_ENDPOINTS=$(oc exec "${RUNNING_ETCD_POD}"  -n openshift-etcd -c ${ETCDCTL_CONTAINER} -- /bin/sh -c "etcdctl member list" | awk -F', ' '{printf "%s%s",sep,$5; sep=","}')
 
 function ocp4etcdctl {
-    oc -n openshift-etcd exec -c ${ETCDCTL_CONTAINER} ${RUNNING_ETCD_POD} -- env ETCDCTL_ENDPOINTS=${ETCDCTL_ENDPOINTS} etcdctl $@
+    oc -n openshift-etcd exec -c "${ETCDCTL_CONTAINER}" "${RUNNING_ETCD_POD}" -- env "ETCDCTL_ENDPOINTS=${ETCDCTL_ENDPOINTS}" etcdctl "$@"
 }
 
 if [[ -z $RUNNING_ETCD_POD ]];then
@@ -57,5 +57,5 @@ ocp4etcdctl alarm list -w json > ${ETCD_LOG_PATH}/alarm_list.json &
 PIDS+=($!)
 
 echo "INFO: Waiting for etcd info collection to complete ..."
-wait ${PIDS[@]}
+wait "${PIDS[@]}"
 echo "INFO: Done collecting etcd information"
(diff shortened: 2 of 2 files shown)


collection-scripts/gather
```diff
@@ -43,3 +43,6 @@ done
 
 # Gather Service Logs (using a suplamental Script); Scoped to Masters.
 /usr/bin/gather_service_logs master
+
+# Gather etcd information
+/usr/bin/gather_etcd
```
sferich888, line 48: Should this follow `gather_resources` (line 39)?

collection-scripts/gather_etcd
```diff
@@ -0,0 +1,49 @@
...
+}
+
+if [[ -z $RUNNING_ETCD_POD ]];then
+    echo "ERROR: No running etcd pods found" >&2
+    exit 1
+fi
+
+mkdir -p ${ETCD_LOG_PATH}
+
+echo "INFO: Starting getting etcd information"
```
sferich888, line 19: Should this print the name of `RUNNING_ETCD_POD`?

collection-scripts/gather_etcd
```diff
@@ -0,0 +1,49 @@
...
+echo "INFO: Getting etcdctl endpoint health"
+ocp4etcdctl endpoint health -w table > ${ETCD_LOG_PATH}/endpoint_health.txt
+ocp4etcdctl endpoint health -w json > ${ETCD_LOG_PATH}/endpoint_health.json
+
+# alarm list
+echo "INFO: Getting etcdctl alarm list"
+ocp4etcdctl alarm list > ${ETCD_LOG_PATH}/alarm_list.txt
+ocp4etcdctl alarm list -w json > ${ETCD_LOG_PATH}/alarm_list.json
+
+# List all etcd keys
```
sferich888, line 41: This seems like it might be a security issue, in some 'rare' circumstances. 
Should we have a toggle for this? It also seems like a `costly` operation to preform. How much data is this going to pull back? @smarterclayton what is the averaged etcd count size shown from telemetry?

collection-scripts/gather_etcd
```diff
@@ -0,0 +1,49 @@
...
+if [[ -z $RUNNING_ETCD_POD ]];then
+    echo "ERROR: No running etcd pods found" >&2
+    exit 1
+fi
+
+mkdir -p ${ETCD_LOG_PATH}
+
+echo "INFO: Starting getting etcd information"
+
+# member list
```
sferich888, line 21: Should we run all of these in parallel?

collection-scripts/gather_etcd
palonsoro: My original intention with getting all the keys was to contrast whether users may claim something has been lost and/or some leftover from an unsupported version was present (due to a bug that prevented some api version migration, for example).

However, I did not thought about the security implications, so maybe I can just remove the line

collection-scripts/gather_etcd
palonsoro: Good idea! I'll implement that.

collection-scripts/gather_etcd
palonsoro: All those commands run very quickly (except for the check perf), so this is why I didn't make this parallel

collection-scripts/gather
palonsoro: I don't see a big difference so I can just move it up.

collection-scripts/gather_etcd
sferich888: It's still ~8 `oc exec` calls. Depending on server load this could be cumbersome. Threading this lets all of this run at the same time and simply stop when completed, vs each call getting hung up and stalling the collection.

palonsoro: Fair enough. I'll use the same parallelism scheme used in other gather scripts

palonsoro:
Changes implemented:
- Moved execution to the place requested
- Print etcd pod name as requested
- Removed key listing as suggested in order to prevent security issues
- `ETCDCTL_ENDPOINTS` computed from `member list` because the ones at the container may include bootstrap VM.

TO-DO:
- Parallelize execution

collection-scripts/gather_etcd
```diff
@@ -0,0 +1,53 @@
...
+ocp4etcdctl endpoint health -w json > ${ETCD_LOG_PATH}/endpoint_health.json
+
+# alarm list
+echo "INFO: Getting etcdctl alarm list"
+ocp4etcdctl alarm list > ${ETCD_LOG_PATH}/alarm_list.txt
+ocp4etcdctl alarm list -w json > ${ETCD_LOG_PATH}/alarm_list.json
+
+# Check perf
+echo "INFO: Performing etcdctl check perf"
+ocp4etcdctl check perf --load="s" > ${ETCD_LOG_PATH}/etcd_check_perf.txt
```
hexfusion, line 51: please do not do this, we monitor perf related metrics in Prometheus. If this command fails it will leave many keys in KVS. We are also capturing metrics through etcd events.

collection-scripts/gather_etcd
```diff
@@ -0,0 +1,53 @@
+#!/bin/bash
+
+BASE_COLLECTION_PATH="/must-gather"
+ETCD_LOG_PATH="${BASE_COLLECTION_PATH}/etcd_info"
+
+ETCD_CONTAINER='etcdctl'
```
hexfusion, line 6: can we clarify this ENV as `ETCDCTL_CONTAINER` so that I don't have a mini heart attack every time I see this code?

collection-scripts/gather_etcd
palonsoro: Understood. I'll remove it.

collection-scripts/gather_etcd
palonsoro: Ok. No problem

palonsoro:
All requested changes implemented.

Thanks for all the feedback.

collection-scripts/gather_etcd
```diff
@@ -0,0 +1,61 @@
+#!/bin/bash
+
```
wking, line 2: This probably needs guards like:

```sh
set -o nounset
set -o errexit
set -o pipefail
```

so that it doesn't blindly grind on if a step like calculating `RUNNING_ETCD_POD` fails.

collection-scripts/gather_etcd
```diff
@@ -0,0 +1,61 @@
...
+BASE_COLLECTION_PATH="/must-gather"
+ETCD_LOG_PATH="${BASE_COLLECTION_PATH}/etcd_info"
+
+ETCDCTL_CONTAINER='etcdctl'
+
+RUNNING_ETCD_POD=$(oc get pod -n openshift-etcd --no-headers|grep Running| grep -o -m 1 '\S*etcd\S*')
+
+# We cannot rely on ETCDCTL_ENDPOINTS on container because it may contain bootstrap VM
+ETCDCTL_ENDPOINTS=$(oc exec ${RUNNING_ETCD_POD}  -n openshift-etcd -c ${ETCDCTL_CONTAINER} -- /bin/sh -c "etcdctl member list" | awk -F', ' '{printf "%s%s",sep,$5; sep=","}')
```
wking, line 11: There are many missing quotes here, and through the rest of this script, that would fail ShellCheck:

```console
$ shellcheck collection-scripts/gather_etcd 

In collection-scripts/gather_etcd line 11:
ETCDCTL_ENDPOINTS=$(oc exec ${RUNNING_ETCD_POD}  -n openshift-etcd -c ${ETCDCTL_CONTAINER} -- /bin/sh -c "etcdctl member list" | awk -F', ' '{printf "%s%s",sep,$5; sep=","}')
                            ^-- SC2086: Double quote to prevent globbing and word splitting.


In collection-scripts/gather_etcd line 14:
    oc -n openshift-etcd exec -c ${ETCDCTL_CONTAINER} ${RUNNING_ETCD_POD} -- env ETCDCTL_ENDPOINTS=${ETCDCTL_ENDPOINTS} etcdctl $@
                                                      ^-- SC2086: Double quote to prevent globbing and word splitting.
                                                                                                   ^-- SC2086: Double quote to prevent globbing and word splitting.
                                                                                                                                ^-- SC2068: Double quote array expansions, otherwise they're like $* and break on spaces.


In collection-scripts/gather_etcd line 60:
wait ${PIDS[@]}
     ^-- SC2068: Double quote array expansions, otherwise they're like $* and break on spaces.

```

I think we should grow a ShellCheck job as a preflight for this repo.

collection-scripts/gather_etcd
wking: This still doesn't include `RUNNING_ETCD_POD`.  Did you forget to commit and/or push?

collection-scripts/gather_etcd
```diff
@@ -0,0 +1,61 @@
...
+
+mkdir -p ${ETCD_LOG_PATH}
+
+echo "INFO: Starting getting etcd information"
+
+PIDS=()
+
+# member list
+echo "INFO: Getting etcdctl member list"
+ocp4etcdctl member list -w table > ${ETCD_LOG_PATH}/member_list.txt &
```
wking, line 33: Table output is convenient, but the JSON output should be sufficient, right?  I'd rather not bloat the must-gathers with redundant information, since folks can always write consuming code that pretty-prints the JSON if they want convenient inspection.

collection-scripts/gather_etcd
palonsoro: I am ok with `nounset` and `pipefail`. `errexit` would only be appropriate for `RUNNING_ETCD_POD` and `ETCDCTL_ENDPOINTS` calculation, because if one of the `ocp4etcdctl` commands return a non-zero exit code, the other ones should be tried.

It would be good to evaluate if any of the other files needs these guards, because none of them use them.

collection-scripts/gather_etcd
palonsoro: Let me fix that on `etcd_gather` script.

And I totally agree there should be an automated `ShellCheck` test, because none of the collection scripts completely passes it. It seems that there are quotes missing in many places. But I think I should just fix the etcd script at this PR, and fixing the other scripts should be another PR.

collection-scripts/gather_etcd
palonsoro: Right: I pasted both JSON and table outputs because one of the most challenging aspects of dealing with must-gather is having all the information in machine-readable format only. However, I understand the need of no redundancy, so let me fix it.

collection-scripts/gather_etcd
palonsoro: It was already added but in another log statement below. Please see: https://github.com/palonsoro/must-gather/blob/gather_etcd/collection-scripts/gather_etcd#L22

palonsoro:
All the points outlined so far covered. Please let me know about any other feedback. And thanks everybody for the ideas.

collection-scripts/gather_etcd
sferich888: Yea I still don't see where we are 'printing' the pod name. Can we add that here or to the items you reference.

sferich888:
I have one minor nit to address but other than that it's. 
/lgtm 

@deads2k do we need a valid bug for this like tide says? Or is a Jira or something feature related acceptable?

collection-scripts/gather_etcd
palonsoro: Hello, sorry the line number in my comment above is no longer correct due to changes happened after it.

Please look now at at: https://github.com/palonsoro/must-gather/blob/gather_etcd/collection-scripts/gather_etcd#L27

PR #156, opened 2020-05-27
Trace ovn processing in cluster

pecameron:
This processes log files for namespace openshift-ovn-kubernetes.
It also captures the master and worker journals for kublet and crio.
It:

- formats time stamps into a common format
- adds a prefix to each log line consisting of:
- time stamp
- node name
- pod name
- container name

Extracts the log and previous log for each continer
- applies the prefix to each line
- combines all resulting files and
- sorts the results into the trace
- deletes work files.

SDN-975 - Logging: add must-gather networking-gather-trace script
https://issues.redhat.com/browse/SDN-975

Signed-off-by: Phil Cameron <pcameron@redhat.com>

### DIFF
--- a/collection-scripts/gather_network_ovn_trace
+++ b/collection-scripts/gather_network_ovn_trace
@@ -1,0 +1,146 @@
+#!/bin/bash
+BASE_COLLECTION_PATH="must-gather"
+NETWORK_TRACE_PATH=${OUT:-"${BASE_COLLECTION_PATH}/network_trace"}
+
+mkdir -p ${NETWORK_TRACE_PATH}/
+
+# Convert the date and add the prefix
+function convert_log_timestamp {
+  grep -e "^I" -e "^E" -e "^W" -e "^N" | \
+  sed 's/ /-/;s/^[IEWDN]/& /' | \
+  gawk '{ print $2  " | "  $0; }' | \
+  sed "s/|/ ${1} ${2} ${3} | /" | \
+  sed '/^[+a-zA-Z/]/d'
+}
+
+function convert_ovs_timestamp {
+  sed 's/2020-//;s/-//;s/T/-/;s/Z//' | \
+  sed "s/|/ ${1} ${2} ${3} | /" | \
+  sed '/^[+a-zA-Z/]/d'
+}
+
+# Collect the log files
+function gather_ovn_kubernetes_data {
+  for NODE in ${NODES}; do
+    OVNKUBE_NODE_POD=$(oc -n openshift-ovn-kubernetes \
+      get pods --no-headers -o custom-columns=":metadata.name" \
+      --field-selector spec.nodeName=${NODE} -l app=ovnkube-node)
+
+
+    OVNKUBE_MASTER=$(oc -n openshift-ovn-kubernetes \
+      get pods --no-headers -o custom-columns=':metadata.name' \
+      --field-selector spec.nodeName=${NODE} -l app=ovnkube-master)
+
+    OVS_POD=$(oc -n openshift-ovn-kubernetes \
+      get pods --no-headers -o custom-columns=':metadata.name' \
+      --field-selector spec.nodeName=${NODE} -l app=ovs-node)
+
+    echo "NODE $NODE    OVNKUBE_NODE_POD $OVNKUBE_NODE_POD     OVS_POD $OVS_POD     OVNKUBE_MASTER $OVNKUBE_MASTER" 
+
+    oc -n openshift-ovn-kubernetes logs ${OVS_POD} 2>/dev/null | \
+      convert_ovs_timestamp "${NODE}" "${OVS_POD}" "ovs-daemons" \
+      > ${NETWORK_TRACE_PATH}/${NODE}_${OVS_POD}_ovs-daemons.log &
+    PIDS+=($!)
+
+    oc -n openshift-ovn-kubernetes logs ${OVS_POD} -p 2>/dev/null | \
+      convert_ovs_timestamp "${NODE}" "${OVS_POD}" "ovs-daemons" \
+      > ${NETWORK_TRACE_PATH}/${NODE}_${OVS_POD}_ovs-daemons-previous.log &
+    PIDS+=($!)
+
+    oc -n openshift-ovn-kubernetes logs ${OVNKUBE_NODE_POD} -c ovn-controller 2>/dev/null | \
+      convert_ovs_timestamp "${NODE}" "${OVNKUBE_NODE_POD}" "ovn-controller" \
+      > ${NETWORK_TRACE_PATH}/${NODE}_${OVNKUBE_NODE_POD}_ovn-controller.log &
+    PIDS+=($!)
+
+    oc -n openshift-ovn-kubernetes logs ${OVNKUBE_NODE_POD} -c ovn-controller -p 2>/dev/null | \
+      convert_ovs_timestamp "${NODE}" "${OVNKUBE_NODE_POD}" "ovn-controller" \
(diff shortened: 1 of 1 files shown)


collection-scripts/gather_network_trace
```diff
@@ -0,0 +1,146 @@
...
+   done
+}
+
+function convert_journal_timestamp {
+  sed 's/Jan /01/; s/Feb /02/; s/Mar /03/; s/Apr /04/; s/May /05/; s/Jun /06/;
+  s/Jul /07/; s/Aug /08/; s/Sep /09/; s/Oct /10/; s/Nov /11/; s/Dec /12/;
+  s/ /-/' | sed '/^---/d'
+}
+
+function gather_service_data {
```
sferich888, line 103: We already have scripts, in this repo that collect this why not simply call those?

collection-scripts/gather_network_trace
```diff
@@ -0,0 +1,146 @@
...
+}
+
+function convert_ovs_timestamp {
+  sed 's/2020-//;s/-//;s/T/-/;s/Z//' | \
+  sed "s/|/ ${1} ${2} ${3} | /" | \
+  sed '/^[+a-zA-Z/]/d'
+}
+
+# Collect the log files
+function gather_ovn_kubernetes_data {
```
sferich888, line 23: Within the gather_networking script we have a block where this could be run, why add an extra script when this could be added to that file?

collection-scripts/gather_network_trace
pecameron: There is a post process step that converts the date to a common format.

collection-scripts/gather_network_trace
pecameron: There is post processing that converts the date to a common format and adds the node, pod, and container as a prefix to the line.

pecameron: Ultimately all logs are catted together and sorted bu the prefix to get the trace.

pecameron:
@sferich888 The e2e-aws has been pending  overnight. Is there some way I can this 
to run?

collection-scripts/gather_network_trace
pecameron: A second point is we need to know which node a pod is running on, the ovs/ovn/ovn-master are tightly coupled. Some connections are unix sockets so we need to know which container is talking to which container on a node. When raft selects a new master we need to make sure the corresponding containers are running properly.

pecameron: With the basic trace we can add lines to the various logs that help is tracking progress of operations. For example noting when an item is created, used, and destroyed. We can find races using this and we can find out how long things take especially when several containers are involved. For example when an IP is assigned to a pod, it goes through 2 paths, one is kubelet creates the pod and crio/cni grabs the IP from ovn and the second is getting the IP into the ovn SDB and available to crio/cni. If crio asks before its available we get CrashLoopBackoff. The trace lets us track when the key events happen and how long things take to get done.

collection-scripts/gather_network_trace
sferich888: Should we do this in other places too? Should we generalize this?

collection-scripts/gather_network_trace
sferich888: I don't think Bim questioning the value, what I am asking is if this can be condensed into the other scripts were already supporting.

collection-scripts/gather_network_trace
pecameron: @sferich888 I think a common timestamp and prefix in all of our logging would be helpful. With that we can generate a variety of traces picking up interesting logs. Also, scripts can be written to analyze the traces for expected and deviant behavior. I was interested in finding out how ovn networking sets up a pod's network. The trace is very enlightening, it shows how processing goes from daemon to daemon. Pick a pod and grep for it in the trace.

The downside is log lines get longer and files get bigger.

SOAPBOX
In a cluster there are numerous possible race conditions. For OCP to make it as a world class product we need to find and fix the races. Its a big part of making things "just work". In my years working on kernels, traces became the great salvation. Having a trace, like this, is the first small step. The heavy lifting comes in generating lines at important times and places in a fashion that is easy to analyze. That is the next step.
SOAPBOX OFF

This PR specifically generates a trace for ovn networking. I didn't want to expand and work with other teams until I got comfortable with doing this and analyzing results. I want to have something to show the other teams. My driving motivation is finding a way that anyone can look at a misbehaving cluster and quickly determine whether or not it is likely a network problem.

collection-scripts/gather_network_trace
pecameron: @sferich888 Do you mind changing each log line with the prefix and common time stamp? Part of this is relating the container to the node, which you usually don't do. Also, when we modify existing files there is more testing to do. This can be done at this stage if you think it is important. Which scripts are you thinking about? I will take a look.

sferich888:
I'm not sure? @stevekuznetsov might know.

pecameron:
@sferich888 Is there a way for me to test this script in must-gather CI test? I have a another card SDN-935 to make a ci step that calls this for use in CNO and ovnkube repos. I think I need this merged to use it in CI. Do you know a another way? I test run the script locally on an aws cluster and it has been working for me. I have not tried it from a container.

pecameron:
@stevekuznetsov PTAL

pecameron:
@dcbw PTAL

pecameron:
@sferich888 @RickJWagner PTAL
This is a new and independent script. Could we get this merged so that I can make progress on some related changes?

sferich888:
Why does it need to be an independent script?
From a support perspective when will we know to run this vs the other networking script [s] we have? 

Can this be merged or enable with a flag in the existing networking script?

pecameron:
@sferich888 The gather_network_logs script, gather_network_logs, gathers ovs flows and ovn database information as well as iptables rules. This is much different than generating a trace of execution in ovn. The extras script gathers the log files but it does not add the prefix, including node name, that identifies the source of each line. Using extras and subsequently post processing here places a dependency on extras. Also, extras will need to be refactored to capture the node name of the log files. All of this is only needed to process log files in the openshift-ovn-kubernetes namespace. Trace solves a different problem than just capturing log files. This structure of the script and what we are capturing may change over time and I prefer to not have to worry about breaking a common script that others are depending on.

This trace is the foundation for a lot of work in tracing execution. It gathers the parts and generates the trace report. We will be adding trace lines (see cno 569) to help track the sequence of events. Since this is the foundation of a lot of work that we are doing, I am very interested in getting it merged so we can start using it. I expect we will want changes in this script over time as we better understand how to use the tool and make it more effective.

Is there a slack channel we can discuss this on? I'm pcameron on forum-sdn. I am part of the ci-team (@dave-tucker leader). We are trying to understand the ci failures and triage network related failures. If you are interested you (and anyone else working on ci triage) are welcome to join.

collection-scripts/gather_network_trace
sferich888: I think I'm game. But I am concerned about the implications and performance impacts of such processing.

As you say, this will require additional testing. 
Let's explore this after we prove out the impacts with OVN. 

Can we rename this file to be explicitly for ovn_tracing?

pecameron: OK Changed the name to gather_network_ovn_trace

pecameron:
@sferich888 The new script is not used in any tests, It is just built into the image.
So test failures don't relate to this PR

sferich888:
/lgtm

@pecameron  I still think the tests need to pass - for this to merge, @deads2k and @sanchezl  to confirm

pecameron:
@sferich888 @dcbw What do I need to change to get this in? It is blocking on-going work.
Please help.

pecameron:
@dcbw @sferich888 this is blocking https://github.com/openshift/release/pull/8654
and it soon will be blocking SDN-1028.

PR #165, opened 2020-07-30
Separate Linux and Windows node log collection

sebsoto:
The '-u' flag for 'oc adm node-logs' cannot be used against Windows
nodes. This commit ensures that the '-u' method of collection is only
used for services running on Linux nodes, and the '--path' option is
used for collecting Windows node log files.

### DIFF
--- a/collection-scripts/gather_network_ovn_trace
+++ b/collection-scripts/gather_network_ovn_trace
@@ -108,7 +108,7 @@ function gather_service_data {
     > ${NETWORK_TRACE_PATH}/master_crio.log &
   PIDS+=($!)
 
-  oc adm node-logs --role="worker" -u crio | \
+  oc adm node-logs --role="worker" -l kubernetes.io/os=linux -u crio | \
     convert_journal_timestamp \
     > ${NETWORK_TRACE_PATH}/worker_crio.log &
   PIDS+=($!)
@@ -119,7 +119,7 @@ function gather_service_data {
     > ${NETWORK_TRACE_PATH}/master_kubelet.log &
   PIDS+=($!)
 
-  oc adm node-logs --role="worker" -u kubelet | \
+  oc adm node-logs --role="worker" -l kubernetes.io/os=linux -u kubelet | \
     grep -e "SyncLoop" -e "\]: E" | \
     convert_journal_timestamp \
     > ${NETWORK_TRACE_PATH}/worker_kubelet.log &
--- a/collection-scripts/gather_service_logs
+++ b/collection-scripts/gather_service_logs
@@ -15,11 +15,11 @@ SERVICE_LOG_PATH="${BASE_COLLECTION_PATH}/host_service_logs/"
 function collect_serivce_logs {  ## Takes a node role input (master or worker)
     PIDS=()
     DIR_PATH=${SERVICE_LOG_PATH}/${1}s
-    echo "WARNING: Collecting one or more service logs on ALL $1 nodes in your cluster. This could take a large amount of time." >&2
+    echo "WARNING: Collecting one or more service logs on ALL linux $1 nodes in your cluster. This could take a large amount of time." >&2
     mkdir -p ${DIR_PATH}
     for service in ${NODE_SERVICES[@]}; do
         echo "INFO: Collecting host service logs for $service"
-        /usr/bin/oc adm node-logs --role=$1 -u $service > ${DIR_PATH}/${service}_service.log &
+        /usr/bin/oc adm node-logs --role=$1 -l kubernetes.io/os=linux -u $service > ${DIR_PATH}/${service}_service.log &
 	PIDS+=($!)
     done
     echo "INFO: Waiting for worker host service log collection to complete ..."
--- a/collection-scripts/gather_windows_node_logs
+++ b/collection-scripts/gather_windows_node_logs
@@ -1,0 +1,22 @@
+#!/bin/bash
+BASE_COLLECTION_PATH="/must-gather"
+WINDOWS_NODE_LOGS=$BASE_COLLECTION_PATH/host_service_logs/windows
+
+# Logfile list
+LOGS=(kube-proxy/kube-proxy.exe.INFO kube-proxy/kube-proxy.exe.ERROR kube-proxy/kube-proxy.exe.WARNING)
+LOGS+=(hybrid-overlay/hybrid-overlay.log kubelet/kubelet.log)
+
+# if the cluster has no Windows nodes skip this script
+WIN_NODES=$(/usr/bin/oc get no -l kubernetes.io/os=windows)
+if [ -z "$WIN_NODES" ]; then
+    exit 0
+fi
+
+echo INFO: Collecting logs for all Windows nodes
+for log in ${LOGS[@]}; do
+    LOG_DIR=${WINDOWS_NODE_LOGS}/$(dirname $log)
+    mkdir -p ${LOG_DIR}
+    /usr/bin/oc adm node-logs -l kubernetes.io/os=windows --path=$log > ${LOG_DIR}/$(basename $log) &
+    PIDS+=($!)
+done
+wait ${PIDS[@]}
--- a/collection-scripts/gather
+++ b/collection-scripts/gather
@@ -46,3 +46,6 @@ done
 
 # Gather Service Logs (using a suplamental Script); Scoped to Masters.
 /usr/bin/gather_service_logs master
+
+# Gather Windows Kubernetes component logs
+/usr/bin/gather_windows_node_logs


collection-scripts/gather_service_logs
```diff
@@ -15,11 +15,11 @@ SERVICE_LOG_PATH="${BASE_COLLECTION_PATH}/host_service_logs/"
 function collect_serivce_logs {  ## Takes a node role input (master or worker)
     PIDS=()
     DIR_PATH=${SERVICE_LOG_PATH}/${1}s
-    echo "WARNING: Collecting one or more service logs on ALL $1 nodes in your cluster. This could take a large amount of time." >&2
+    echo "WARNING: Collecting one or more service logs on ALL linux $1 nodes in your cluster. This could take a large amount of time." >&2
     mkdir -p ${DIR_PATH}
     for service in ${NODE_SERVICES[@]}; do
         echo "INFO: Collecting host service logs for $service"
-        /usr/bin/oc adm node-logs --role=$1 -u $service > ${DIR_PATH}/${service}_service.log &
+        /usr/bin/oc adm node-logs --role=$1 -l kubernetes.io/os=linux -u $service > ${DIR_PATH}/${service}_service.log &
```
sferich888, line 22: Can we use one or more labels here?

collection-scripts/gather_windows_node_logs
```diff
@@ -0,0 +1,22 @@
+#!/bin/bash
```
sferich888, line 1: Instead of this; would it be better to just merge this into one bash script and select the things we want using labels?

collection-scripts/gather_service_logs
sebsoto: Yes a comma separated list can be used

collection-scripts/gather_network_ovn_trace
```diff
@@ -108,7 +108,7 @@ function gather_service_data {
     > ${NETWORK_TRACE_PATH}/master_crio.log &
   PIDS+=($!)
 
-  oc adm node-logs --role="worker" -u crio | \
+  oc adm node-logs --role="worker" -l kubernetes.io/os=linux -u crio | \
```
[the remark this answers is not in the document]
sebsoto: For OVN onn Windows, there is a Windows service named hybrid-overlay.exe instead. the -u option does not work on windows as well, as that is specific to journald, which windows does not have

collection-scripts/gather_windows_node_logs
sebsoto: Most of the Linux log collection relies on the fact that the service logs being collected are from journald. This is not the case for Windows, and why I felt it made sense to seperate them

collection-scripts/gather_network_ovn_trace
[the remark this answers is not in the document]
sferich888: Shouldn't `oc adm node-logs` account for this? Is this a bug / issue we should address in that command?

collection-scripts/gather_windows_node_logs
sferich888: We are missing container runtime, logs (docker), as well as kublet how do we get those added?

collection-scripts/gather_network_ovn_trace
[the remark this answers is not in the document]
sebsoto: I would not consider it a bug, as there really isn't a way to get it to work on Windows. The closest parallel to journald would be the Windows Event Viewer, but we none of the Windows Kubernetes components are logging to that.

collection-scripts/gather_windows_node_logs
sferich888: We want to collect everything required to debug what is happening on the node (service logs wise); as it pertains to OpenShift. So I think we want to collect at the very least the container runtime and the kublet process logs (thinks that are installed outside of OpenShift that are pre-requisites for OpenShift to run or be connected to the cluster). .

aravindhp: @sferich888 collecting kubelet logs is doable. But as @sebsoto mentioned, collecting the Docker logs will be difficult. Please open an issue on our board so that we can track this.

sebsoto: I've added the kubelet log.

The issue with the Docker logs is that the logging location for that is not under our control. Our product asks that our users bring their own Windows VM image, with docker pre-installed. We'd either have to increase the customer requirements, or investigate reconfiguring Docker.

sdodson:
/lgtm
This looks like a step in the right direction and given the constraints we're working under with respect to Docker installation and configuration this may be the best we can do right now.

@sferich888 @davemulford I'd appreciate if you approved this.

collection-scripts/gather_windows_node_logs
```diff
@@ -0,0 +1,22 @@
...
+BASE_COLLECTION_PATH="/must-gather"
+WINDOWS_NODE_LOGS=$BASE_COLLECTION_PATH/windows
```
sferich888, line 3: Do we want to update https://github.com/openshift/must-gather/blob/master/collection-scripts/gather_service_logs#L13 to say `${BASE_COLLECTION_PATH}/host_service_logs/linux`, and then the referenced line can read `${BASE_COLLECTION_PATH}/host_service_logs/windows` Thus helping with the find-ability and placement consistency of the logs.

davemulford: Yes, having uniformly named directories would benefit both customers and support. We should definitely have this change.

sferich888:
> /lgtm
> This looks like a step in the right direction and given the constraints we're working under with respect to Docker installation and configuration this may be the best we can do right now.
> 
> @sferich888 @davemulford I'd appreciate if you approved this.

So after looking into this more, I don't think this PR will work? But I could be wrong. 

The way we get logs from nodes is done with a
[carry patch on the kubelet](https://github.com/openshift/kubernetes/blob/master/pkg/kubelet/kubelet_server_journal.go). 

As I understand it, the window nodes will use an upstream kubelet, and as such the method or approach to getting node logs will not be available on windows. 

Does anyone know, how / what API `oc adm node-logs` is connecting to on a window node?

sebsoto:
@sferich888 I have confirmed that it does work, to answer your question for why it works I am looking into that

sebsoto:
@sferich888 That carry patch seems to be specific to the -u `node-logs` option, we are using --path.

aravindhp:
@sferich888 the carry patch is for `Provide an administrator a streaming view of journal logs without them having to implement a client side reader. Only available to cluster admins.`. There is no concept of a journal log in Windows. We are using `oc adm node-logs --path` to work around it.

sferich888:
@aravindhp yes, but I think[ `oc adm node-logs --path`](https://github.com/openshift/oc/blob/master/pkg/cli/admin/node/logs.go#L249-L251) still uses the shim (carry patch) to get at the hosts logs (correct)? @smarterclayton

aravindhp:
@sferich888 `oc adm node-logs --path` is something we can confirm is working with an upstream kubelet running on a Windows node. And for the product due to another carry patch https://github.com/openshift/kubernetes/pull/301 and release requirements we will be switching to a downstream kubelet for Windows also.

collection-scripts/gather_windows_node_logs
sebsoto: I have made this change

davemulford:
Thanks for the quick changes @sebsoto. 

This looks fine to me, @sferich888 what do you think?

/LGTM

sferich888:
/lgtm

I am reluctantly approving this (only because I feel its not enough), outside of that the PR looks fine provided the shims work on windows. 
To make sure we don't loose sight of the minimum requirements for a positive customer experience we need to tread  https://bugzilla.redhat.com/show_bug.cgi?id=1862643 and a release/feature blocker bug, and look at adding the missing log collection quickly.

sdodson:
@sebsoto Test failure looks legitimate so we're losing crio logs for some reason.

```
STEP: Destroying namespace "e2e-test-oc-adm-must-gather-8jkq2" for this suite.
Aug  3 11:14:19.516: INFO: Running AfterSuite actions on all nodes
Aug  3 11:14:19.516: INFO: Running AfterSuite actions on node 1
fail [github.com/openshift/origin@/test/extended/cli/mustgather.go:95]: Expected
    <string>: /tmp/test.oc-adm-must-gather.195409629/registry-svc-ci-openshift-org-ci-op-myszpvy0-stable-sha256-b2b88645ed01f2028b807015ad7eefc06b48e3a9f58cb884e01ee19c8ead9f27/host_service_logs/masters/crio_service.log
to exist
```

sebsoto:
@sdodson There seems to be a _repo race condition_ here. I was asked to change the directory structure, which moved the linux logs into an unexpected place for this test: https://github.com/openshift/origin/blob/master/test/extended/cli/mustgather.go#L75.

I cant change this repo unless that repo's test is changed, and I cant change that repo's test until this repo is changed.

sebsoto:
The test expects log files in: `{pluginOutputDir, "host_service_logs", "masters", "crio_service.log"},` my changes moved them to `	{pluginOutputDir, "host_service_logs", "linux", "masters", "crio_service.log"},

collection-scripts/gather_windows_node_logs
sdodson: I think given we have both tests and users expecting to find linux logs in one location today we should leave that unchanged but write out windows logs, which are net new, to a new location. So lets revert the relocation of linux logs.

PR #166, opened 2020-08-04
Update gather_service_logs

sferich888:
To match with https://github.com/openshift/must-gather/blob/master/collection-scripts/gather_windows_node_logs#L3 we need to move this under a `linux` directory.

### DIFF
--- a/collection-scripts/gather_service_logs
+++ b/collection-scripts/gather_service_logs
@@ -10,7 +10,7 @@ NODE_SERVICES+=${2:-${GENERAL_SERVICES[@]}}
 NODE_SERVICES+=()          ### Placeholder to extend Node only services
 
 # Collect System Service Logs
-SERVICE_LOG_PATH="${BASE_COLLECTION_PATH}/host_service_logs/"
+SERVICE_LOG_PATH="${BASE_COLLECTION_PATH}/host_service_logs/linux"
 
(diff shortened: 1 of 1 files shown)


sdodson:
I disagree with this change. Test cases, humans, and potentially any number of tools expect linux logs to be located at their current location and by moving them we're breaking all of those consumers.

sferich888:
We need to keep consistency with how we arrange files in the archive. 
If we are not ok changing this (and the associated tests) we should revert https://github.com/openshift/must-gather/pull/165

sferich888:
https://github.com/openshift/origin/pull/25371 can hopefully fix the test issues connected to this.

dougsland:
> I disagree with this change. Test cases, humans, and potentially any number of tools expect linux logs to be located at their
> current location and by moving them we're breaking all of those consumers.

I agree with Scott. Can we have a symbol link to avoid tools breaking mentioned by @sdodson ?

Example:
$ ls -la .
     myfile  -> linux/myfile

Another question: How about documentation to users? Do we have any documentation showing the new location and why we changed?

sferich888:
@deads2k / @mfojtik what direction would the CLI team like to take with this?

sferich888:
@deads2k / @mfojtik bump (what direction should we take with this)?

cc: @davemulford

PR #169, opened 2020-08-24
Bug 1868716: Implements more log collection for Openshift networking

astoycos:
This PR implements more log collection for Openshift Networking,

with OVN-Kubernetes

-It fixes https://bugzilla.redhat.com/show_bug.cgi?id=1868716

-Specifically it adds the collection script `gather_core_dumps`
to collect core dumps on hosts

-Zips the ovn-nb/sb.db files and adds flow/port
dumping for the br-ex interface in `gather_network_logs`

-Ensures for both `gather_network_logs` and `gather_core_dumps`
that we only query linux based nodes

-Leaves the networking logging off by default in `gather`

-to run the networking collection scripts use the following
command:
`oc adm must-gather -- gather && gather_core_dumps &&
gather_network_logs && gather_network_ovn_trace`

### DIFF
--- a/collection-scripts/gather_core_dumps
+++ b/collection-scripts/gather_core_dumps
@@ -1,0 +1,24 @@
+#!/bin/bash
+BASE_COLLECTION_PATH="must-gather"
+CORE_DUMP_PATH=${OUT:-"${BASE_COLLECTION_PATH}/node_core_dumps"}
+
+mkdir -p ${CORE_DUMP_PATH}/
+
+function gather_core_dump_data {
+  for NODE in ${NODES}; do
+    oc debug node/${NODE} -- chroot /host /bin/bash -c "coredumpctl info" > ${CORE_DUMP_PATH}/${NODE}_core_dump & PIDS+=($!)
+  done
+}
+
+if [ $# -eq 0 ]; then
+    echo "WARNING: Collecting network logs on ALL linux nodes in your cluster. This could take a long time." >&2
+fi
+
+PIDS=()
+NODES="${@:-$(oc get nodes -o jsonpath='{range .items[*]}{@.metadata.name} {.status.nodeInfo.operatingSystem==linux}')}"
+
+gather_core_dump_data
+
+echo "INFO: Waiting for node core dump collection to complete ..."
+wait ${PIDS[@]}
+echo "INFO: Node core dump collection to complete."
--- a/collection-scripts/gather_network_logs
+++ b/collection-scripts/gather_network_logs
@@ -56,6 +56,13 @@ function gather_ovn_kubernetes_nodes_data {
       oc -n openshift-ovn-kubernetes exec ${OVS_NODE_POD} -- bash -c \
       "ovs-ofctl dump-flows br-local" > ${NETWORK_LOG_PATH}/${NODE}_${OVS_NODE_POD}_ovs_ofctl_dump_flows_br_local &
       PIDS+=($!)
+       oc -n openshift-ovn-kubernetes exec ${OVS_NODE_POD} -- bash -c \
+      "ovs-ofctl dump-ports-desc br-ex" > ${NETWORK_LOG_PATH}/${NODE}_${OVS_NODE_POD}_ovs_ofctl_dump_ports_br_ex &
+      PIDS+=($!)
+      oc -n openshift-ovn-kubernetes exec ${OVS_NODE_POD} -- bash -c \
+      "ovs-ofctl dump-flows br-ex" > ${NETWORK_LOG_PATH}/${NODE}_${OVS_NODE_POD}_ovs_ofctl_dump_flows_br_ex &
+      PIDS+=($!)
+
       oc -n openshift-ovn-kubernetes exec ${OVS_NODE_POD} -- bash -c \
       "ovs-vsctl show" > ${NETWORK_LOG_PATH}/${NODE}_${OVS_NODE_POD}_ovs_dump &
       PIDS+=($!)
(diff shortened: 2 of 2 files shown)


collection-scripts/gather
```diff
@@ -49,3 +49,9 @@ done
 
 # Gather Windows Kubernetes component logs
 /usr/bin/gather_windows_node_logs
+
```
trozet, line 52: can we also gather core dumps for ovn-controller, northd?

trozet: also can we get RPM versions of OVN

astoycos: Yep should be able to handle that! Just need to figure out the necessary core dump procedure + commands

dcbw:
@trozet @astoycos so we have not traditionally done this becuase it could be a huge amount of data, for *all* PRs, not just networking ones.

We actually have a different approach, which is done in the openshift/release repo's "step registry" for specific jobs like ovnkube, sdn, and CNO. See https://github.com/openshift/release/pull/8654 which adds a CI "step" for gathering OVN trace and enables that step from aws-ovn jobs. Also https://github.com/openshift/release/pull/7865 which gathers network info.

The problem with https://github.com/openshift/release/pull/7865 is that it doesn't use must-gather, and that's what we really want. But to do that we need to get the must-gather changes into an image that CI actually uses, and unfortunately CI uses an older must-gather or something like that...

astoycos:
Sounds good, It seems that [openshift/release#8654](https://github.com/openshift/release/pull/8654) collects almost everything we want to log, this PR was just activating the must gather scripts that are used by the step registry (Assuming that the ` /usr/bin/gather network logs` script referenced [here](https://github.com/openshift/release/blob/master/ci-operator/step-registry/gather/network/gather-network-commands.sh#L14) is the same as the one [defined within `must gather` ](https://github.com/openshift/must-gather/blob/master/collection-scripts/gather_network_logs)).  

1. @trozet I should just be able to update this PR to only edit the `collection-scripts/gather_network_logs` file in order to additionally log the `br-ex` interface port/flow dumps, core dumps, and OVN RPM version. 

 2. @dcbw  I am a bit confused here, how can we use `must gather` without activating the network logging scripts as done in this PR? And how would a general user be able to run this and collect data for debugging purposes? They would have to run something like [this](https://github.com/openshift/release/blob/master/ci-operator/step-registry/gather/network/gather-network-commands.sh#L14) right? Why wouldn't we just have a subset of commands within `must-gather`? I.E `must-gather <sub-component> `  .  

3. I can help out with this after getting everything Ironed out.

trozet:
@dcbw we want must-gather to include the things we need to be able to debug networking correctly in CI as well as in the field. I don't think there is any way around this. I'm not sure where the huge amounts of data come from...besides dbs which are raw json that we can compress when we gather.

collection-scripts/gather
```diff
@@ -49,3 +49,12 @@ done
...
+# Gather OVN traces 
+/usr/bin/gather_network_ovn_trace
```
trozet, line 54: just to confirm @astoycos this wont cause any failures if it is run on openshift-sdn cluster right? It just wont collect anything?

astoycos: Right this (https://github.com/openshift/must-gather/blob/master/collection-scripts/gather_network_logs#L131) Checks what type of networking is being used before gathering

trozet:
We have 2 DBs in OVN, NB and SB. Each of these DBs runs on the kubernetes master node (so 3 nodes). From looking at some bzs and current clusters it looks like the biggest size I have seen thus far is at scaling to 179 nodes, without a lot of pods:

NB- 3MB 
SB - 31MB

After gzip compression:
NB- 458KB
SB- 6MB

So that means we can guesstimate around 6.5MB *3 nodes so roughly around 20mb of data per gather for the dbs.

collection-scripts/gather_core_dumps
```diff
@@ -0,0 +1,20 @@
...
+mkdir -p ${CORE_DUMP_PATH}/
+
+function gather_core_dump_data {
+  for NODE in ${NODES}; do
+    oc debug node/${NODE} -- chroot /host /bin/bash -c "coredumpctl info" > ${CORE_DUMP_PATH}/${NODE}_core_dump & PIDS+=($!)
+  done
+}
+
+PIDS=()
+NODES="${@:-$(oc get nodes --no-headers -o custom-columns=':metadata.name')}"
```
sferich888, line 14: We need to scope this to linux only hosts.

collection-scripts/gather
```diff
@@ -49,3 +49,12 @@ done
...
+
+#Gather network logs 
+/usr/bin/gather_network_logs
+
+#Gather node core dumps 
+/usr/bin/gather_core_dumps
```
sferich888, line 60: Reading over this script (as part of the PR). This would run on all nodes, so if you had 200 nodes you would have to wait for all 200 nodes to collect this data. 

As a result I don't think we want this to run by default.

collection-scripts/gather
```diff
@@ -49,3 +49,12 @@ done
 
 # Gather Windows Kubernetes component logs
 /usr/bin/gather_windows_node_logs
+
+# Gather OVN traces 
+/usr/bin/gather_network_ovn_trace
+
+#Gather network logs 
+/usr/bin/gather_network_logs
```
sferich888, line 57: This would run on all nodes, so if you had 200 nodes you would have to wait for all 200 nodes to collect this data.

As a result I don't think we want this to run by default.

collection-scripts/gather_core_dumps
astoycos: Agreed, updated accordingly

collection-scripts/gather
astoycos: +1 I deactivated by default

astoycos:
@sferich888 I addressed your comments and ensured that the networking log collection scripts were not enabled by default

collection-scripts/gather_network_logs
```diff
@@ -65,10 +72,15 @@ function gather_ovn_kubernetes_nodes_data {
...
       if [[ ${OVNKUBE_MASTER} != "" ]] ; then
         oc cp openshift-ovn-kubernetes/${OVNKUBE_MASTER}:/etc/openvswitch/ovnsb_db.db ${NETWORK_LOG_PATH}/${NODE}_ovnsb_db.db > ${NETWORK_LOG_PATH}/${NODE}_sbdb &
+        gzip ${NETWORK_LOG_PATH}/${NODE}_sbdb &        
         PIDS+=($!)
+        
         oc cp openshift-ovn-kubernetes/${OVNKUBE_MASTER}:/etc/openvswitch/ovnnb_db.db ${NETWORK_LOG_PATH}/${NODE}_ovnnb_db.db > ${NETWORK_LOG_PATH}/${NODE}_nbdb &
+        gzip ${NETWORK_LOG_PATH}/${NODE}_nbdb &
         PIDS+=($!)
       fi
+
```
trozet, line 82: extra newlines here

davemulford:
All of the changes asked for here and on slack have been made.

/LGTM

PR #171, opened 2020-08-25
[wip] collection-scripts: add sos networking

hexfusion:
This is a first attempt at bringing back some of the sos functionality into must gather. I took the approach of making the ip calls inside of a privileged pod using HostNetworking (quorum-guard) to see if we could avoid using oc debug directly.

The list of commands was just a mapping of what existed in sos and available in the cli base image.

The output of these commands needs to be vetted as some show details on container-specific networking vs host.

### DIFF
--- a/collection-scripts/gather_sos_networking
+++ b/collection-scripts/gather_sos_networking
@@ -1,0 +1,58 @@
+#!/bin/bash
+
+set -o errexit
+set -o pipefail
+
+BASE_COLLECTION_PATH="/must-gather"
+LOG_PATH="${BASE_COLLECTION_PATH}/sos_info/networking"
+
+# use quorum-guard becuase it is privileged and uses hostNetwork
+QUORUM_GUARD_DATA=($(oc get pods -n openshift-etcd -l name=etcd-quorum-guard -o jsonpath='{range .items[*]}{.metadata.name}{","}{.status.hostIP}{" "}{end}')) || {
+  echo "ERROR: Skipping network data collection"
+  exit 0
+}
+
+if [[ -z "$QUORUM_GUARD_DATA" ]];then
+  echo "ERROR: Skipping network data collection"
+  exit 0
+fi
+
+set -o nounset
+
+COMMANDS=(
+  "ip -4 rule"
+  "ip -6 rule"
+  "ip maddr show"
+  "ip neigh show nud noarp"
+  "ip -o addr"
+  "ip -s -d link"
+  "ip -6 route show table all"
+  "ip -d address"
+  "ip mroute show"
+  "ip netns"
+  "ip route show table all"
+  "ip -s -s neigh show"
+)
+
+run_command(){
+  local pod="$1"
+  local node="$2"
+  local command="$3"
+  oc rsh -n openshift-etcd $pod $command 1> "$LOG_PATH/${node}-${command//_/ }.log" || true
+}
+
+mkdir -p ${LOG_PATH}
+echo "INFO: Start getting networking information"
+
+for pod in "${QUORUM_GUARD_DATA[@]}"; do
+  # create array element [0] is pod name [1] is node ip
+  IFS=',' read -r -a data <<< "$pod"
+  echo "Getting networking data from hostNetwork on \"${data[1]}\" via \"${data[0]}\""
+  for command in "${COMMANDS[@]}"; do
+    run_command "${data[0]}" "${data[1]}" "$command" &
+  done
+done
+
+echo "INFO: Waiting for networking info collection to complete ..."
+wait
+echo "INFO: Done collecting networking information"


collection-scripts/gather_sos_networking
```diff
@@ -0,0 +1,59 @@
...
+set -o errexit
+set -o pipefail
+
+BASE_COLLECTION_PATH="/must-gather"
+LOG_PATH="${BASE_COLLECTION_PATH}/sos_info/networking"
+
+# use quorum-guard becuase it is privileged and uses hostNetwork
+QUORUM_GUARD_DATA=($(oc get pods -n openshift-etcd -l name=etcd-quorum-guard -o jsonpath='{range .items[*]}{.metadata.name}{","}{.status.hostIP}{" "}{end}')) || {
+  echo "ERROR: No networking data was collected"
+  exit 1
```
deads2k, line 12: will this affect the return code of the entire command?  I don't think we want to fail like this from a gathering tool

hexfusion: yeah if oc has error we exit 1 I can tolerate oc failure.

collection-scripts/gather_sos_networking
```diff
@@ -0,0 +1,59 @@
...
+
+for pod in "${QUORUM_GUARD_DATA[@]}"; do
+  # create array element [0] is pod name [1] is node ip
+  IFS=',' read -a data <<< "$pod"
+  echo "Getting networking data from hostNetwork on \"${data[1]}\" via \"${data[0]}\""
+  for command in "${COMMANDS[@]}"; do
+    run_command ${data[0]} ${data[1]} $command &
+  done
+done
+
```
hexfusion, line 55: extra space

collection-scripts/gather_sos_networking
```diff
@@ -0,0 +1,59 @@
...
+  ip_netns
+  ip_route_show_table_all
+  ip_-s_-s_neigh_show
+)
+
+run_command(){
+  local pod="$1"
+  local node="$2"
+  local command="$3"
+  eval oc exec  -n openshift-etcd $pod -- /bin/sh -c "'"${command//_/ }"'" 1> ${node}-${command}.log || true
```
hexfusion, line 41: all log path

collection-scripts/gather_sos_networking
hexfusion: now exits 0 I need to test this within context of the tool to understand failure better.

collection-scripts/gather_sos_networking
```diff
@@ -0,0 +1,58 @@
...
+LOG_PATH="${BASE_COLLECTION_PATH}/sos_info/networking"
+
+# use quorum-guard becuase it is privileged and uses hostNetwork
+QUORUM_GUARD_DATA=($(oc get pods -n openshift-etcd -l name=etcd-quorum-guard -o jsonpath='{range .items[*]}{.metadata.name}{","}{.status.hostIP}{" "}{end}')) || {
+  echo "ERROR: No networking data was collected"
+  exit 0
+}
+
+if [[ -z $QUORUM_GUARD_DATA ]];then
+  echo "ERROR: No data was returned from oc request"
```
hexfusion, line 16: message could use improving

collection-scripts/gather_sos_networking
```diff
@@ -0,0 +1,58 @@
...
+  ip_-o_addr
+  ip_-s_-d_link
+  ip_-6_route_show_table_all
+  ip_-d_address
+  ip_mroute_show
+  ip_netns
+  ip_route_show_table_all
+  ip_-s_-s_neigh_show
+)
+
```
retroflexer, line 36: The script can be more readable if the commands are with spaces and the output file is replaced with underscores. The following works for me:
```
COMMANDS=(
  "ip -4 rule"
  "ip -6 rule"
  "ip maddr show"
  "ip neigh show nud noarp"
  "ip -o addr"
  "ip -s -d link"
  "ip -6 route show table all"
  "ip -d address"
  "ip mroute show"
  "ip netns"
  "ip route show table all"
  "ip -s -s neigh show"
)
```
and 
```
oc rsh  -n openshift-etcd $pod $command > ${node}-${command// /_}.log || true
```

collection-scripts/gather_sos_networking
```diff
@@ -0,0 +1,58 @@
...
+run_command(){
+  local pod="$1"
+  local node="$2"
+  local command="$3"
+  eval oc exec  -n openshift-etcd $pod -- /bin/sh -c '"${command//_/ }"' 1> "$LOG_PATH/${node}-${command}.log" || true
```
retroflexer, line 41: This can be simplified by using `oc rsh`. No eval and complicated quotes would be needed. 
```
oc rsh  -n openshift-etcd $pod $command > ${node}-${command// /_}.log || true
```
See the above comment:

https://github.com/openshift/must-gather/pull/171/files#r476628760

collection-scripts/gather_sos_networking
```diff
@@ -0,0 +1,58 @@
...
+
+mkdir -p ${LOG_PATH}
+echo "INFO: Starting getting networking information"
+
+for pod in "${QUORUM_GUARD_DATA[@]}"; do
+  # create array element [0] is pod name [1] is node ip
+  IFS=',' read -a data <<< "$pod"
+  echo "Getting networking data from hostNetwork on \"${data[1]}\" via \"${data[0]}\""
+  for command in "${COMMANDS[@]}"; do
+    run_command ${data[0]} ${data[1]} $command &
```
retroflexer, line 52: If the commands are defined with spaces and double-quotes, this needs to be quoted:
```
run_command ${data[0]} ${data[1]} "$command"
```

collection-scripts/gather_sos_networking
hexfusion: makes sense I just created array with filenames but yours is easier to read

hexfusion:
This is really not the direction we should go I will add a PR adding some NW data that I personally find interesting in another PR.

hexfusion:
https://github.com/openshift/must-gather/pull/172

PR #172, opened 2020-08-25
Bug 1874301: collection-scripts/gather_network_logs: add detailed status for interfaces

hexfusion:
It is very helpful to understand the details of host network interfaces to understand issues such as dropped packets. This PR adds `ip -s -d link` tot he current list of ip command outputs collected. This command was previously part of `sosreport`.


hexfusion:
infra flakes
/retest

dcbw:
@hexfusion needs an OpenShift bugzilla bug filed though, with Target Release: 4.6, and then you add "Bug <xxx>: " to the start of the PR title.

hexfusion:
cc @sferich888

sferich888:
By default this will not be run, your ok with that?

PR #173, opened 2020-08-28
Bug 1877428: Update gather_core_dumps to get raw coredump files 

astoycos:
The previous iteration of `gather_core_dumps` used
`coredumpsctl info` to get information

It was shown that we in fact need the raw coredump
files for debugging

This PR extracts corefiles from the nodes by
1. Forcing a Debug pod to stay open on node
2. Waiting for the Debug pod to run
3. Using `oc cp` to extract the necessary files

TODO: find a way to do this without forcing the
debug pod to stay running

### DIFF
--- a/collection-scripts/gather_core_dumps
+++ b/collection-scripts/gather_core_dumps
@@ -4,10 +4,35 @@ CORE_DUMP_PATH=${OUT:-"${BASE_COLLECTION_PATH}/node_core_dumps"}
 
 mkdir -p ${CORE_DUMP_PATH}/
 
+function get_dump_off_node { 
+    #Start Debug pod force it to stay up for 60 seconds, supress Stdout 
+    oc debug node/$1 -- /bin/bash -c 'sleep 60' > /dev/null 2>&1 &
+    
+    #Clear local variables before assignment
+    local n=0
+    local debug=""
+
+    # Make sure pod is avaliable exit after 5 seconds if not 
+    while [[ -z "$debug"  && $n -le 10 ]]
+    do    
+        #Get name of Debug Pod for relevant node 
+        debug=$(oc get pods | grep $1-debug | awk -F' ' '{print $2}')
+        (( n++))
+        sleep 1
+    done 
+
+    # Make sure debug pod is ready
+    oc wait --for=condition=Ready pod/$debug --timeout=10s
+
+    #Copy Core Dumps out of Nodes suppress Stdout
+    oc cp $debug:/host/var/lib/systemd/coredump ${CORE_DUMP_PATH}/$1_core_dump > /dev/null 2>&1 && PIDS+=($!)
+}
+
 function gather_core_dump_data {
-  for NODE in ${NODES}; do
-    oc debug node/${NODE} -- chroot /host /bin/bash -c "coredumpctl info" > ${CORE_DUMP_PATH}/${NODE}_core_dump & PIDS+=($!)
-  done
+  #Run coredump pull function on all nodes in parallel
+  for NODE in ${NODES}; do 
+    get_dump_off_node ${NODE} & 
+  done 
 }
 
 if [ $# -eq 0 ]; then


astoycos:
@trozet

collection-scripts/gather_core_dumps
```diff
@@ -4,10 +4,38 @@ CORE_DUMP_PATH=${OUT:-"${BASE_COLLECTION_PATH}/node_core_dumps"}
...
+    oc debug node/$1 -- /bin/bash -c 'sleep 60' > /dev/null 2>&1 &
+
+    #Get name of Debug Pod for relevant node 
+    export DEBUG=$(oc get pods --all-namespaces | grep $1-debug | awk -F' ' '{print $2}')
+
+    # Make sure pod is avaliable
+    while [ -z "${DEBUG}" ]
+    do  
+        sleep 1 
+        export DEBUG=$(oc get pods --all-namespaces | grep $1-debug | awk -F' ' '{print $2}')
```
trozet, line 18: this looks the same as line 12. Why not just get rid of line 12 and have this loop?

collection-scripts/gather_core_dumps
```diff
@@ -4,10 +4,38 @@ CORE_DUMP_PATH=${OUT:-"${BASE_COLLECTION_PATH}/node_core_dumps"}
...
 
+function get_dump_off_node { 
+    #Start Debug pod force it to stay up for 60 seconds, supress Stdout 
+    oc debug node/$1 -- /bin/bash -c 'sleep 60' > /dev/null 2>&1 &
+
+    #Get name of Debug Pod for relevant node 
+    export DEBUG=$(oc get pods --all-namespaces | grep $1-debug | awk -F' ' '{print $2}')
+
+    # Make sure pod is avaliable
+    while [ -z "${DEBUG}" ]
```
trozet, line 15: should we have a counter here as well to ensure this doesn't run forever?

collection-scripts/gather_core_dumps
```diff
@@ -4,10 +4,38 @@ CORE_DUMP_PATH=${OUT:-"${BASE_COLLECTION_PATH}/node_core_dumps"}
...
+    do  
+        sleep 1 
+        export DEBUG=$(oc get pods --all-namespaces | grep $1-debug | awk -F' ' '{print $2}')
+    done 
+
+    # Make sure pod is "Running"
+    until
```
trozet, line 22: ditto do you wan tot use a counter here to stop this from running forever?

collection-scripts/gather_core_dumps
astoycos: Yeah should work, there is the case the pod is already registered and we would skip this loop completely, but either should work, It just checks if "DEBUG" is empty or not

collection-scripts/gather_core_dumps
astoycos: I agree

collection-scripts/gather_core_dumps
astoycos: Yep sounds good

collection-scripts/gather_core_dumps
```diff
@@ -4,10 +4,31 @@ CORE_DUMP_PATH=${OUT:-"${BASE_COLLECTION_PATH}/node_core_dumps"}
 
 mkdir -p ${CORE_DUMP_PATH}/
 
+function get_dump_off_node { 
+    #Start Debug pod force it to stay up for 60 seconds, supress Stdout 
+    oc debug node/$1 -- /bin/bash -c 'sleep 60' > /dev/null 2>&1 &
+
+    # Make sure pod is avaliable exit after 5 seconds if not 
+    while [[ -z "${DEBUG}"  && $n -le 10 ]]
```
trozet, line 12: you probably want to set n to 0 before this, and unset debug, or you make them both local vars to this function. Otherwise calling this function multiple times may have n already incremented or DEBUG already have a value.

astoycos: Ahh you're right, nice catch, I will convert `debug` and the counter to local variables, and then since each function call is being run in the background, i.e separate processes, there shouldn't be any chance of variable overwriting

collection-scripts/gather_core_dumps
```diff
@@ -4,10 +4,35 @@ CORE_DUMP_PATH=${OUT:-"${BASE_COLLECTION_PATH}/node_core_dumps"}
...
+    
+    #Clear local variables before assignment
+    local n=0
+    local debug=""
+
+    # Make sure pod is avaliable exit after 5 seconds if not 
+    while [[ -z "$debug"  && $n -le 10 ]]
+    do    
+        #Get name of Debug Pod for relevant node 
+        debug=$(oc get pods --all-namespaces | grep $1-debug | awk -F' ' '{print $2}')
```
rbbratta, line 19: won't the debug pod be in the current namespace?   line 9 should be in the current namespace.

astoycos: Yep it should always pop up in the current namespace, I can remove the `--all-namespaces` flag

astoycos:
@rbbratta should be all good now

PR #174, opened 2020-09-11
Bug 1877428: Update gather_core_dumps collection of debug pod's name, clean up resources 

astoycos:
This Commit fixes the failed validation on 
PR https://github.com/openshift/must-gather/pull/173`

in response to the [feedback from QA](https://bugzilla.redhat.com/show_bug.cgi?id=1877428)

It updated how the name of the debug pod is
determined

Specifically it uses `sed` to extract the pod name
when the pod is created with `oc debug` rather than 
using `oc get` to try and determine the name after the fact
since there is more room for errors in string formatting 

Also ensures that the debug pod is delete the moment we 
are finished with it rather than allowing it to hang 


sferich888 (commented):
@davemulford you did the first review here, do you have issues with this?

davemulford:
No issues here, @sferich888!

PR #175, opened 2020-09-12
Bug 1868794: Gather OVS logs from host

rcarrillocruz:
Since we switched to host OVS, it is useful to gather it.

### DIFF
--- a/collection-scripts/gather_network_logs
+++ b/collection-scripts/gather_network_logs
@@ -69,6 +69,11 @@ function gather_ovn_kubernetes_nodes_data {
       "ovs-vsctl show" > ${NETWORK_LOG_PATH}/${NODE}_${OVS_NODE_POD}_ovs_dump &
       PIDS+=($!)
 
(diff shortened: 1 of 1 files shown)


dcbw:
Note that these should be available from the sdn-ovs container since we tail those logs in the container script even for the host, no?

rcarrillocruz:
Not sure I follow your suggestion. If you check the gather_network_logs script it seems we only gather ovs logs for openshift-sdn.
Are you suggesting that instead of doing oc cp from /host I do the same as done by gather_openshiftsdn_nodes_data function:

i.e.

oc -n openshift-sdn exec ${OVS_POD} -- bash -c "cat /var/log/openvswitch/ovs-vswitchd.log" > ${NETWORK_LOG_PATH}/${NODE}_${OVS_POD}_vswitchd_log &

dcbw:
@rcarrillocruz not quite. What I'm saying is that from the OVS containers themselves, we tail each of the OVS component logs:

https://github.com/openshift/cluster-network-operator/blob/master/bindata/network/ovn-kubernetes/006-ovs-node.yaml#L63
https://github.com/openshift/cluster-network-operator/blob/master/bindata/network/openshift-sdn/sdn-ovs.yaml#L51

so I'm just pointing out that we already have some of this information. That said, we'd loose historical logs if the containers are restarted, so I think this change is still OK.

/lgtm

PR #176, opened 2020-09-15
Bug 1861746: force disk flush after collecting audit logs

soltysh:
Proof in https://github.com/openshift/origin/pull/25493


deads2k:
holding because we probably need it in the main script too, right?

PR #177, opened 2020-09-16
Bug 1877428: Increase gather_core_dump robustness after seeing some CI issues

astoycos:
There were still some edge case bugs here
involving the script's integration with
must-gather

This commit fixes it by getting the debug pod's
name and and now namespace in a much more 
robust manner

specifically it uses `oc debug -o jsonpath` to
determine the name of the debug pod prior to
actual pod creation rather than bash string 
processing

It also extends the time a debug pod may persist
to 5 minutes in the case of large coredump files
that need to be extracted from the nodes.

lastly it cleans up some bash script formatting 
issues

### DIFF
--- a/collection-scripts/gather_core_dumps
+++ b/collection-scripts/gather_core_dumps
@@ -2,33 +2,39 @@
 BASE_COLLECTION_PATH="must-gather"
 CORE_DUMP_PATH=${OUT:-"${BASE_COLLECTION_PATH}/node_core_dumps"}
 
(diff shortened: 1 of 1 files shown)


astoycos:
assign/@davemulford Thanks for all your help so far, if you could check this edit out i'd appreciate it!

astoycos:
@davemulford Thanks for all your help so far, if you could check this edit out i'd appreciate it!

PR #179, opened 2020-09-21
Bug 1877428: Update `gather_core_dumps` to fix unexpected behavior in CI

astoycos:
The previous iteration of `gather_core_dumps` broke
CI because some unexpected behaviour was exhibited
where `oc debug` would make a new namespace for
the `debug` pod that we use to copy the coredumps
off each node

This commit fixes that by forcing `oc debug` to spawn
it's debug pod in the `default` namespace by using
the --to-namespace flag

It also adds a bit of extra logging

### DIFF
--- a/collection-scripts/gather_core_dumps
+++ b/collection-scripts/gather_core_dumps
@@ -6,28 +6,27 @@ mkdir -p "${CORE_DUMP_PATH}"/
 
 function get_dump_off_node {
     local debugPod=""
-    local debugNamespace=""
+    
+    #Get debug pod's name
+    debugPod=$(oc debug --to-namespace="default" node/"$1" -o jsonpath='{.metadata.name}')
 
-    #Get debug pod's name and
-    debugPod=$(oc debug node/"$1" -o jsonpath='{.metadata.name}')
-    debugNamespace=$(oc debug node/"$1" -o jsonpath='{.metadata.namespace}')
-
-    #Start Debug pod force it to stay up for 5 minutes max
-    oc debug node/"$1" -- /bin/bash -c 'sleep 300' > /dev/null 2>&1 &
+    #Start Debug pod force it to stay up until removed in "default" namespace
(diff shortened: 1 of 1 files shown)


collection-scripts/gather_core_dumps
```diff
@@ -6,28 +6,27 @@ mkdir -p "${CORE_DUMP_PATH}"/
...
-
-    #Start Debug pod force it to stay up for 5 minutes max
-    oc debug node/"$1" -- /bin/bash -c 'sleep 300' > /dev/null 2>&1 &
+    #Start Debug pod force it to stay up until removed in "default" namespace
+    oc debug --to-namespace="default" node/"$1" -- /bin/bash -c 'sleep 300' > /dev/null 2>&1 &
 
     #Mimic a normal oc call, i.e pause between two successive calls to allow pod to register
     sleep 2
-    oc wait -n "$debugNamespace" --for=condition=Ready pod/"$debugPod" --timeout=10s
+    oc wait -n "default" --for=condition=Ready pod/"$debugPod" --timeout=30s
```
davemulford, line 18: Are you consistently seeing this timeout reached or exceeded?  It may be worth running must-gather a few times to get an average pod spin up time, then add about 15 seconds to try and account for slow customer environments?

davemulford (commented):
A single question about timeouts.

collection-scripts/gather_core_dumps
astoycos: So @davemulford  I was seeing an average pod spinup time of less than 5 seconds, i.e it is usually really fast, I extended the time here just in case we ran into some very slow environments, I can reduce it to 20 if you think that would be a bit better

davemulford: I think it's good as-is, then.  The command will return before the timeout if the pod is ready, I was making sure we were accounting for slower environments (and it sounds like we are).

PR #182, opened 2020-10-06
Add olm resources to default must-gather

ecordell:
We've had some requests for this - it may be controversial, but this seemed as good a place as any to collect feedback.

### DIFF
--- a/collection-scripts/gather
+++ b/collection-scripts/gather
@@ -33,10 +33,12 @@ resources+=(imagecontentsourcepolicies.operator.openshift.io)
 resources+=(networks.operator.openshift.io)
 
 # Run the Collection of Resources using inspect
-# running accross all-namespaces for the few "Autoscaler" resouces.
 for resource in "${resources[@]}"; do
     oc adm inspect --dest-dir must-gather ${resource}
 done
+
+# Gather optional operator resources from all namespaces
+/usr/bin/gather_olm
 
 # Gather etcd information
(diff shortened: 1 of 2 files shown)


collection-scripts/gather
```diff
@@ -33,11 +33,13 @@ resources+=(imagecontentsourcepolicies.operator.openshift.io)
 resources+=(networks.operator.openshift.io)
 
 # Run the Collection of Resources using inspect
-# running accross all-namespaces for the few "Autoscaler" resouces.
 for resource in "${resources[@]}"; do
     oc adm inspect --dest-dir must-gather ${resource}
 done
 
+# Gather optional operator resources from all namespaces
```
sferich888, line 40: IMO this should go into its own script (encase we ever need to add/filter things). Calling it by default from this script is fine (provided the collection is not astronomical amounts of data/information).

collection-scripts/gather
```diff
@@ -33,11 +33,13 @@ resources+=(imagecontentsourcepolicies.operator.openshift.io)
...
+oc adm inspect --dest-dir must-gather -A olm
```
sferich888, line 41: `-A` is limited to what `oc` versions (we need to make sure we don't back-port this past that point (and this introduce errors/failures in the collection process). 

cc: @mfojtik

sferich888 (approved):
/lgtm

PR #183, opened 2020-10-06
Bug 1887446: [release-4.5]Backport `gather_core_dumps` 

astoycos:
Backport `gather_core_dump` Functionality to OCP 4.5 for CI Runs that use the functionality, and also bring in port and flow dumping in `gather_network_logs`

### DIFF
--- a/collection-scripts/gather_core_dumps
+++ b/collection-scripts/gather_core_dumps
@@ -1,0 +1,24 @@
+#!/bin/bash
+BASE_COLLECTION_PATH="must-gather"
+CORE_DUMP_PATH=${OUT:-"${BASE_COLLECTION_PATH}/node_core_dumps"}
+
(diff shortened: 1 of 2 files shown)


davemulford:
@astoycos the changes look okay, but this needs a valid bug in order to be merged. Do you have one?

astoycos:
@davemulford Sorry about that should be all set now :)

davemulford:
No worries!

/lgtm

davemulford:
@sdodson, if you agree with this change, can you please add the cherry-pick-approved label?

sdodson:
> @sdodson, if you agree with this change, can you please add the cherry-pick-approved label?

patch manager is a rotation and they only accept changes when the window is open so they should get to this according to that schedule

PR #189, opened 2020-10-28
NE-356: Add HAProxy config files to must-gather

sgreene570:
Add script file collection-scripts/gather_haproxy_config to collect haproxy config files for each ingress controller in the cluster.

This is in support of https://issues.redhat.com/browse/NE-356.

### DIFF
--- a/collection-scripts/gather
+++ b/collection-scripts/gather
@@ -57,5 +57,8 @@ oc adm inspect --dest-dir must-gather "${group_resources_text}"
 # Gather Windows Kubernetes component logs
 /usr/bin/gather_windows_node_logs
 
+# Gather HAProxy config files
+/usr/bin/gather_haproxy_config
+
 # force disk flush to ensure that all data gathered is accessible in the copy container
 sync
--- a/collection-scripts/gather_haproxy_config
+++ b/collection-scripts/gather_haproxy_config
@@ -1,0 +1,28 @@
+#!/bin/bash
+
+BASE_COLLECTION_PATH="must-gather"
+HAPROXY_CONFIG_PATH=${OUT:-"${BASE_COLLECTION_PATH}/ingress_controllers"}
+
+mkdir -p "${HAPROXY_CONFIG_PATH}"/
+
+function gather_haproxy_config {
+    echo "INFO: Gathering HAProxy config files"
+    INGRESS_CONTROLLERS=$(oc get ingresscontroller -n openshift-ingress-operator --no-headers -o custom-columns=":metadata.name")
+    for IC in ${INGRESS_CONTROLLERS}; do
+        PODS=$(oc get pods -n openshift-ingress --no-headers -o custom-columns=":metadata.name" --selector ingresscontroller.operator.openshift.io/deployment-ingresscontroller="${IC}")
+        for POD in ${PODS}; do
+            oc cp openshift-ingress/"${POD}":/var/lib/haproxy/conf/haproxy.config "${HAPROXY_CONFIG_PATH}"/"${IC}"/"${POD}"/haproxy.config &
+            PIDS+=($!)
+        done
+    done
+}
+
+PIDS=()
+gather_haproxy_config
+
+echo "INFO: Waiting for HAProxy config collection to complete ..."
+wait "${PIDS[@]}"
+echo "INFO: HAProxy config collection complete."
+
+# force disk flush to ensure that all data gathered is accessible in the copy container
+sync


collection-scripts/gather_haproxy_config
```diff
@@ -0,0 +1,22 @@
...
+HAPROXY_CONFIG_PATH=${OUT:-"${BASE_COLLECTION_PATH}/haproxy"}
+
+mkdir -p "${HAPROXY_CONFIG_PATH}"/
+
+function gather_haproxy_config {
+    echo "INFO: Gathering HAProxy config files"
+    INGRESS_CONTROLLERS=$(oc get ingresscontroller -n openshift-ingress-operator --no-headers -o custom-columns=":metadata.name")
+    for IC in ${INGRESS_CONTROLLERS}; do
+        POD_STRING=$(oc get pods -n openshift-ingress --no-headers -o custom-columns=":metadata.name" --selector ingresscontroller.operator.openshift.io/deployment-ingresscontroller="${IC}")
+        PODS=($POD_STRING)
```
Miciah, line 13: Just a suggestion:  You could use `-o name`, although that has the disadvantage that you need to strip off the leading resource qualifier (e.g., `oc get ingresscontrollers -o name` prints "ingresscontroller.operator.openshift.io/default", and you want just "default"):
```suggestion
    INGRESS_CONTROLLERS=$(oc get ingresscontroller -n openshift-ingress-operator -o name)
    for IC in ${INGRESS_CONTROLLERS}; do
        PODS=($(oc get pods -n openshift-ingress -o name -l ingresscontroller.operator.openshift.io/deployment-ingresscontroller="${IC#*/}"))
```

collection-scripts/gather_haproxy_config
```diff
@@ -0,0 +1,22 @@
...
+        # Use first pod per ingress controller
+        oc cp openshift-ingress/"${PODS[0]}":/var/lib/haproxy/conf/haproxy.config "${HAPROXY_CONFIG_PATH}"/"${IC}"/haproxy.config
```
Miciah, line 15: It might be useful to have `haproxy.config` from each pod in some cases.

Miciah, line 15: You did `mkdir -p "${HAPROXY_CONFIG_PATH}"/` earlier, but you didn't `mkdir "${HAPROXY_CONFIG_PATH}"/"${IC}"/`...  But all right, it looks like `oc` takes care of making the directory: https://github.com/openshift/oc/blob/1bd1f5ff4aa98d6e4184c2948928b6111fa99191/vendor/k8s.io/kubectl/pkg/cmd/cp/cp.go#L466
So I think this is fine.

collection-scripts/gather_haproxy_config
sgreene570: Other scripts in `must-gather` seem to more often use the `columns=...` approach, so i'll stick with the current approach for consistency sake. That being said, I will wrap the line in `()` to assign the values as an array. Thanks!

collection-scripts/gather_haproxy_config
sgreene570: Right, I was going back on fourth on this. But you are correct, in the event of 1/n routers becoming wedged, the configs may be different (and having all of them would help with debugging this case).

sgreene570: Yea, I noticed the oc command did the directory creations for me, so I decided that I didn't need to explicitly make them.

sgreene570: `7d138e0` gathers each pod's HAProxy config using a for loop.

davemulford:
@sferich888 is this something we want to happen for every must-gather that gets run?

sgreene570:
> @sferich888 is this something we want to happen for every must-gather that gets run?

Any particular reason we would not want this to run for every must-gather? Do you have particular concerns in mind?

knobunc (approved):
/approve

collection-scripts/gather_haproxy_config
```diff
@@ -0,0 +1,22 @@
...
+
+mkdir -p "${HAPROXY_CONFIG_PATH}"/
+
+function gather_haproxy_config {
+    echo "INFO: Gathering HAProxy config files"
+    INGRESS_CONTROLLERS=$(oc get ingresscontroller -n openshift-ingress-operator --no-headers -o custom-columns=":metadata.name")
+    for IC in ${INGRESS_CONTROLLERS}; do
+        PODS=$(oc get pods -n openshift-ingress --no-headers -o custom-columns=":metadata.name" --selector ingresscontroller.operator.openshift.io/deployment-ingresscontroller="${IC}")
+        for POD in ${PODS}; do
+            oc cp openshift-ingress/"${POD}":/var/lib/haproxy/conf/haproxy.config "${HAPROXY_CONFIG_PATH}"/"${IC}"/"${POD}"/haproxy.config
```
sferich888, line 14: In the 3.x days (and now even) we also collected, things that affected the control loop. Should we also collect those here? 

https://access.redhat.com/solutions/1542293#routing

collection-scripts/gather_haproxy_config
```diff
@@ -0,0 +1,22 @@
...
+BASE_COLLECTION_PATH="must-gather"
+HAPROXY_CONFIG_PATH=${OUT:-"${BASE_COLLECTION_PATH}/haproxy"}
+
+mkdir -p "${HAPROXY_CONFIG_PATH}"/
+
+function gather_haproxy_config {
+    echo "INFO: Gathering HAProxy config files"
+    INGRESS_CONTROLLERS=$(oc get ingresscontroller -n openshift-ingress-operator --no-headers -o custom-columns=":metadata.name")
+    for IC in ${INGRESS_CONTROLLERS}; do
+        PODS=$(oc get pods -n openshift-ingress --no-headers -o custom-columns=":metadata.name" --selector ingresscontroller.operator.openshift.io/deployment-ingresscontroller="${IC}")
```
sferich888, line 12: Should we thread this (try an make all these pod collection calls in parallel vs serial)?

collection-scripts/gather_haproxy_config
```diff
@@ -0,0 +1,22 @@
+#!/bin/bash
+
+BASE_COLLECTION_PATH="must-gather"
+HAPROXY_CONFIG_PATH=${OUT:-"${BASE_COLLECTION_PATH}/haproxy"}
```
sferich888, line 4: Do we want to put this under, ingress-controllers? What if we move away from haproxy in the future?

collection-scripts/gather_haproxy_config
```diff
@@ -0,0 +1,22 @@
...
+
+mkdir -p "${HAPROXY_CONFIG_PATH}"/
+
+function gather_haproxy_config {
+    echo "INFO: Gathering HAProxy config files"
+    INGRESS_CONTROLLERS=$(oc get ingresscontroller -n openshift-ingress-operator --no-headers -o custom-columns=":metadata.name")
+    for IC in ${INGRESS_CONTROLLERS}; do
+        PODS=$(oc get pods -n openshift-ingress --no-headers -o custom-columns=":metadata.name" --selector ingresscontroller.operator.openshift.io/deployment-ingresscontroller="${IC}")
+        for POD in ${PODS}; do
```
sferich888, line 13: Do we need to make this loop configurable, what happens if the ingress-controller pod is not an haproxy pod? Do we need to make this configurable or integrated with the pod (some data collection convention), so that this works with 3rd party ingress-controllers?

collection-scripts/gather
```diff
@@ -57,5 +57,8 @@ oc adm inspect --dest-dir must-gather "${group_resources_text}"
 # Gather Windows Kubernetes component logs
 /usr/bin/gather_windows_node_logs
 
+# Gather HAProxy config files
+/usr/bin/gather_haproxy_config
```
sferich888, line 61: Do we want this collected by default?

sferich888 (approved):
I think this is fine (in general), but I wonder if we should consider a few possible improvements/alterations (possibly for a later date/time).

collection-scripts/gather_haproxy_config
sgreene570: 3rd party ingress controllers wouldn't be able to make use of the OCP ingress controller API, and this scripts selects `PODS` by OCP ingress controller, so I think we are OK here.

collection-scripts/gather_haproxy_config
sgreene570: I believe must-gather already collects everything else for the router pods that would make sense to have.

sgreene570:
Pushing some fixes shortly
/hold 
(since this already has lgtm and approved tags)

collection-scripts/gather_haproxy_config
sgreene570: Fixed.

collection-scripts/gather_haproxy_config
sgreene570: change to `ingress_controllers`

collection-scripts/gather
sgreene570: Any reason not to? Do you feel that this should not be collected by default?

sferich888: I don't like collecting things in this manner (by default), but .. in this situations I don't think we have a better solution.

collection-scripts/gather_haproxy_config
sferich888: Thanks.

PR #194, opened 2020-11-16
Remove namespace

astoycos:
The Way information was collected here for EgressIP,
EgressFirewall and NetworkPolicies was not scalable
and showed some problems under scale testing

Instead those objects will be included under
networking's related objects so that `oc adm inspect`
will automatically collect the required data


astoycos:
FYI @vpickard

PR #198, opened 2020-12-04
Bug 1906968: Add kubernetes-nmstate resources to must-gather

bcrochet:
If kubernetes-nmstate is on a cluster, these resources would be helpful
to ascertain what possible network configurations on present on nodes.


bcrochet:
It seems that if the resources don't exist, the gather script fails. This needs to be investigated. So rather than waste more resources testing every hour... 

/hold

PR #199, opened 2020-12-09
Bug 1906655: Add openvswitch services to must-gather

juanluisvaladas:
We will remove the ovs pod, therefore we need must-gather to attach these logs in the sosreport.
/assign @squeed

### DIFF
--- a/collection-scripts/collect_service_logs
+++ b/collection-scripts/collect_service_logs
@@ -1,18 +1,30 @@
 #!/bin/bash
-function collect_service_logs {  ## Takes a node role input (master or worker) and a list of services
-    role=${1}
+function collect_service_logs {
+# Takes a node role input (master or worker) and a list of services
+# assumes there is a SERVICE_LOG_PATH variable defined used to set the logs
+# destination and a PIDS variable which is an array.
+
+    selector=${1}
     shift
 
-    PIDS=()
-    DIR_PATH=${SERVICE_LOG_PATH}/${role}s
-    echo "WARNING: Collecting one or more service logs on ALL linux $1 nodes in your cluster. This could take a large amount of time." >&2
+    if [[ ${selector} == '--role=master' ]]; then
+        DIR_PATH="${SERVICE_LOG_PATH}/masters"
+    elif [[ ${selector} == '--role=worker' ]]; then
+        DIR_PATH="${SERVICE_LOG_PATH}/workers"
+        echo "WARNING: Collecting one or more service logs on ALL linux $1 workers in your cluster. This could take a large amount of time." >&2
+    else
+        DIR_PATH="${SERVICE_LOG_PATH}/${selector}"
+    fi
+
     mkdir -p "${DIR_PATH}"
     for service in "${@}"; do
         echo "INFO: Collecting host service logs for $service"
-        /usr/bin/oc adm node-logs --role=${role} -l kubernetes.io/os=linux -u "${service}" --since "${SINCE_TIMEFRAME:--7d}"> "${DIR_PATH}/${service}_service.log" &
-	PIDS+=($!)
+        if [[ ${selector} =~ '--role=' ]]; then
+            /usr/bin/oc adm node-logs ${selector} -l kubernetes.io/os=linux -u "${service}" --since "${SINCE_TIMEFRAME:--7d}"> "${DIR_PATH}/${service}_service.log" &
+	    PIDS+=($!)
+        else
+            /usr/bin/oc adm node-logs ${selector} -u "${service}" --since "${SINCE_TIMEFRAME:--7d}"> "${DIR_PATH}/${service}_service.log" &
+	    PIDS+=($!)
+        fi
     done
-    echo "INFO: Waiting for worker host service log collection to complete ..."
-    wait "${PIDS[@]}"
-    echo "INFO: Worker host service log collection to complete."
 }
--- a/collection-scripts/collect_service_logs
+++ b/collection-scripts/collect_service_logs
@@ -1,0 +1,18 @@
+#!/bin/bash
+function collect_service_logs {  ## Takes a node role input (master or worker) and a list of services
+    role=${1}
+    shift
+
+    PIDS=()
+    DIR_PATH=${SERVICE_LOG_PATH}/${role}s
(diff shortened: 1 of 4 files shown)


squeed:
Seems fine to me, but if the OWNERS want this in `gather_network_logs` instead, I'd understand.
/lgtm

davemulford:
I think these should go under the gather_network_logs script, but will defer to @sferich888.

/assign @sferich888

sferich888:
@davemulford given that OVN is now a systemd unit I could get behind this being in the service logs collector (as it is now) if we only wanted to collect this for masters, due to its limited scoping. However after talking with @juanluisvaladas we want this for all nodes or select nodes, so moving this to gather_network_logs seems like the right approach. 

An alternative would/could be to rewrite the service collector, to better suite this use-cases. It wasn't really written to take into account what we are now doing / what to achieve with this request.

juanluisvaladas:
/hold

This is still missing one commit, I'm using CI to test that the change is really working as I expect.

juanluisvaladas:
After speaking with @sferich888 we decided that the most logical place for this was the gather_network_logs
I tested manually that  gather_network_logs adds the new logs and the gather_service_logs works like it used to

collection-scripts/collect_service_logs
```diff
@@ -0,0 +1,18 @@
+#!/bin/bash
```
sferich888, line 1: This implementation doesn't take into account how how scripts like https://github.com/openshift/must-gather/blob/b308b60f9501006342e1fdd918961cf0ae9a4f96/collection-scripts/gather_network_logs#L221 supply limited like node lists (to run the collection on). 

This could confuse users, as host logs could be collected from all nodes, while only some pod based operations are only run on specific nodes. I think we need to explore how to account for this.

collection-scripts/collect_service_logs
```diff
@@ -0,0 +1,18 @@
...
+function collect_service_logs {  ## Takes a node role input (master or worker) and a list of services
+    selector=${1}
+    shift
+
+    PIDS=()
+    DIR_PATH=${SERVICE_LOG_PATH}/${role}s
+    echo "WARNING: Collecting one or more service logs on ALL linux $1 nodes in your cluster. This could take a large amount of time." >&2
```
sferich888, line 8: I think this will fire all the time (even if your only collecting on masters, which isn't something we want).

sferich888 (changes requested):
This generalization help move us closer to where I think we want this to be, however we may still be missing a few edge cases/user considerations.

collection-scripts/collect_service_logs
juanluisvaladas: Good point! this is fixed now.
I tested this with: `oc adm must-gather --image=quay.io/jdesousa/must-gather:journalovs gather_network_logs ci-ln-g7g54tb-f76d1-lrzkc-master-1 ci-ln-g7g54tb-f76d1-lrzkc-master-2` and just acquired  it for these two nodes.

collection-scripts/collect_service_logs
juanluisvaladas: Now I only print the warning if --role=workers

juanluisvaladas:
Solved both your concerns, it should be good to go!

juanluisvaladas:
/hold
I just noticed I'm also responsible for this task on OVN, so I will add yet another commit to fix this in OVN as . The hard work should  be already done so I'll be quick. I'll handle this after lunch

juanluisvaladas:
/hold cancel 

OVN is also patched and tested. Everything keeps working without invoking the ovs pod.

juanluisvaladas:
/retitle Bug 1907345: Add openvswitch services to must-gather
/retest

squeed:
oh nice, this looks awesome.
/lgtm

juanluisvaladas:
/hold
This is depends on https://github.com/openshift/cluster-network-operator/pull/876 which was delayed to 4.8.

juanluisvaladas:
/retitle Bug 1906655: Add openvswitch services to must-gather

@sferich888 turns out there is a problem with a must-gather so this is now a bugfix instead of a feature.

huiran0826:
/lgtm

Verified the bug 1906655 with this PR using cluster-bot, looks good to me.

oc adm must-gather -- gather_network_logs
.......
[must-gather-7mxw9] POD INFO: Collecting host service logs for openvswitch
[must-gather-7mxw9] POD INFO: Collecting host service logs for ovs-vswitchd
[must-gather-7mxw9] POD INFO: Collecting host service logs for ovsdb-server
[must-gather-7mxw9] POD INFO: Collecting host service logs for ovs-configuration
........

In the directory  host_service_logs,
[fedora@preserved-huirwang ci-ln-101jts2-f76d1-2trpb-master-0]$ ls
openvswitch_service.log  ovs-configuration_service.log  ovs-vswitchd_service.log  ovsdb-server_service.log
[fedora@preserved-huirwang ci-ln-101jts2-f76d1-2trpb-worker-c-lwr47]$ ls
ovs-configuration_service.log  ovsdb-server_service.log       ovs-vswitchd_service.log

astoycos:
@davemulford some eyes on this when you get the chance would be great, thanks!

davemulford:
@astoycos thanks for highlighting me. I plan to discuss this with @sferich888 later today.

astoycos:
cc @jluhrsen

davemulford:
@juanluisvaladas and @astoycos:

I spoke with @sferich888 about this and we like the scripts and what they do, but would like to see the `collect_service_logs` script renamed to avoid confusion with `gather_service_logs`. Specifically, we'd like to see it renamed to emphasize that it is a utility script that shouldn't be called directly (e.g. `oc adm must-gather -- collect_service_logs` shouldn't happen).

Our recommendation is one of the following.

- Rename the file to `gather_service_logs_util` with comments at the top explaining its purpose.
- Move the file into a `utils` directory and add a README explaining the purpose of scripts in that directory.

juanluisvaladas:
@davemulford fair enough, I renamed it to gather_service_logs_util and added the header.
Also the file doesn't have execution rights, so `oc adm must-gather gather_service_logs_util` won't work at all even if passing the right arguments.

davemulford:
Thanks @juanluisvaladas, this looks good!

/lgtm

PR #200, opened 2020-12-18
BZ 1906655: Fix `gather-network-logs` SDN gathering

astoycos:
Since moving the OVS process out of the containers
and back to the host the locations of `ovs-vswitchd.log`
and `ovsdb-server.log` has changed, this commit fixes
that and uses `oc cp` instead of `oc exec`


astoycos:
I need to talk to @juanluisvaladas regarding https://github.com/openshift/must-gather/pull/199

astoycos:
closed in favor of #199

PR #204, opened 2021-01-15
Bug 1916392: Gather API priority and fairness debug endpoints

zerodayz:
API Priority and Fairness exposes a set of debug endpoints (HTTP) that could help us troubleshoot p&f related issues.

Closes: BZ#1916392

### DIFF
--- a/collection-scripts/gather_priority_and_fairness
+++ b/collection-scripts/gather_priority_and_fairness
@@ -1,0 +1,85 @@
+#!/bin/bash
+# We don't want to collect /debug/api_priority_and_fairness in oc adm inspect,
+# https://github.com/openshift/oc/blob/master/pkg/cli/admin/inspect/pod.go#L58-L62
+# because the tokens could be used in a replay attack,
+# thus the route we are taking here is forced by this.
+
+# Safeguards
+set -o nounset
+set -o errexit
+set -o pipefail
+
+BASE_COLLECTION_PATH="/must-gather"
+NAMESPACE_PATH="namespaces/openshift-kube-apiserver"
+POD_PATH="kube-apiserver/kube-apiserver/api_priority_and_fairness"
+APF_PATH="${BASE_COLLECTION_PATH}/${NAMESPACE_PATH}"
+
+# If the cluster version is < 4.6 skip this test
+
+# If cluster version is >=4.6 then the COMPARE will always == 4.6
+#  For example cluster version 4.7:
+#  VERSION=4.7
+#  COMPARE=$(echo -e "4.6\n${VERSION}" | sort -V | head -n1)
+#  echo $COMPARE
+#  4.6
+
+# If cluster version is <4.6 then the COMPARE will be the always the cluster version.
+#  VERSION=4.4
+#  COMPARE=$(echo -e "4.6\n${VERSION}" | sort -V | head -n1)
+#  echo $COMPARE
+#  4.4
+
+VERSION=$(oc get clusterversion version -o jsonpath='{.status.desired.version}')
+COMPARE=$(echo -e "4.6\n${VERSION}" | sort -V | head -n1)
+if [[ ${COMPARE} != "4.6" ]]; then
+  echo "INFO: Cluster version is ${VERSION}. Skipping."
+  exit 0
+fi
+
+# Fetches the POD_NAME where app label is apiserver.
+# Example:
+#  oc get po -n openshift-kube-apiserver -l apiserver -o jsonpath='{range .items[*]}{.metadata.name}{"\n"}'
+#  kube-apiserver-master-0.example.com
+#  kube-apiserver-master-1.example.com
+#  kube-apiserver-master-2.example.com
+KUBE_API_SERVERS=$(oc get po -n openshift-kube-apiserver -l apiserver -o jsonpath='{range .items[*]}{.metadata.name}{"\n"}')
+if [[ -z ${KUBE_API_SERVERS} ]]; then
+    echo "ERROR: No running kube-apiserver pods found" >&2
+    exit 1
+fi
+
+for API_SERVER in ${KUBE_API_SERVERS}
+do
+  OUTPUT_DIR="${APF_PATH}/pods/${API_SERVER}/${POD_PATH}"
+  mkdir -p "${OUTPUT_DIR}"
+
+  # Fetches the POD_IP from the apiserver pod.
+  # Example:
+  #  oc get po -n openshift-kube-apiserver kube-apiserver-master-0.example.com -o jsonpath='{.status.podIP}'
+  #  192.168.9.43
+  POD_IP=$(oc get po -n openshift-kube-apiserver "${API_SERVER}" -o jsonpath='{.status.podIP}')
+  if [[ -z ${POD_IP} ]]; then
+      echo "ERROR: ${API_SERVER}: unable to get .status.podIP" >&2
+      continue
+  fi
+
+  # Fetches the POD_PORT where container name is kube-apiserver.
+  # Example:
+  #  oc get po -n openshift-kube-apiserver kube-apiserver-master-0.example.com -o jsonpath='{range .spec.containers[?(.name=="kube-apiserver")]}{.ports[0].hostPort}'
+  #  6443
+  POD_PORT=$(oc get po -n openshift-kube-apiserver "${API_SERVER}" -o jsonpath='{range .spec.containers[?(.name=="kube-apiserver")]}{.ports[0].hostPort}')
+  if [[ -z ${POD_PORT} ]]; then
+      echo "ERROR: ${API_SERVER}: unable to get .spec.containers.ports[0].hostPort" >&2
+      continue
+  fi
+
+  echo "INFO: Collecting /debug/api_priority_and_fairness endpoint from ${API_SERVER}"
+
+  oc get --raw https://"${POD_IP}":"${POD_PORT}"/debug/api_priority_and_fairness/dump_priority_levels \
+               > "${OUTPUT_DIR}"/priority_levels
+  oc get --raw https://"${POD_IP}":"${POD_PORT}"/debug/api_priority_and_fairness/dump_queues \
+               > "${OUTPUT_DIR}"/queues
+  oc get --raw https://"${POD_IP}":"${POD_PORT}"/debug/api_priority_and_fairness/dump_requests \
+               > "${OUTPUT_DIR}"/requests
+
+done
--- a/collection-scripts/gather
+++ b/collection-scripts/gather
@@ -48,6 +48,9 @@ oc adm inspect --dest-dir must-gather "${group_resources_text}"
 # Gather optional operator resources from all namespaces
 /usr/bin/gather_olm
 
+# Gather API Priority and Fairness Endpoints
+/usr/bin/gather_priority_and_fairness
+
 # Gather etcd information
 /usr/bin/gather_etcd
 


zerodayz:
/hold

checking with API team for detailed endpoint that may need to be added.

collection-scripts/gather_priority_and_fairness
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
+BASE_COLLECTION_PATH="/must-gather"
+APF_PATH="${BASE_COLLECTION_PATH}/api_priority_and_fairness"
+
+mkdir -p ${APF_PATH}
+oc get --raw /debug/api_priority_and_fairness/dump_priority_levels > ${APF_PATH}/dump_priority_levels
```
p0lyn0mial, line 6: will this collect data from all API servers or just one?

p0lyn0mial: based on must-gather from https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/openshift_must-gather/204/pull-ci-openshift-must-gather-master-e2e-aws/1350052523385294848/artifacts/e2e-aws/gather-must-gather/ it looks like it is just for a single API.

p0lyn0mial: It would be nice if we could get data from all API servers.

Would it be easy to place the files in `namespaces/openshift-kube-apiserver/pods/kube-apiserver/kube-apiserver/api_priority_and_fairness` ? (for each instance)

zerodayz: Good morning p0lyn0mial! I will look at what is being done for other scripts and if it is possible and update the patch. Thank you for your review!

zerodayz: @p0lyn0mial What do you think about it now? It should collect all API Servers, I am filtering on role=master, then scraping the endpoints from all of them in separated directories under` namespaces/openshift-kube-apiserver/pods/kube-apiserver/kube-apiserver/api_priority_and_fairness`

It will create corresponding directories based on API Server targeted.

```
master-1.rcernin46.lab.upshift.rdu2.redhat.com
master-0.rcernin46.lab.upshift.rdu2.redhat.com
master-2.rcernin46.lab.upshift.rdu2.redhat.com

```
And each of those contains:

```
ls -la master-1.rcernin46.lab.upshift.rdu2.redhat.com
total 72
drwxr-xr-x  5 rrobin  staff   160 16 Jan 15:14 .
drwxr-xr-x  6 rrobin  staff   192 16 Jan 15:14 ..
-rw-r--r--  1 rrobin  staff  9796 16 Jan 15:14 dump_priority_levels
-rw-r--r--  1 rrobin  staff  9796 16 Jan 15:14 dump_queues
-rw-r--r--  1 rrobin  staff  9796 16 Jan 15:14 dump_requests
```

zerodayz: Unsure why the Tests failed but here it is how it looks now: https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/openshift_must-gather/204/pull-ci-openshift-must-gather-master-e2e-aws/1350313571078115328/artifacts/e2e-aws/gather-must-gather/


```
pwd
/must-gather-2/registry-build02-ci-openshift-org-ci-op-q54bk8m9-stable-sha256-9d47fc7795d6366555a7b0c2f7d665879bb86e0360b11ac5c321b0999e4085ec/namespaces/openshift-kube-apiserver/pods/kube-apiserver/kube-apiserver

kube-apiserver ls
api_priority_and_fairness

ls -la **/*

api_priority_and_fairness:
total 0
drwxr-xr-x@ 5 rrobin  staff  160 16 Jan 16:50 .
drwxr-xr-x@ 4 rrobin  staff  128 16 Jan 17:09 ..
drwxr-xr-x@ 5 rrobin  staff  160 16 Jan 16:50 ip-10-0-128-30.us-west-2.compute.internal
drwxr-xr-x@ 5 rrobin  staff  160 16 Jan 16:50 ip-10-0-187-34.us-west-2.compute.internal
drwxr-xr-x@ 5 rrobin  staff  160 16 Jan 16:50 ip-10-0-201-215.us-west-2.compute.internal

api_priority_and_fairness/ip-10-0-128-30.us-west-2.compute.internal:
total 72
drwxr-xr-x@ 5 rrobin  staff    160 16 Jan 16:50 .
drwxr-xr-x@ 5 rrobin  staff    160 16 Jan 16:50 ..
-rw-r--r--@ 1 rrobin  staff  11399 16 Jan 16:50 dump_priority_levels
-rw-r--r--@ 1 rrobin  staff  11399 16 Jan 16:50 dump_queues
-rw-r--r--@ 1 rrobin  staff  11399 16 Jan 16:50 dump_requests

api_priority_and_fairness/ip-10-0-187-34.us-west-2.compute.internal:
total 72
drwxr-xr-x@ 5 rrobin  staff    160 16 Jan 16:50 .
drwxr-xr-x@ 5 rrobin  staff    160 16 Jan 16:50 ..
-rw-r--r--@ 1 rrobin  staff  11399 16 Jan 16:50 dump_priority_levels
-rw-r--r--@ 1 rrobin  staff  11399 16 Jan 16:50 dump_queues
-rw-r--r--@ 1 rrobin  staff  11399 16 Jan 16:50 dump_requests

api_priority_and_fairness/ip-10-0-201-215.us-west-2.compute.internal:
total 72
drwxr-xr-x@ 5 rrobin  staff    160 16 Jan 16:50 .
drwxr-xr-x@ 5 rrobin  staff    160 16 Jan 16:50 ..
-rw-r--r--@ 1 rrobin  staff  11399 16 Jan 16:50 dump_priority_levels
-rw-r--r--@ 1 rrobin  staff  11399 16 Jan 16:50 dump_queues
-rw-r--r--@ 1 rrobin  staff  11399 16 Jan 16:50 dump_requests
```

zerodayz: Hm it actually failed to get those /debug endpoints, I thought they were exposed. I ll check why it didn't work.

zerodayz: Check now, I think it should be all good ? https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/openshift_must-gather/204/pull-ci-openshift-must-gather-master-e2e-aws/1350343532023386112/artifacts/e2e-aws/gather-must-gather/

zerodayz:
One thing I am not happy about is to hardcode the port value "8443" and "https", I am not sure how to do that without it. Maybe some of the reviewers can help with that?

/assign @sferich888

collection-scripts/gather_priority_and_fairness
p0lyn0mial: > `namespaces/openshift-kube-apiserver/pods/kube-apiserver/kube-apiserver/api_priority_and_fairness`

I would prefer under `namespaces/openshift-kube-apiserver/pods/kube-apiserver-ip-10-0-170-247.us-east-2.compute.internal/kube-apiserver/kube-apiserver` or `namespaces/openshift-kube-apiserver/pods/kube-apiserver-ip-10-0-170-247.us-east-2.compute.internal/kube-apiserver/kube-apiserver/api_priority_and_fairness`

p0lyn0mial:
> One thing I am not happy about is to hardcode the port value "8443" and "https", I am not sure how to do that without it. Maybe some of the reviewers can help with that?

I think that you could read IPs and the port number from the standard `kubernetes` service. 

`oc get endpoints -n default kubernetes -oyaml`
```
subsets:
- addresses:
  - ip: 10.0.142.76
  - ip: 10.0.148.59
  - ip: 10.0.173.241
  ports:
  - name: https
    port: 6443
    protocol: TCP
```

collection-scripts/gather_priority_and_fairness
```diff
@@ -0,0 +1,21 @@
+#!/bin/bash
+BASE_COLLECTION_PATH="/must-gather"
+NAMESPACE_PATH="namespaces/openshift-kube-apiserver"
+POD_PATH="pods/kube-apiserver/kube-apiserver"
+APF_PATH="${BASE_COLLECTION_PATH}/${NAMESPACE_PATH}/${POD_PATH}/api_priority_and_fairness"
+
+mkdir -p ${APF_PATH}
+MASTER_NODES=$(oc get endpoints -n default kubernetes -o jsonpath='{.subsets[*].addresses [*].ip}')
```
p0lyn0mial, line 8: nit: I would rename it to `KUBE_API_SERVERS`

p0lyn0mial:
@zerodayz could we change the names to?

`priority_levels`
`queues`
`requests`

@zerodayz many thanks for your PR, in general, it looks good to me. We only need to find a good place for the new files. Let's wait for input from @tkashem as well.

collection-scripts/gather_priority_and_fairness
zerodayz: Done

zerodayz:
Changed, agreed let's wait for reviewers.

Thank you @p0lyn0mial

tkashem:
yes, i agree with @p0lyn0mial in that we should collect from all three api servers.

Also, p&f is enabled from 4.5. what happens if the server is < 4.5 or the feature gate is disabled, will the entire must-gather fail?

zerodayz:
Hi @tkashem in general if you query an endpoint thats not available, it will just give you list of available ones. /api, /apis so its harmless but I can add check for version and if we run on anything <4.5 will skip.

zerodayz:
@tkashem Can you please review again, thank you!

We skip if version is < 4.5. I couldn't find other more safe way than using sort -V

https://github.com/openshift/must-gather/pull/204/files#diff-659ec29e4d828b2a6d7332f939387a6583aa8269bb6e5b30fa18c96b9d21b7d4R7-R12

And iterating over all API Servers

https://github.com/openshift/must-gather/pull/204/files#diff-659ec29e4d828b2a6d7332f939387a6583aa8269bb6e5b30fa18c96b9d21b7d4R18-R27

collection-scripts/gather_priority_and_fairness
```diff
@@ -0,0 +1,28 @@
+#!/bin/bash
+BASE_COLLECTION_PATH="/must-gather"
+NAMESPACE_PATH="namespaces/openshift-kube-apiserver"
+POD_PATH="pods/kube-apiserver/kube-apiserver"
+APF_PATH="${BASE_COLLECTION_PATH}/${NAMESPACE_PATH}/${POD_PATH}/api_priority_and_fairness"
+
+# If the cluster version is < 4.5 skip this test
+VERSION=$(oc get clusterversion version -o jsonpath='{.status.desired.version}')
+COMPARE=$(echo -e "4.5\n${VERSION}" | sort -V | head -n1)
+if [[ ${COMPARE} != "4.5" ]]; then
```
p0lyn0mial, line 10: this PR is targeting `master` branch, not `4.5` branch. I think we don't have to take the cluster version into account

soltysh: I'd leave it, you can always use newer image with on an older cluster, so it won't hurt.

tkashem:
@p0lyn0mial @zerodayz this may be a follow up PR. what do you think of adding a must-gather script to capture the pprof and trace data for each apiserver? I don't think we collect it today, do we?

collection-scripts/gather_priority_and_fairness
zerodayz: Yes it does return just list of available endpoints in case its unavailable < 4.5. I did it as we discussed it but do we need to keep version check? Honestly it just adds unnecessary complexity and further requirements. What do you think? Shall I remove that part? It wont break must-gather not having it. @tkashem @p0lyn0mial @soltysh

p0lyn0mial: > do we need to keep version check

yes, based on the input from @soltysh we know that users might run arbitrary images (must-gather) against clusters.

p0lyn0mial:
> @p0lyn0mial @zerodayz this may be a follow up PR. what do you think of adding a must-gather script to capture the pprof and trace data for each apiserver? I don't think we collect it today, do we?

To be honest I haven't used them yet :) Could they be collected on demand? What's the typical size of pprof and trace data?

collection-scripts/gather_priority_and_fairness
```diff
@@ -0,0 +1,28 @@
+#!/bin/bash
+BASE_COLLECTION_PATH="/must-gather"
+NAMESPACE_PATH="namespaces/openshift-kube-apiserver"
+POD_PATH="pods/kube-apiserver/kube-apiserver"
+APF_PATH="${BASE_COLLECTION_PATH}/${NAMESPACE_PATH}/${POD_PATH}/api_priority_and_fairness"
```
p0lyn0mial, line 5: thinking more about the location, could we place to 

`openshift-kube-apiserver/pods/kube-apiserver-ip-10-0-145-102.us-east-2.compute.internal/kube-apiserverkube-apiserver/api_priority_and_fairness`

replacing `kube-apiserver-ip-10-0-145-102.us-east-2.compute.internal` for each server ?

collection-scripts/gather_priority_and_fairness
```diff
@@ -0,0 +1,28 @@
...
+
+mkdir -p ${APF_PATH}
+KUBE_API_SERVERS=$(oc get endpoints -n default kubernetes -o jsonpath='{.subsets[*].addresses [*].ip}')
+PORT=$(oc get endpoints -n default kubernetes -o jsonpath='{.subsets[*].ports [*].port}')
+
+for API_SERVER in ${KUBE_API_SERVERS}; do
+  mkdir -p ${APF_PATH}/${MASTER}
+
+  oc get --raw https://${API_SERVER}:${PORT}/debug/api_priority_and_fairness/dump_priority_levels \
+               > ${APF_PATH}/${MASTER}/priority_levels
```
p0lyn0mial, line 22: this doesn't look correct, where we set `MASTER` :) ?

collection-scripts/gather_priority_and_fairness
```diff
@@ -0,0 +1,29 @@
+#!/bin/bash -x
```
p0lyn0mial, line 1: `-x` was for debugging, right? Do we need it?

we usually set - for safety 
```
set -o nounset; set -o errexit; set -o pipefail
```

collection-scripts/gather_priority_and_fairness
```diff
@@ -0,0 +1,29 @@
...
+
+# If the cluster version is < 4.5 skip this test
+VERSION=$(oc get clusterversion version -o jsonpath='{.status.desired.version}')
+COMPARE=$(echo -e "4.5\n${VERSION}" | sort -V | head -n1)
+if [[ ${COMPARE} != "4.5" ]]; then
+  exit 0
+fi
+
+KUBE_API_SERVERS=$(oc get po -n openshift-kube-apiserver -l apiserver -o jsonpath='{range .items[*]}{.metadata.name}{" "}{.status.podIP}{" "}{range .spec.containers[?(.name=="kube-apiserver")]}{.ports[0].hostPort}{"\n"}')
```
p0lyn0mial, line 10: nit: could you add a comment that would show what we would get from running this command?

collection-scripts/gather_priority_and_fairness
zerodayz: decided to keep it since older versions of cluster can still run newer must-gather

sferich888 (changes requested):

collection-scripts/gather_priority_and_fairness
```diff
@@ -0,0 +1,37 @@
...
+# Fetches the podName podIP and podPort where container name is
+# kube-apiserver.
+# Example:
+# oc get po -n openshift-kube-apiserver -l apiserver -o jsonpath='{range .items[*]}{.metadata.name}{" "}{.status.podIP}{" "}{range .spec.containers[?(.name=="kube-apiserver")]}{.ports[0].hostPort}{"\n"}'
+# kube-apiserver-master-0.example.com 192.168.9.43 6443
+# kube-apiserver-master-1.example.com 192.168.13.112 6443
+# kube-apiserver-master-2.example.com 192.168.10.112 6443
+
+KUBE_API_SERVERS=$(oc get po -n openshift-kube-apiserver -l apiserver -o jsonpath='{range .items[*]}{.metadata.name}{" "}{.status.podIP}{" "}{range .spec.containers[?(.name=="kube-apiserver")]}{.ports[0].hostPort}{"\n"}')
+PORT=$(oc get endpoints -n default kubernetes -o jsonpath='{.subsets[*].ports [*].port}')
```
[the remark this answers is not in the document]
zerodayz: Hi, the KUBE_API_SERVER is used for the for while below. PORT was leftover and I have removed that. Thanks!

collection-scripts/gather_priority_and_fairness
```diff
@@ -0,0 +1,37 @@
...
+
+BASE_COLLECTION_PATH="./must-gather"
+NAMESPACE_PATH="namespaces/openshift-kube-apiserver"
+POD_PATH="kube-apiserver/kube-apiserver/api_priority_and_fairness"
+APF_PATH="${BASE_COLLECTION_PATH}/${NAMESPACE_PATH}"
```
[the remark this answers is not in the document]
zerodayz: Moved to the top of the script. Basically my intention was not to allocate any variables if the script is skipped due to version.

collection-scripts/gather_priority_and_fairness
```diff
@@ -0,0 +1,37 @@
+#!/bin/bash
+
+# If the cluster version is < 4.5 skip this test
+VERSION=$(oc get clusterversion version -o jsonpath='{.status.desired.version}')
+COMPARE=$(echo -e "4.5\n${VERSION}" | sort -V | head -n1)
+if [[ ${COMPARE} != "4.5" ]]; then
+  exit 0
+fi
```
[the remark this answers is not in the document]
zerodayz: @sferich888 I understand, the '!=' is not clear enough. 

The reason for != is simply because that's how it works when we use sort -V. It creates sorted list of versions.

I didn't find bash to be good at comparing decimal numbers. We could use bc? Whats your suggestions here to use instead of sort -V?

Would help if I add a comment to the script with the way how this works?

We set the current version to the VERSION variable:

VERSION=4.7
COMPARE=$(echo -e "4.5\n${VERSION}" | sort -V | head -n1)

echo $COMPARE
4.5

if version is >=4.5 then the COMPARE will always == 4.5

Or example if the version is lower:

VERSION=4.4
COMPARE=$(echo -e "4.5\n${VERSION}" | sort -V | head -n1)
echo $COMPARE
4.4

Then we skip the run.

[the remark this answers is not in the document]
sferich888: Commenting the code is always a good idea (it helps with maintenance). 
This is simply an interesting (and non-intuitive) way to do the comparison (and overcome the inadequacies with bash).

collection-scripts/gather_priority_and_fairness
```diff
@@ -0,0 +1,37 @@
+#!/bin/bash
+
+# If the cluster version is < 4.5 skip this test
+VERSION=$(oc get clusterversion version -o jsonpath='{.status.desired.version}')
+COMPARE=$(echo -e "4.5\n${VERSION}" | sort -V | head -n1)
+if [[ ${COMPARE} != "4.5" ]]; then
```
p0lyn0mial, line 6: I tried the following endpoints on a 4.5 cluster today and they were not available on that version.
I will check if we have them in 4.6 and report back here.

/hold

collection-scripts/gather_priority_and_fairness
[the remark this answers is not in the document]
zerodayz: @sferich888 I agree with you. Thank you. Added comments to explain the code bit better.

collection-scripts/gather_priority_and_fairness
p0lyn0mial: The debug endpoints were added in 1.19 (upstream). OpenShift 4.6 is based on that version. 

The endpoints are defined here https://github.com/openshift/kubernetes/blob/release-4.6/staging/src/k8s.io/apiserver/pkg/util/flowcontrol/apf_controller_debug.go#L41

It looks like we need to update the check to skip if the version is < 4.6

zerodayz: Updated to >=4.6

p0lyn0mial:
@zerodayz I have just downloaded the [latest must-gather](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/openshift_must-gather/204/pull-ci-openshift-must-gather-master-e2e-aws/1359638413019451392/artifacts/e2e-aws/gather-must-gather/artifacts/). It looks like there is one redundant directory at  `openshift-kube-apiserver/pods/kube-apiserver/kube-apiserver/api_priority_and_fairness`

collection-scripts/gather_priority_and_fairness
p0lyn0mial: thanks.

/hold cancel

p0lyn0mial:
I see you have pushed a few commits. 

I have checked https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_must-gather/204/pull-ci-openshift-must-gather-master-e2e-aws/1360102429114241024 which was build today (12.02) and it looks like the redundant directory is still present.

zerodayz:
> I see you have pushed a few commits. 
> 
> 
> 
> I have checked https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_must-gather/204/pull-ci-openshift-must-gather-master-e2e-aws/1360102429114241024 which was build today (12.02) and it looks like the redundant directory is still present.

Yeah I removed extra \n from print where I thought the new line is comings from, it seems the  redundant directory is created because the  extra empty line at the end. It doesn't reproduce on my machine so it feels like something specific to the testing env. Still trying to figure out. Sorry for the time it takes.

p0lyn0mial:
> Yeah I removed extra \n from print where I thought the new line is comings from, it seems the redundant directory is created because the extra empty line at the end. It doesn't reproduce on my machine so it feels like something specific to the testing env. Still trying to figure out. Sorry for the time it takes.

sure, np, take your time. Many thanks for pushing it forward!

zerodayz:
So indeed it seems that problem is new line after each host:
```
KUBE_API_SERVERS=$(oc get po -n openshift-kube-apiserver -l apiserver -o jsonpath='{range .items[*]}{.metadata.name}{" "}{.status.podIP}{" "}{range .spec.containers[?(.name=="kube-apiserver")]}{.ports[0].hostPort}{"\n"}')
```
It adds new line at the end:
```
echo "=======" > ${DEBUG_DIR}/debug_kas
echo "${KUBE_API_SERVERS}" >> ${DEBUG_DIR}/debug_kas
echo "=======" >> ${DEBUG_DIR}/debug_kas
```
```
=======
kube-apiserver-ip-10-0-131-143.us-west-2.compute.internal 10.0.131.143 6443

kube-apiserver-ip-10-0-146-112.us-west-2.compute.internal 10.0.146.112 6443

kube-apiserver-ip-10-0-223-158.us-west-2.compute.internal 10.0.223.158 6443

  
=======
```
And then we can see

```
=======
==== Entering the loop ===
==== Processing ===
Name:kube-apiserver-ip-10-0-131-143.us-west-2.compute.internal IP:10.0.131.143 Port:6443
=======
==== Entering the loop ===
==== Processing ===
Name: IP: Port:
=======
==== Entering the loop ===
==== Processing ===
Name:kube-apiserver-ip-10-0-146-112.us-west-2.compute.internal IP:10.0.146.112 Port:6443
=======
==== Entering the loop ===
==== Processing ===
Name: IP: Port:
=======
==== Entering the loop ===
==== Processing ===
Name:kube-apiserver-ip-10-0-223-158.us-west-2.compute.internal IP:10.0.223.158 Port:6443
=======
==== Entering the loop ===
==== Processing ===
Name: IP: Port:
=======
```

I have tried removing the new line but what I ended up with was:
```
=======
kube-apiserver-ip-10-0-131-143.us-west-2.compute.internal 10.0.131.143 6443kube-apiserver-ip-10-0-146-112.us-west-2.compute.internal 10.0.146.112 6443kube-apiserver-ip-10-0-223-158.us-west-2.compute.internal 10.0.223.158 6443
=======
```

I think maybe inserting ';' and post-processing with `tr` would do the job. 

Doing it now.

zerodayz:
I think I am hitting this bug in the test environment with oc
https://github.com/kubernetes/kubernetes/pull/98057

It's messing up the output in the script.

zerodayz:
```'
KUBE_API_SERVERS=$(oc get po -n openshift-kube-apiserver -l apiserver -o jsonpath='{range .items[*]}{.metadata.name}{","}{.status.podIP}{","}{range .spec.containers[?(.name=="kube-apiserver")]}{.ports[0].hostPort}{"\n"}')

 echo ${KUBE_API_SERVERS} > ${BASE_COLLECTION_PATH}/debug-KUBE_API_SERVERS
```

outputs

```
kube-apiserver-ip-10-0-144-236.us-west-2.compute.internal,10.0.144.236,6443 kube-apiserver-ip-10-0-182-101.us-west-2.compute.internal,10.0.182.101,6443 kube-apiserver-ip-10-0-214-34.us-west-2.compute.internal,10.0.214.34,6443 ,,
```

extra ,, meaning the extra line, hence kubernetes/kubernetes#98057

zerodayz:
@p0lyn0mial What about now ? https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/openshift_must-gather/204/pull-ci-openshift-must-gather-master-e2e-aws/1364060998268358656/artifacts/e2e-aws/gather-must-gather/artifacts/ Seems good now?

collection-scripts/gather_priority_and_fairness
```diff
@@ -0,0 +1,60 @@
...
+if [[ ${COMPARE} != "4.6" ]]; then
+  exit 0
+fi
+
+# Fetches the POD_NAME where app label is apiserver.
+# Example:
+#  oc get po -n openshift-kube-apiserver -l apiserver -o jsonpath='{range .items[*]}{.metadata.name}{"\n"}'
+#  kube-apiserver-master-0.example.com 192.168.9.43 6443
+#  kube-apiserver-master-1.example.com 192.168.13.112 6443
+#  kube-apiserver-master-2.example.com 192.168.10.112 6443
```
p0lyn0mial, line 33: nit: the command below will return only `kube-apiserver-master-0.example.com` so we could update the comment.

p0lyn0mial:
> @p0lyn0mial What about now ? https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/openshift_must-gather/204/pull-ci-openshift-must-gather-master-e2e-aws/1364060998268358656/artifacts/e2e-aws/gather-must-gather/artifacts/ Seems good now?

yes, it looks good to me. thanks!

collection-scripts/gather_priority_and_fairness
```diff
@@ -0,0 +1,60 @@
...
+  POD_PORT=$(oc get po -n openshift-kube-apiserver "${API_SERVER}" -o jsonpath='{range .spec.containers[?(.name=="kube-apiserver")]}{.ports[0].hostPort}')
+
+  oc get --raw https://"${POD_IP}":"${POD_PORT}"/debug/api_priority_and_fairness/dump_priority_levels \
+               > "${OUTPUT_DIR}"/priority_levels
+  oc get --raw https://"${POD_IP}":"${POD_PORT}"/debug/api_priority_and_fairness/dump_queues \
+               > "${OUTPUT_DIR}"/queues
+  oc get --raw https://"${POD_IP}":"${POD_PORT}"/debug/api_priority_and_fairness/dump_requests \
+               > "${OUTPUT_DIR}"/requests
+
+done
```
tkashem, line 60: line break here?

collection-scripts/gather_priority_and_fairness
```diff
@@ -0,0 +1,60 @@
...
+  #  192.168.9.43
+  POD_IP=$(oc get po -n openshift-kube-apiserver "${API_SERVER}" -o jsonpath='{.status.podIP}')
+
+  # Fetches the POD_PORT where container name is kube-apiserver.
+  # Example:
+  #  oc get po -n openshift-kube-apiserver kube-apiserver-master-0.example.com -o jsonpath='{range .spec.containers[?(.name=="kube-apiserver")]}{.ports[0].hostPort}'
+  #  6443
+  POD_PORT=$(oc get po -n openshift-kube-apiserver "${API_SERVER}" -o jsonpath='{range .spec.containers[?(.name=="kube-apiserver")]}{.ports[0].hostPort}')
+
+  oc get --raw https://"${POD_IP}":"${POD_PORT}"/debug/api_priority_and_fairness/dump_priority_levels \
```
tkashem, line 53: what happens if the request fails, does it get retried by must-gather?

in my opinion, `must-gather` is usually run when the cluster is under some sort of duress. We have three requests here, if one fails will the entire must-gather process abort?

tkashem (commented):
looks good to me, i have a question about retry.

collection-scripts/gather_priority_and_fairness
zerodayz: Correct. Fixed.

collection-scripts/gather_priority_and_fairness
zerodayz: Fixed.

collection-scripts/gather_priority_and_fairness
zerodayz: No retries have been implemented. I will take a look at it today. Thanks

tkashem: if we want to do retry, it can be done in a follow up PR.

p0lyn0mial:
/assign @sferich888

for approval

collection-scripts/gather_priority_and_fairness
sferich888: Is this really where we want to do this? 
Should this go in: https://github.com/openshift/oc/blob/master/pkg/cli/admin/inspect/pod.go#L161-L180 which already gets called when we collect the clusteroperators (related resources) with https://github.com/openshift/must-gather/blob/master/collection-scripts/gather#L19 and https://github.com/openshift/must-gather/blob/master/collection-scripts/gather#L43

p0lyn0mial: I don't think we should retry. In fact, a request will be already retried by `oc`.
We should make sure that the `must-gather` won't fail if we can not get some data. AFAIK it already works like this.

@tkashem thoughts?

p0lyn0mial: > Should this go in: https://github.com/openshift/oc/blob/master/pkg/cli/admin/inspect/pod.go#L161-L180 which

oh, I wasn't aware of it. It looks really good. Though I have never seen debug data attached to a must-gather.

p0lyn0mial: ah, so it gets collected on `oc adm inspect` not `oc adm must-gather`. It seems that `inspect` is for collecting debugging data

zerodayz: @sferich888 

I could use that debugEndpoints instead and it's pretty much better place to put it into. But how does the `oc adm inspect` work without token?

All my endpoint collections using `oc adm inspect` were skipped because Token is not empty.
https://github.com/openshift/oc/blob/master/pkg/cli/admin/inspect/pod.go#L58-L62

So it never got to:
https://github.com/openshift/oc/blob/master/pkg/cli/admin/inspect/pod.go#L72-L74

Which calls debug Endpoints collection:
https://github.com/openshift/oc/blob/f749845d71de73620f5e66553e0dd5abeb7efede/pkg/cli/admin/inspect/pod.go#L112-L113

That calls: https://github.com/openshift/oc/blob/f749845d71de73620f5e66553e0dd5abeb7efede/pkg/cli/admin/inspect/pod.go#L168-L177

zerodayz: So to me it seems that first we exit out of `gatherContainerInfo` before collecting debug endpoints because we have a token.

```
	if len(o.RESTConfig.BearerToken) > 0 {
		// token authentication is vulnerable to replays if the token is sent to a potentially untrustworthy source.
		klog.V(1).Infof("        Skipping container endpoint collection for pod %q container %q: Using token authentication\n", pod.Name, container.Name)
		return nil
	}
```

And then in `gatherContainerDebug` which we never get into, actually expect the token?

```
	for _, endpoint := range endpoints {
		// we need a token in order to access the /debug endpoint
		result, err := o.podUrlGetter.Get("/debug/pprof/"+endpoint, pod, o.restConfig, debugPort)
		if err != nil {
			return err
		}
		if err := o.fileWriter.WriteFromSource(path.Join(destDir, endpoint), &TextWriterSource{Text: result}); err != nil {
			return err
		}
	}
```

p0lyn0mial: I have just run `oc adm inspect clusteroperators` on my local cluster and here is what I have found for `kas`

```
ls openshift-kube-apiserver/pods/kube-apiserver-ip-10-0-149-59.ec2.internal/kube-apiserver/kube-apiserver
healthz      heap         logs         metrics.json profile      trace        version.json
```

sferich888:
Ok I will approve this; but can someone update the code with a note about why this isn't in `oc adm inspect` and point to https://github.com/openshift/oc/blob/master/pkg/cli/admin/inspect/pod.go#L58-L62. In short we don't want to collect this with `inspect` because the tokens could be used in a replay attack, thus the route we are taking here is forced by this.

zerodayz:
> Ok I will approve this; but can someone update the code with a note about why this isn't in `oc adm inspect` and point to https://github.com/openshift/oc/blob/master/pkg/cli/admin/inspect/pod.go#L58-L62. In short we don't want to collect this with `inspect` because the tokens could be used in a replay attack, thus the route we are taking here is forced by this.

@sferich888 What about now? The comment https://github.com/openshift/must-gather/blob/12c07d276742e503e476bd206fcb3268402b9dee/collection-scripts/gather_priority_and_fairness#L2-L5

collection-scripts/gather_priority_and_fairness
```diff
@@ -0,0 +1,64 @@
...
```
sferich888, line 57: We should capture errors with this (other scripts have samples on how to do this) and we should record them so we know why the collection fails.

sferich888 (changes requested):

sferich888:
> @sferich888 What about now? The comment https://github.com/openshift/must-gather/blob/12c07d276742e503e476bd206fcb3268402b9dee/collection-scripts/gather_priority_and_fairness#L2-L5

Looks good, found another nit, but once we consider that's I'll lgtm

collection-scripts/gather_priority_and_fairness
zerodayz: Updated. What do you think now?

collection-scripts/gather_priority_and_fairness
```diff
@@ -0,0 +1,85 @@
...
+  exit 0
+fi
+
+# Fetches the POD_NAME where app label is apiserver.
+# Example:
+#  oc get po -n openshift-kube-apiserver -l apiserver -o jsonpath='{range .items[*]}{.metadata.name}{"\n"}'
+#  kube-apiserver-master-0.example.com
+#  kube-apiserver-master-1.example.com
+#  kube-apiserver-master-2.example.com
+KUBE_API_SERVERS=$(oc get po -n openshift-kube-apiserver -l apiserver -o jsonpath='{range .items[*]}{.metadata.name}{"\n"}')
```
atiratree, line 45: @zerodayz
the ranges are not using a correct syntax and are missing an `{end}` element.

Please see:
- https://kubernetes.io/docs/reference/kubectl/jsonpath/
- https://bugzilla.redhat.com/show_bug.cgi?id=1900755#c16

PR #205, opened 2021-01-22
collect alerts from the cluster

deads2k:
Add a script to collect alerts when running must-gather. 

/assign @openshift/openshift-team-monitoring

### DIFF
--- a/collection-scripts/gather
+++ b/collection-scripts/gather
@@ -45,6 +45,9 @@ oc adm inspect --dest-dir must-gather "${named_resources[@]}"
 group_resources_text=$(IFS=, ; echo "${group_resources[*]}")
 oc adm inspect --dest-dir must-gather "${group_resources_text}"
 
+# Gather monitoring data from the cluster
+/usr/bin/gather_monitoring
+
 # Gather optional operator resources from all namespaces
 /usr/bin/gather_olm
 
@@ -52,6 +55,7 @@ oc adm inspect --dest-dir must-gather "${group_resources_text}"
 /usr/bin/gather_etcd
 
 # Gather Validation and Mutating Admission Webhook Objects
+# TODO the kube-apiserver operator should list these as related resources
 /usr/bin/gather_admission_webhooks
 
 # Gather Service Logs (using a supplemental Script); Scoped to Masters.
--- a/collection-scripts/gather_monitoring
+++ b/collection-scripts/gather_monitoring
@@ -1,0 +1,29 @@
+#!/bin/bash
+
+#Safeguards
+set -o nounset
+set -o errexit
+set -o pipefail
+
+BASE_COLLECTION_PATH="../must-gather"
+MONITORING_PATH="${BASE_COLLECTION_PATH}/monitoring"
+
+mkdir -p "${MONITORING_PATH}"
+
+MONITORING_ROUTE="$(oc get routes -n openshift-monitoring prometheus-k8s -o jsonpath={.status.ingress[0].host})"
+SA_TOKEN="$(oc sa get-token default)"
+
+# this is a CA bundle we need to verify the monitoring route, we will write it to disk so we can use it in the flag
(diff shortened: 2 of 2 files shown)


collection-scripts/gather
```diff
@@ -45,13 +45,17 @@ oc adm inspect --dest-dir must-gather "${named_resources[@]}"
...
+/usr/bin/gather_monitoring
+
 # Gather optional operator resources from all namespaces
 /usr/bin/gather_olm
 
 # Gather etcd information
 /usr/bin/gather_etcd
 
 # Gather Validation and Mutating Admission Webhook Objects
+# TODO the kube-apiserver operator should list these as related resources
```
sferich888, line 58: If this is the case, should we merge this? We have pushed other teams in the past to do this over adding resource collectors in this way.

collection-scripts/gather_monitoring
```diff
@@ -0,0 +1,27 @@
...
+
+BASE_COLLECTION_PATH="../must-gather"
+MONITORING_PATH="${BASE_COLLECTION_PATH}/monitoring"
+
+mkdir -p "${MONITORING_PATH}"
+
+MONITORING_ROUTE="$(oc get routes -n openshift-monitoring prometheus-k8s -o jsonpath={.status.ingress[0].host})"
+SA_TOKEN="$(oc sa get-token default)"
+
+# this is a CA bundle we need to verify the monitoring route, we will write it to disk so we can use it in the flag
```
sferich888, line 16: Should we clean this up (remove it from the archive) after we use it?

sferich888: I ask because its not valuable to debugging and its extra data we have to transfer off the cluster.

collection-scripts/gather_monitoring
```diff
@@ -0,0 +1,27 @@
...
+oc -n openshift-config-managed get cm default-ingress-cert -o jsonpath='{.data.ca-bundle\.crt}' > "${MONITORING_PATH}/ca-bundle.crt"
+
+# using oc get --raw because we directly control it and have standardized debugging on it
+oc get\
+  --server="https://${MONITORING_ROUTE}"\
+  --token="${SA_TOKEN}"\
+  --certificate-authority="${MONITORING_PATH}/ca-bundle.crt"\
+  --raw=/api/v1/rules?type=alert 2>/dev/null > "${MONITORING_PATH}/alerts.json"
```
sferich888, line 24: Do we want to send errors to `/dev/null` or would capturing them in a file (for latter debugging be valuable)?

sferich888: Or just letting them print to standard error (we capture the collector pod logs) so we would see these in that, and pull them back with that capture.

collection-scripts/gather_monitoring
deads2k: > Should we clean this up (remove it from the archive) after we use it?

sure, I'll try

collection-scripts/gather_monitoring
deads2k: > Do we want to send errors to `/dev/null` or would capturing them in a file (for latter debugging be valuable)?

they need to be separate so that we can always parse the json.  Do you  have a filename preference for preserving them?

collection-scripts/gather
deads2k: > If this is the case, should we merge this? We have pushed other teams in the past to do this over adding resource collectors in this way.

this PR doesn't make it worse.  I happened to catch this on the way though as I integrated and wanted to suggest a vision.

deads2k:
Updated for comments.

collection-scripts/gather
sferich888: +1 who can look at that as a bug/rfe (in 4.8/4.9)?

sferich888:
@deads2k what do we need to do about: https://github.com/openshift/must-gather/pull/205#issuecomment-776774242 (I think we are going to want this in 4.6).

PR #206, opened 2021-01-28
remove duplicate collection of admissionwebhooks

deads2k:
this appears to already be collected in https://github.com/openshift/cluster-kube-apiserver-operator/blob/master/pkg/operator/starter.go#L217-L218 and https://github.com/openshift/cluster-kube-apiserver-operator/blob/master/manifests/0000_20_kube-apiserver-operator_07_clusteroperator.yaml#L40-L43

/hold 

holding so we can look at artifacts and make sure we get the data we expect.


deads2k:
/hold cancel

I think I see them

PR #207, opened 2021-02-02
Gather ethtool info from driver and offload

rcarrillocruz:


rcarrillocruz:
@davemulford can you please review ? Thanks

PR #208, opened 2021-02-04
Collect configuration and events for vsphere clusters

rvanderp3:
vSphere events and configuration which are helpful to understanding the state of the environment housing the cluster are stored in prometheus and fed from the vsphere-problem-detector.  This pull-request adds a script which, when run on a vSphere cluster, retrieves and stores the gathered information.

### DIFF
--- a/collection-scripts/gather_vsphere_config
+++ b/collection-scripts/gather_vsphere_config
@@ -1,0 +1,16 @@
+#!/bin/bash
+BASE_COLLECTION_PATH="must-gather"
+VSPHERE_CONFIG_EVENTS_PATH=$BASE_COLLECTION_PATH/vsphere_config_events
+
+if [ -z "$(oc get deployment -n openshift-cluster-storage-operator vsphere-problem-detector-operator | grep vsphere-problem-detector-operator)" ]; then
+exit
+fi
+
+METRICS="vsphere_vcenter_info vsphere_features vsphere_cluster_check_total vsphere_cluster_check_errors vsphere_node_check_errors vsphere_node_check_total vsphere_node_hw_version_total"
+mkdir -p $VSPHERE_CONFIG_EVENTS_PATH
+for METRIC in $METRICS; do
+/usr/bin/oc exec -c prometheus -n openshift-monitoring prometheus-k8s-0 -- curl -s --data-urlencode "query=$METRIC" http://localhost:9090/api/v1/query > $VSPHERE_CONFIG_EVENTS_PATH/$METRIC.json
+done
+
+# force disk flush to ensure that all data gathered is accessible in the copy container
+sync
--- a/collection-scripts/gather
+++ b/collection-scripts/gather
@@ -63,5 +63,8 @@ oc adm inspect --dest-dir must-gather "${group_resources_text}"
 # Gather HAProxy config files
 /usr/bin/gather_haproxy_config
 
+# Gather vSphere Config and Events
+/usr/bin/gather_vsphere_config
+
 # force disk flush to ensure that all data gathered is accessible in the copy container
 sync


collection-scripts/gather_vsphere_config
```diff
@@ -0,0 +1,16 @@
...
+VSPHERE_CONFIG_EVENTS_PATH=$BASE_COLLECTION_PATH/vsphere_config_events
+
+if [ -z "$(oc get deployment -n openshift-cluster-storage-operator vsphere-problem-detector-operator | grep vsphere-problem-detector-operator)" ]; then
+exit
+fi
+
+METRICS="vsphere_vcenter_info vsphere_features vsphere_cluster_check_total vsphere_cluster_check_errors vsphere_node_check_errors vsphere_node_check_total vsphere_node_hw_version_total"
+mkdir -p $VSPHERE_CONFIG_EVENTS_PATH
+for METRIC in $METRICS; do
+/usr/bin/oc exec -c prometheus -n openshift-monitoring prometheus-k8s-0 -- curl -s --data-urlencode "query=$METRIC" http://localhost:9090/api/v1/query > $VSPHERE_CONFIG_EVENTS_PATH/$METRIC.json
```
sferich888, line 12: If we use oc inspect we already collect metrics for the operator. 

https://github.com/openshift/oc/blob/master/pkg/cli/admin/inspect/pod.go#L109

Is this a duplicate collection on the OpenShift Storage Operator?

collection-scripts/gather_vsphere_config
```diff
@@ -0,0 +1,16 @@
+#!/bin/bash
+BASE_COLLECTION_PATH="must-gather"
+VSPHERE_CONFIG_EVENTS_PATH=$BASE_COLLECTION_PATH/vsphere_config_events
+
+if [ -z "$(oc get deployment -n openshift-cluster-storage-operator vsphere-problem-detector-operator | grep vsphere-problem-detector-operator)" ]; then
+exit
+fi
+
+METRICS="vsphere_vcenter_info vsphere_features vsphere_cluster_check_total vsphere_cluster_check_errors vsphere_node_check_errors vsphere_node_check_total vsphere_node_hw_version_total"
```
sferich888, line 9: What of these should we be collecting with insights/telemetry vs the must-gater archive?

sferich888 (changes requested):
I want to have @radekvokal and @mfojtik  review this as I feel like metrics collection may want to be in one script vs N (individual scripts).

collection-scripts/gather_vsphere_config
rvanderp3: I was not aware and that `oc adm inspect` collected metrics.  If it does(I couldn't find it in a must-gather I took yesterday but I could have missed it) this would make this PR unnecessary.

collection-scripts/gather_vsphere_config
rvanderp3: > I want to have @radekvokal and @mfojtik review this as I feel like metrics collection may want to be in one script vs N (individual scripts).

That sounds reasonable to me.  Definitely don't want to go down a custom path here if it's not necessary.

sferich888: Not sure if 'custom path' is the wording I would choose here, but there may be a different approach to take when it comes to 'metrics' collection. Its why I want our folks in CCX to take a look t this, as well as for @mfojtik (as we may want to have a 'generic' collect_metrics script (that simply lists all the metrics we want data on (think the `gather` script, a resource builder that then collects all of what we want in an efficient way).

collection-scripts/gather_vsphere_config
sferich888: @rvanderp3 it gathers metrics on 'pods' - https://github.com/openshift/oc/blob/6352e607f262df44964a63004f85d81366accbbc/pkg/cli/admin/inspect/pod.go#L183-L196 so .. depending on the 'pod' your asking information from (and if you expose a `/metrics` endpoint) you get metrics for that pod. 

In short, going to the central metrics store (as you are here) may be breaking isolation (just a bit), however that dependent on how the metrics end up where they end up. 

You might need to check the Prometheus pods (data) to see the metrics (if they are collected). They may not be because were not exposing a query.

collection-scripts/gather_vsphere_config
```diff
@@ -0,0 +1,16 @@
...
+mkdir -p $VSPHERE_CONFIG_EVENTS_PATH
+for METRIC in $METRICS; do
```
sferich888, line 11: Can we collect this in asynchronously ? vs in sequence the way we are in a loop here?

rvanderp3:
I did more digging on this today.  It looks like this data is already being aggregated for telemetry.  I'm going to close this PR in preference to that.

PR #209, opened 2021-02-05
Add non-baremetal on-prem platform namespaces

cybertron:
Initially only the baremetal (kni) namespace was included in the
must-gather contents. Unfortunately, different namespaces were used
for all of the on-prem platforms so this misses the non-baremetal
ones. This change adds adds the other on-prem platforms so we will
have the necessary logs from those platforms as well.

### DIFF
--- a/collection-scripts/gather
+++ b/collection-scripts/gather
@@ -25,7 +25,7 @@ group_resources+=(certificatesigningrequests)
 group_resources+=(nodes)
 
 # Namespaces/Project Resources
-named_resources+=(ns/default ns/openshift ns/kube-system ns/openshift-etcd ns/openshift-kni-infra)
+named_resources+=(ns/default ns/openshift ns/kube-system ns/openshift-etcd ns/openshift-kni-infra ns/openshift-openstack-infra ns/openshift-vsphere-infra ns/openshift-ovirt-infra)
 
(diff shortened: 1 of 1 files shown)


sferich888:
This is a lot of namespaces, should / could these be added / collected with relatedObjects (resources)?

collection-scripts/gather
```diff
@@ -25,7 +25,7 @@ group_resources+=(certificatesigningrequests)
 group_resources+=(nodes)
 
 # Namespaces/Project Resources
-named_resources+=(ns/default ns/openshift ns/kube-system ns/openshift-etcd ns/openshift-kni-infra)
+named_resources+=(ns/default ns/openshift ns/kube-system ns/openshift-etcd ns/openshift-kni-infra ns/openshift-openstack-infra ns/openshift-vsphere-infra ns/openshift-ovirt-infra)
```
mandre, line 28: Looks like we could remove all these `ns/openshift-*-infra` namespaces if https://github.com/openshift/machine-config-operator/pull/2460 makes it through.

cybertron:
/hold

Yeah, once the MCO patch merges I'll rework this to remove the kni namespace.

cybertron:
Superseded by #238

PR #214, opened 2021-02-25
feat/gather-monitoring: Support to collect Prometheus metrics

mtulio:
Easy  way to collect a group of metrics from Prometheus without requesting a entire dump.

The strategy is to read config from CM that can customize the collector script, like: time frame, metrics name/queries, metrics prefix to collect entire group and resolution.

The idea is to help the investigation and troubleshooting of the cluster looking the metrics for specific components without requesting GiB of data from Prometheus database.

There's a usage [samples here](https://github.com/mtulio/must-gather-monitoring/tree/main/must-gather#samples) (the "Grafana discovery feature" was removed from this PR - discussions bellow).

### DIFF
--- a/collection-scripts/gather
+++ b/collection-scripts/gather
@@ -70,5 +70,8 @@ oc adm inspect --dest-dir must-gather "${group_resources_text}"
 # Gather HAProxy config files
 /usr/bin/gather_haproxy_config
 
+# Gather Monitoring metrics
+/usr/bin/gather_monitoring
+
 # force disk flush to ensure that all data gathered is accessible in the copy container
 sync
--- a/collection-scripts/gather
+++ b/collection-scripts/gather
@@ -70,8 +70,5 @@ oc adm inspect --dest-dir must-gather "${group_resources_text}"
 # Gather HAProxy config files
 /usr/bin/gather_haproxy_config
 
-# Gather Monitoring metrics
-/usr/bin/gather_monitoring
-
 # force disk flush to ensure that all data gathered is accessible in the copy container
 sync
--- a/collection-scripts/gather_monitoring
+++ b/collection-scripts/gather_monitoring
@@ -4,8 +4,6 @@
 set -o nounset
 set -o errexit
 set -o pipefail
-set -o nounset
-set -o errexit
 
 declare -r BASE_COLLECTION_PATH="../must-gather"
 declare -r MONITORING_PATH="${BASE_COLLECTION_PATH}/monitoring"
@@ -67,7 +65,6 @@ alertmanager_get() {
       > "$result_path.json" \
       2> "$result_path.stderr"
 }
-
 
 monitoring_gather(){
   init
--- a/collection-scripts/gather_monitoring
+++ b/collection-scripts/gather_monitoring
@@ -4,6 +4,8 @@
 set -o nounset
 set -o errexit
 set -o pipefail
+set -o nounset
+set -o errexit
 
 declare -r BASE_COLLECTION_PATH="../must-gather"
 declare -r MONITORING_PATH="${BASE_COLLECTION_PATH}/monitoring"
--- a/collection-scripts/gather-monitoring-data
+++ b/collection-scripts/gather-monitoring-data
@@ -106,7 +106,7 @@ function get_query_range {
     echo_info "Collecting queries range..."
 
     for query in ${GATHER_PROM_QUERIES_RANGE}; do
-        prom_get_expression_query "query_range" "${query}"
+        prom_get_expression_query "query_range" "${query}" || true
     done
 }
 
@@ -125,7 +125,7 @@ function get_query {
     echo_info "Collecting instant queries..."
 
     for query in ${GATHER_PROM_QUERIES}; do
-        prom_get_expression_query "query" "${query}"
+        prom_get_expression_query "query" "${query}" || true
     done
 }
 
@@ -154,7 +154,7 @@ function get_query_range_discovery {
     while read metric; do
         for prefix in ${GATHER_PROM_QUERIES_RANGE_PREFIX}; do
             if [[ "${metric}" =~ ^${prefix}.* ]]; then
-                prom_get_expression_query "query_range" "${metric}"
+                prom_get_expression_query "query_range" "${metric}" || true
             fi
         done
     done < "${MG_PROM_PATH}/prometheus-metrics.txt"
@@ -213,7 +213,7 @@ function init {
 
     # Setting: Session token, overwriten by GATHER_MONIT_TOKEN
     declare -g SS_TOKEN="${GATHER_MONIT_TOKEN:-$(oc whoami -t)}"
-    declare -g PROM_HOST=$(oc -n "${MONIT_NS}" get route prometheus-k8s -o jsonpath='{.spec.host}{"\n"}')
+    declare -g PROM_HOST=$(oc -n "${MONIT_NS}" get route thanos-querier -o jsonpath='{.spec.host}{"\n"}')
     declare -g PROM_API_PATH="/api/v1"
 
     # Query Param: 'start' and 'end' timestamp
--- a/collection-scripts/gather-monitoring-data
+++ b/collection-scripts/gather-monitoring-data
@@ -199,6 +199,10 @@ function init {
     declare -g MONIT_NS="openshift-monitoring"
     declare -g MONIT_ENV="must-gather-env"
 
+    # prometheus-k8s | thanos-querier : TODO need to validate thanos-querier endpoints
+    local prom_route_name="${PROM_ROUTE_NAME:-"prometheus-k8s"}"
+    declare -g PROM_API_PATH="/api/v1"
+
     declare -g MG_BASE_PATH="${MG_BASE_PATH:-"/must-gather"}"
     declare -g MG_MONITORING_PATH="${MG_BASE_PATH}/monitoring"
     declare -g MG_PROM_PATH="${MG_MONITORING_PATH}/prometheus"
@@ -213,8 +217,7 @@ function init {
 
     # Setting: Session token, overwriten by GATHER_MONIT_TOKEN
     declare -g SS_TOKEN="${GATHER_MONIT_TOKEN:-$(oc whoami -t)}"
-    declare -g PROM_HOST=$(oc -n "${MONIT_NS}" get route thanos-querier -o jsonpath='{.spec.host}{"\n"}')
-    declare -g PROM_API_PATH="/api/v1"
+    declare -g PROM_HOST=$(oc -n "${MONIT_NS}" get route "${prom_route_name}" -o jsonpath='{.spec.host}{"\n"}')
 
     # Query Param: 'start' and 'end' timestamp
     local date_start_human=${GATHER_MONIT_START_DATE:-${GATHER_MONIT_START_DATE_DEFAULT}}
--- a/collection-scripts/gather-monitoring-data
+++ b/collection-scripts/gather-monitoring-data
@@ -1,26 +1,9 @@
 #!/bin/bash
 
-# Gather Monitoring Metrics and Dashboards.
-# - Discovery and save Grafana dashboards, extracting the metrics' datapoints from Prometheus.
-# - Gather custom metrics (when defined on Config variables).
-# - Discovery metrics from prefix name.
 #
-# Config variables:
-# - GATHER_MONIT_START_DATE: starting human readable date string (date -d). Default: "7 days ago".
-# - GATHER_MONIT_END_DATE: ending human readable date string (date -d). Default: "now".
-# - GATHER_MONIT_QUERY_STEP: metric resolution to get from Prometheus. Default: "1m".
-# - GATHER_PROM_QUERIES_RANGE: list of additional metrics name to be collected. Default: Undefined
-# - GATHER_PROM_QUERIES_RANGE_PREFIX: discovery metrics by prefixes. Default: ''
-# - GATHER_MONIT_TOKEN: optional service account token to gather Prometheus and Grafana API. Default: Undefined
+# Collect metrics data from Prometheus API.
+# See helper function (show_help|-h) for more information.
 #
-# To create the ConfigMap:
-# $ echo -e 'GATHER_MONIT_START_DATE="15 days ago"\nGATHER_MONIT_QUERY_STEP="5m"' > env
-# $ echo -e 'GATHER_PROM_QUERIES_RANGE="up"' > env
-# $ echo -e 'GATHER_PROM_QUERIES_RANGE_PREFIX="apiserver_"' > env
-# $ oc create configmap must-gather-env -n openshift-monitoring --from-file=env=env
-#
-# References:
-# - Prometheus API: https://prometheus.io/docs/prometheus/latest/querying/api/
 
 #Safeguards
 set -o pipefail
@@ -55,7 +38,7 @@ function echo_info {
 # ##############
 
 function oc_get {
-    oc get \
+    ${OC_CLI[@]} get \
         --certificate-authority="${CA_BUNDLE}" \
         --token="${SS_TOKEN}" \
         "$@"
@@ -136,10 +119,11 @@ function get_query {
 function get_query_range_discovery {
 
     if [[ -z "${GATHER_PROM_QUERIES_RANGE_PREFIX:-}" ]]; then
-        echo_info " No metrics was found to collect from env GATHER_PROM_QUERIES_RANGE_PREFIX"
+        echo_info "No metrics was found to collect from env GATHER_PROM_QUERIES_RANGE_PREFIX"
         return
     fi
     echo_info "Collecting metrics by prefixes..."
+    mkdir -p "${MG_METRICS_PATH}"
 
     # dump current available metrics to discovery/filter desired as defined on env var
     prom_get_api "label/__name__/values" \
@@ -171,53 +155,104 @@ function cleanup {
 # this is a CA bundle we need to verify the monitoring route,
 # we will write it to disk so we can use it in the flag
 function get_ca_bundle {
-    oc -n openshift-config-managed \
+    ${OC_CLI[@]} -n openshift-config-managed \
         get cm default-ingress-cert \
         -o jsonpath='{.data.ca-bundle\.crt}' > "${CA_BUNDLE}"
 }
 
-function get_env {
-    oc -n "${MONIT_NS}" extract "cm/${MONIT_ENV}" \
-        --keys=env \
-        --to="${MG_MONITORING_PATH}/" 2>&1 >/dev/null || true
+# Handle configuration from ConfigMap (less precedence than CLI)
+function set_env_from_config {
 
-    echo_info "Get custom metrics from ConfigMap ${MONIT_ENV} on project ${MONIT_NS}"
+    if [[ ${OPT_CONFIG} == false ]]; then return; fi
+
+    # parse cm from 'namespace/configMapName' to 'cm_namespace' and 'cm_name'
+    local cm_namespace=$(echo ${OPT_CONFIG_NAME} | awk -F'/' '{print$1}')
+    local cm_name=$(echo ${OPT_CONFIG_NAME} | awk -F'/' '{print$2}')
+
+    ${OC_CLI[@]} -n "${cm_namespace}" \
+        extract "cm/${cm_name}" \
+        --keys=env \
+        --confirm \
+        --to="${MG_MONITORING_PATH}/" 2>&1 >/dev/null
+
+    echo_info "Get custom metrics from ConfigMap ${cm_name} on project ${cm_namespace}"
     if [[ -s "${MG_MONITORING_PATH}/env" ]]; then
         echo_info "Loading custom environments variables from ${MG_MONITORING_PATH}/env"
-        source "${MG_MONITORING_PATH}/env"
+        eval $(cat "${MG_MONITORING_PATH}/env")
     else
         echo_info "Unable to load custom environments from ConfigMap, ignoring."
     fi
 }
 
-function init {
+# Handle configuration from CLI args (more precedence than CM)
+function set_env_from_cli {
 
-    declare -g GATHER_MONIT_START_DATE_DEFAULT="7 days ago"
+    if [[ ! -z "${OPT_GATHER_QUERY}" ]]; then
+        declare -g GATHER_PROM_QUERIES="${OPT_GATHER_QUERY}"
+    fi
+
+    if [[ ! -z "${OPT_GATHER_QUERY_RANGE}" ]]; then
+        declare -g GATHER_PROM_QUERIES_RANGE="${OPT_GATHER_QUERY_RANGE}"
+    fi
+
+    if [[ ! -z "${OPT_GATHER_QUERY_RANGE_PREFIX}" ]]; then
+        declare -g GATHER_PROM_QUERIES_RANGE_PREFIX="${OPT_GATHER_QUERY_RANGE_PREFIX}"
+    fi
+
+}
+
+function set_env_default {
+
+    declare -g OPT_VERBOSE=false
+    declare -g OPT_DEBUG=false
+    declare -g OPT_CONFIG=false
+    declare -g OPT_CONFIG_NAME=
+    declare -g OPT_GATHER_QUERY=
+    declare -g OPT_GATHER_QUERY_RANGE=
+    declare -g OPT_GATHER_QUERY_RANGE_PREFIX=
+
+    declare -g GATHER_BASE_PATH_DEFAULT="/must-gather"
+    declare -g GATHER_MONIT_START_DATE_DEFAULT="1 day ago"
     declare -g GATHER_MONIT_END_DATE_DEFAULT="now"
     declare -g GATHER_MONIT_QUERY_STEP_DEFAULT="1m"
 
-    declare -g MONIT_NS="openshift-monitoring"
-    declare -g MONIT_ENV="must-gather-env"
+    declare -g OC_REQUEST_TIMEOUT_DEFAULT="15s"
 
-    # prometheus-k8s | thanos-querier : TODO need to validate thanos-querier endpoints
-    local prom_route_name="${PROM_ROUTE_NAME:-"prometheus-k8s"}"
+}
+
+function init_env {
+
     declare -g PROM_API_PATH="/api/v1"
 
-    declare -g MG_BASE_PATH="${MG_BASE_PATH:-"/must-gather"}"
+    declare -g MG_BASE_PATH="${GATHER_BASE_PATH_DEFAULT:-"/must-gather"}"
     declare -g MG_MONITORING_PATH="${MG_BASE_PATH}/monitoring"
     declare -g MG_PROM_PATH="${MG_MONITORING_PATH}/prometheus"
     declare -g MG_METRICS_PATH="${MG_PROM_PATH}/metrics"
     declare -g CA_BUNDLE="${MG_MONITORING_PATH}/ca-bundle.crt"
 
+    declare -g OC_CLI=()
+    OC_CLI+=(`which oc`)
+    OC_CLI+=("--request-timeout=${OC_REQUEST_TIMEOUT_DEFAULT}")
+}
+
+function init {
+
+    init_env
     mkdir -p "${MG_MONITORING_PATH}"
 
-    get_env
+    # Set Configurations from dynamic sources (CM or CLI)
+    set_env_from_config
+    set_env_from_cli
 
+    # Cache default ingress CA
     get_ca_bundle
 
-    # Setting: Session token, overwriten by GATHER_MONIT_TOKEN
-    declare -g SS_TOKEN="${GATHER_MONIT_TOKEN:-$(oc whoami -t)}"
-    declare -g PROM_HOST=$(oc -n "${MONIT_NS}" get route "${prom_route_name}" -o jsonpath='{.spec.host}{"\n"}')
+    # Session token, overwriten by GATHER_MONIT_TOKEN
+    declare -g SS_TOKEN="${GATHER_MONIT_TOKEN:-$(${OC_CLI[@]} whoami -t)}"
+
+    # prometheus-k8s | thanos-querier : TODO need to validate thanos-querier endpoints is compatible
+    local prom_route_name="${PROM_ROUTE_NAME:-"prometheus-k8s"}"
+    declare -g PROM_HOST=$(${OC_CLI[@]} -n openshift-monitoring get route "${prom_route_name}" -o jsonpath='{.spec.host}{"\n"}')
 
     # Query Param: 'start' and 'end' timestamp
     local date_start_human=${GATHER_MONIT_START_DATE:-${GATHER_MONIT_START_DATE_DEFAULT}}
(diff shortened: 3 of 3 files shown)


mtulio:
Seem to had a recent implementation of the same script, I didn't see that at the begin of the implementation.

So, I am waiting to the feedback since the both does not do the same thing, are complementary.

https://github.com/openshift/must-gather/pull/211 get alerts and this one get the metrics. Wanting to feedback to rebase and follow to merge both ideas.

mtulio:
I am not sure if we need to leave the Dashboard discovery by default enabled, or "7 days ago" the default time range, since the mass of data generated (from prometheus metrics) will be quite large.

I got a non-prod cluster with few nodes and the total uncompressed data was arround 2.5GB.

hexfusion:
cc @retroflexer can you review please

retroflexer:
Pretty cool stuff. Thanks @mtulio for the initiative!

collection-scripts/gather_monitoring
```diff
@@ -1,29 +1,235 @@
...
+    echo "INFO: Extracting metrics from dashboards"
+    sort -u ${GF_PATH}/dashboard_*_metrics.txt > ${GF_PATH}/dashboards_metrics.txt
+    while read METRIC; do
+        echo "INFO: Getting metric range: $METRIC"
+        prom_query_range "$METRIC" "$METRIC"
+    done < ${GF_PATH}/dashboards_metrics.txt
+}
+
+function get_grafana_metrics() {
+    ENABLE_DISCOVERY=${GATHER_MONIT_DISCOVERY_GRAFANA:-"yes"}
```
retroflexer, line 188: nit: in place of the default value of explicit "yes", you should perhaps use `GATHER_MONIT_DISCOVERY_GRAFANA_DEFAULT`.

collection-scripts/gather_monitoring
```diff
@@ -1,29 +1,235 @@
...
+
+PROM_HOST=$(oc -n ${MONIT_NS} get route prometheus-k8s -o jsonpath='{.spec.host}{"\n"}')
+PROM_URL="https://${PROM_HOST}"
+
+GF_HOST=$(oc -n ${MONIT_NS} get route grafana  -o jsonpath='{.spec.host}{"\n"}')
+GF_URL="https://${GF_HOST}"
+
+OCP_TOKEN=$(oc whoami -t)
+
+JQ_PATH=$(which jq 2>/dev/null)
```
retroflexer, line 50: which is non-standard. Use builtin 'command -v' instead.

collection-scripts/gather_monitoring
```diff
@@ -1,29 +1,235 @@
...
+    export JQ_PATH=/usr/local/bin/jq
+    curl -s -o ${JQ_PATH} \
+      http://stedolan.github.io/jq/download/linux64/jq && \
+      chmod +x ${JQ_PATH}
+}
+
+# Make a query (single metric) to Prometheus API /api/v1/query_range
+function prom_query_range() {
+    QUERY=$1
+    METRIC_FILE=${2:-$(echo ${QUERY} |awk -F'\(' '{print$1}')}
```
retroflexer, line 93: Double quote "${QUERY}" to prevent globbing and word splitting.

collection-scripts/gather_monitoring
```diff
@@ -1,29 +1,235 @@
...
+
+    # The metrics are ziped to decrease the datatransfer
+    curl ${CURL_OPTIONS} \
+        -H "Authorization: Bearer ${OCP_TOKEN}" \
+        -H 'Accept-encoding: gzip' \
+        --data-urlencode "query=${QUERY}" \
+        --data-urlencode "start=${DATE_START}" \
+        --data-urlencode "end=${DATE_END}" \
+        --data-urlencode "step=${QUERY_STEP}" \
+        $PROM_URL/api/v1/query_range > ${PROM_PATH}/metric-${METRIC_FILE}.json.gz
```
retroflexer, line 103: Nit: Double quote $PROM_URL and ${METRIC_FILE}. Also add curly braces for PROM_URL for consistency.

collection-scripts/gather_monitoring
```diff
@@ -1,29 +1,235 @@
...
+# Discovery metrics from prefixes defined on environment GATHER_MONIT_DISCOVERY_PREFIXES
+function get_discovery_metrics() {
+
+    if [[ -z "${GATHER_MONIT_DISCOVERY_PREFIXES}" ]]; then
+        return
+    fi
+
+    echo "INFO: Starting discovery metrics by prefixes..."
+    # TODO: fix it shame~~ O(n*m) ;(
+    while read METRIC; do
```
retroflexer, line 129: Use `read -r`.  A read call without `-r` will mangle backslashes.

collection-scripts/gather_monitoring
```diff
@@ -1,29 +1,235 @@
...
+        for PREFIX in ${GATHER_MONIT_DISCOVERY_PREFIXES}; do
+            if [[ "${METRIC}" =~ ^${PREFIX}.* ]]; then
+                echo "INFO: Getting metric range: ${METRIC}"
+                prom_query_range "${METRIC}" "${METRIC}" ;
+            fi
+        done
+    done < <( curl -sk ${CURL_OPTS} \
```
retroflexer, line 136: Double quotes around CURL_OPTS.

collection-scripts/gather_monitoring
```diff
@@ -1,29 +1,235 @@
...
+# then collect it's metrics from Prometheus API.
+function gf_discovery_metrics_from_dashboards() {
+
+    install_jq
+    ${JQ_PATH} \
+        -r '.[] | .uri +":"+ .uid' ${GF_PATH}/dashboards.json \
+        | awk -F'db/' '{print$2}' \
+        | awk '{print$1":"$2}' > ${GF_PATH}/dashboards.txt
+
+    while read DASHBOARD; do
```
retroflexer, line 164: Use `read -r`

collection-scripts/gather_monitoring
```diff
@@ -1,29 +1,235 @@
...
+        NAME=$(echo $DASHBOARD |awk -F':' '{print$1}')
+        DUID=$(echo $DASHBOARD |awk -F':' '{print$2}')
```
retroflexer, line 166: Use double quotes for $DASHBOARD.

collection-scripts/gather_monitoring
```diff
@@ -1,29 +1,235 @@
...
+
+        echo "INFO: Getting Dashboard ${NAME}"
+        curl ${CURL_OPTIONS} \
+            -H "Authorization: Bearer ${OCP_TOKEN}" \
+            $GF_URL/api/dashboards/uid/${DUID} > ${GF_PATH}/dashboard_${NAME}.json
```
retroflexer, line 171: Double quote to prevent globbing and word splitting.

collection-scripts/gather_monitoring
```diff
@@ -1,29 +1,235 @@
...
+
+        ${JQ_PATH} \
+            .dashboard.rows[].panels[].targets[].expr ${GF_PATH}/dashboard_${NAME}.json \
```
retroflexer, line 174: Double quotes.

collection-scripts/gather_monitoring
mtulio: I changed to the curl-ing global var. 👍🏼

mtulio:
@retroflexer reviewed, thanks.   https://github.com/mtulio/must-gather/pull/8

mtulio:
@hexfusion @retroflexer loading the metrics to a remote storage is working as described [here](https://github.com/mtulio/must-gather-monitoring#usage)

mtulio:
Bug fix on parsing recording rules: https://github.com/mtulio/must-gather/pull/9

collection-scripts/gather
```diff
@@ -67,5 +67,8 @@ oc adm inspect --dest-dir must-gather "${group_resources_text}"
 # Gather HAProxy config files
 /usr/bin/gather_haproxy_config
 
+# Gather Monitoring metrics
+/usr/bin/gather_monitoring
```
dgrisonnet, line 71: gather_monitoring is already executed above, at line 49

dgrisonnet:
pretty nice addition @mtulio :+1:

dgrisonnet:
Just an after thought, I recall being told that the must-gather is quite size sensitive, so I wonder how it would be perceive to collect all the metrics for a certain time range as this represent quite a lot of data.

collection-scripts/gather
mtulio: @dgrisonnet good catch. Thanks. Removed!

mtulio:
@dgrisonnet totally agree with you. I believe that we need to be more specific when collecting some metrics, due this query_range will read all data points from disk and return it, so could generate disk IO, when we are working in clusters that share the storage, it seems to be a problem. At the same time, this [script could be customized](https://github.com/mtulio/must-gather-monitoring/tree/main/must-gather#usage) to collect specific metrics, that is pretty useful to avoid entire dump of Prometheus.

Here https://github.com/mtulio/must-gather/pull/16 I just disable by default the "Grafana discovery" , and I am doing some final validatoins. I suggest by default don't collect any metrics, and when desired we can set the variables and run the script.

@dgrisonnet @hexfusion @retroflexer 
Another topic that I would like to discuss is: not sure if downloading `jq` every execution is a good approach, at the same time I don't know if it's a good practice to embed by default the `jq` binary inside base Image (as I did here https://github.com/mtulio/must-gather/pull/17). What do you think, folks? The binary size, stripped, is ~3.5M

collection-scripts/gather_monitoring
```diff
@@ -1,29 +1,259 @@
...
+echo "INFO: Config Env GAHTER_MONIT_START_DATE: ${GAHTER_MONIT_START_DATE}"
+echo "INFO: Config Env GAHTER_MONIT_END_DATE: ${GAHTER_MONIT_END_DATE}"
+echo "INFO: Config Env GAHTER_MONIT_QUERY_STEP: ${GAHTER_MONIT_QUERY_STEP}"
+echo "INFO: Config Env GATHER_MONIT_CUSTOM_METRICS: ${GATHER_MONIT_CUSTOM_METRICS}"
+echo "INFO: Config Env GATHER_MONIT_DISCOVERY_PREFIXES: ${GATHER_MONIT_DISCOVERY_PREFIXES}"
+echo "INFO: Config Env GATHER_MONIT_DISCOVERY_GRAFANA: ${GATHER_MONIT_DISCOVERY_GRAFANA}"
+echo "INFO: Prometheus endpoint URL: ${PROM_URL}"
+echo "INFO: Grafana endpoint URL: ${GF_URL}"
+
+function install_jq() {
```
sthaha, line 100: This may fail if the cluster is disconnected / behind a proxy that blocks github.com etc ..

sthaha:
@mtulio I have a patch similar to this albeit  smaller in scope. https://github.com/openshift/must-gather/pull/234/files. It may be worth collaborating on this.

collection-scripts/gather_monitoring
```diff
@@ -1,29 +1,208 @@
...
+# - Prometheus API: https://prometheus.io/docs/prometheus/latest/querying/api/
+
 #Safeguards
-set -o nounset
-set -o errexit
 set -o pipefail
 
-BASE_COLLECTION_PATH="../must-gather"
-MONITORING_PATH="${BASE_COLLECTION_PATH}/monitoring"
+err_report() {
```
sthaha, line 28: nit: do you want to use the same convention `function name {` used else where ? 
My personal preference is name()  😜

collection-scripts/gather_monitoring
```diff
@@ -1,29 +1,208 @@
...
+    echo "ERROR: on line $1 from ${0}"
+}
+trap 'err_report $LINENO' ERR
```
sthaha, line 31: nice ❤️

collection-scripts/gather_monitoring
```diff
@@ -1,29 +1,208 @@
...
+# we will write it to disk so we can use it in the flag
+function get_ca_bundle {
+    oc -n openshift-config-managed \
+        get cm default-ingress-cert \
+        -o jsonpath='{.data.ca-bundle\.crt}' > "${CA_BUNDLE}"
+}
+
+function init {
+
+    test -d "${PROM_PATH}" || mkdir -p "${PROM_PATH}"
```
sthaha, line 77: mkdir -p "${PROM_PATH}" should suffice since `mkdir` does the testing already

collection-scripts/gather_monitoring
```diff
@@ -1,29 +1,208 @@
...
+declare -r CURL_CMD=(curl --fail -sk)
+
+declare -r BASE_COLLECTION_PATH="/must-gather"
+declare -r MONITORING_PATH="${BASE_COLLECTION_PATH}/monitoring"
+declare -r PROM_PATH="${MONITORING_PATH}/prometheus"
+declare -r CA_BUNDLE="${MONITORING_PATH}/ca-bundle.crt"
+
+# Setting: Session token, overwriten by GAHTER_MONIT_TOKEN
+declare -r SS_TOKEN="${GAHTER_MONIT_TOKEN:-$(oc whoami -t)}"
+declare -r SA_TOKEN="$(oc sa get-token default)"
```
sthaha, line 49: Is this used?

collection-scripts/gather_monitoring
```diff
@@ -1,29 +1,208 @@
...
+    echo "INFO: Prometheus endpoint: https://${PROM_HOST}"
+    echo "INFO: Prometheus base query: ${PROM_API}/<uri>?query=<metric>&${PROM_API_QSTR}"
+}
+
+function cleanup {
+  rm "${CA_BUNDLE}"
+}
+
+function oc_get {
+  oc get \
```
sthaha, line 110: nit: inconsistent spacing for indentation: 2/ 4 spaces

collection-scripts/gather_monitoring
```diff
@@ -1,29 +1,208 @@
...
+trap 'err_report $LINENO' ERR
+
+declare -r GAHTER_MONIT_START_DATE_DEFAULT="7 days ago"
+declare -r GAHTER_MONIT_END_DATE_DEFAULT="now"
+declare -r GAHTER_MONIT_QUERY_STEP_DEFAULT="1m"
+
+declare -r MONIT_NS="openshift-monitoring"
+declare -r MONIT_ENV="must-gather-env"
+
+declare -r CURL_CMD=(curl --fail -sk)
```
sthaha, line 40: is this used?

collection-scripts/gather_monitoring
```diff
@@ -1,29 +1,208 @@
...
+    declare -g DATE_START=$(date -d "${_START_HUNAN}" +%s)
+    declare -r _END_HUMAN=${GAHTER_MONIT_END_DATE:-${GAHTER_MONIT_END_DATE_DEFAULT}}
+    declare -g DATE_END=$(date -d "${_END_HUMAN}" +%s)
+
+    # Setting: Metric resolution. Low resolution in longer range could be limited to 11k datapoints
+    declare -g QUERY_STEP=${GAHTER_MONIT_QUERY_STEP:-${GAHTER_MONIT_QUERY_STEP_DEFAULT}}
+
+    declare -g PROM_API_QSTR="start=${DATE_START}&end=${DATE_END}&step=${QUERY_STEP}"
+
+    echo "INFO: Metrics config time range from=${DATE_START} to=${DATE_END} step=${QUERY_STEP}"
```
sthaha, line 94:
```
cat <<-"EOF"
  INFO: Metrics config time range from=${DATE_START} to ${DATE_END}

EOF
```
may be easier on eyes

collection-scripts/gather_monitoring
```diff
@@ -1,29 +1,208 @@
...
+function init {
+
+    test -d "${PROM_PATH}" || mkdir -p "${PROM_PATH}"
+
+    get_ca_bundle
+
+    get_env
+
+    # Setting: Timestamp to Prometheus' query range
+    declare -r _START_HUNAN=${GAHTER_MONIT_START_DATE:-${GAHTER_MONIT_START_DATE_DEFAULT}}
```
sthaha, line 84: did you mean HUMAN? 

Also if the variable isn't used outside the function, how about making it `local ` ?

collection-scripts/gather_monitoring
```diff
@@ -1,29 +1,208 @@
...
+# Config variables:
+# - GAHTER_MONIT_START_DATE: starting human readable date string (date -d). Default: "7 days ago".
+# - GAHTER_MONIT_END_DATE: ending human readable date string (date -d). Default: "now".
+# - GAHTER_MONIT_QUERY_STEP: metric resolution to get from Prometheus. Default: "1m".
+# - GATHER_MONIT_CUSTOM_METRICS: list of additional metrics name to be collected. Default: Undefined
+# - GATHER_MONIT_DISCOVERY_PREFIXES: discovery metrics by prefixes. Default: ''
+# - GAHTER_MONIT_TOKEN: optional service account token to gather Prometheus and Grafana API. Default: Undefined
+#
+# To create the ConfigMap:
+# $ echo -e 'GAHTER_MONIT_START_DATE="15 days ago"\nGAHTER_MONIT_QUERY_STEP="5m"' > env
```
sthaha, line 17: nit: typo GAHTER -> GATHER

collection-scripts/gather_monitoring
```diff
@@ -1,29 +1,208 @@
...
+# $ echo -e 'GATHER_MONIT_DISCOVERY_PREFIXES="apiserver_"' > env
+# $ oc create configmap must-gather-env -n openshift-monitoring --from-file=env=env
+#
+# References:
+# - Prometheus API: https://prometheus.io/docs/prometheus/latest/querying/api/
+
 #Safeguards
-set -o nounset
-set -o errexit
 set -o pipefail
```
sthaha, line 26: we need to ensure that the script works with -e -u set, otherwise it quite hard to track errors due to `typo`

collection-scripts/gather_monitoring
mtulio: mtulio~~

collection-scripts/gather_monitoring
mtulio: Done, thanks! :)

sferich888:
https://github.com/openshift/must-gather/pull/234 is a similar PR

collection-scripts/gather_monitoring
```diff
@@ -1,29 +1,208 @@
...
-BASE_COLLECTION_PATH="../must-gather"
-MONITORING_PATH="${BASE_COLLECTION_PATH}/monitoring"
+function err_report {
+    echo "ERROR: on line $1 from ${0}"
+}
+trap 'err_report $LINENO' ERR
+
+declare -r GATHER_MONIT_START_DATE_DEFAULT="7 days ago"
+declare -r GATHER_MONIT_END_DATE_DEFAULT="now"
+declare -r GATHER_MONIT_QUERY_STEP_DEFAULT="1m"
```
sferich888, line 37: Should this be set higher? to say 10 min to align with the way SAR breaks down data? 
How much data will this pull back (7 days worth of data in 1 min intervals) what does that translate to in MB or size on disk?

collection-scripts/gather_monitoring
```diff
@@ -1,29 +1,208 @@
...
+    prom_get_api "${API_URI}" 2> "${FILE_ERR}" | gzip > "${FILE_OK}"
+}
+
+#
+# Gather specific metrics by name.
+# Env var: GATHER_MONIT_CUSTOM_METRICS
+#
+function get_custom_metrics {
+
+    if [[ -z "${GATHER_MONIT_CUSTOM_METRICS:-}" ]]; then
```
sferich888, line 141: Should we provide an ERROR/WARNING here? to let people know that we didn't collect/get custom metrics? 

Should this link / explain how to collect custom metrics if they want to?

sferich888 (commented):
@davemulford this looks good to me; what are your thoughts on this?

sferich888:
Can and should this replace the https://github.com/openshift/must-gather/blob/master/collection-scripts/gather_monitoring file or script?

mtulio:
@sferich888 This PR, [1] and [2] are complementary. Both collects more information from monitoring stack. [1] collect only alerts, [2] collect more resources and here we have a alternative of entire Prometheus' DB dump to collect specific metrics without needing to write external scripts.

[1] https://github.com/openshift/must-gather/blob/master/collection-scripts/gather_monitoring
[2] https://github.com/openshift/must-gather/pull/234

collection-scripts/gather_monitoring
mtulio: > Should we provide an ERROR/WARNING here? to let people know that we didn't collect/get custom metrics?

Maybe a "Warning" as it is a **optional** config. I leave a dump of current config on the begging of script execution:

https://github.com/openshift/must-gather/pull/214/files#diff-a34d5a93515e38dfc675ecd82be2d0c92c9a384475fc1a6b157b311a262daebeR93-R103

> Should this link / explain how to collect custom metrics if they want to?

This could be a good idea, but I not sure where we can put this kind of information as the most reference of must-gather is in OCP doc. I [wrote some examples](https://github.com/openshift/must-gather/pull/214/files#diff-a34d5a93515e38dfc675ecd82be2d0c92c9a384475fc1a6b157b311a262daebeR16) on the script comment, but it's not readable by users. 
Any suggestion?

collection-scripts/gather_monitoring
mtulio: @sferich888  is not so easy to estimate those values. It will depends of the carnality for each metric - how often the job read those metrics on the targets (interval), how many targets the job has, how many labels/carnality each metric has, etc. 

AFIK the Prometheus api does not allow more than 11k datapoints for each query, so depending of the metric if use "7 days" && "1m" precision it will be big for one etcd metric, for instance.

Looking only to the number of labels, check the difference of time/data between collecting 1d for `up` and `etcd_request_duration_seconds_bucket` metrics.

Have a look into three examples bellow using:
[1] precision=`1m`, period=`1 day` for `up` metric
~~~
[must-gather-zqb7g] POD 2021-06-21T19:38:17.906651437Z INFO: Collecting custom metrics...
[must-gather-zqb7g] POD 2021-06-21T19:38:17.906651437Z INFO: Collecting metric /query_range?query=up
[must-gather-zqb7g] POD 2021-06-21T19:38:18.950175375Z INFO: Must-gather monitoring finished

$ du -sh mgm-1d-1m-up/*/monitoring/prometheus/metric-up.json.gz 
60K	mgm-1d-1m-up/quay-io-rhn-support-mrbraga-must-gather-sha256-7af103a5e6a1c74549dd8440ee3e611008a0504e3ade74496d6b3f7c4e7da0d4/monitoring/prometheus/metric-up.json.gz

$ zcat mgm-1d-1m-up/*/monitoring/prometheus/metric-up.json.gz  |jq -r ' .data.result[].metric.__name__'|wc -l
145
~~~
[2] precision=`1m`, period=`1 day` for `up` metric
~~~
[must-gather-r472v] POD 2021-06-21T20:01:07.337430185Z INFO: Collecting custom metrics...
[must-gather-r472v] POD 2021-06-21T20:01:07.337634071Z INFO: Collecting metric /query_range?query=up
[must-gather-r472v] POD 2021-06-21T20:01:27.791372093Z INFO: Must-gather monitoring finished


$ du -sh mgm-1d-10m-up/*/monitoring/prometheus/metric-up.json.gz 
8.0K	mgm-1d-10m-up/*/monitoring/prometheus/metric-up.json.gz

$ zcat mgm-1d-10m-up/*/monitoring/prometheus/metric-up.json.gz |jq -r ' .data.result[].metric.__name__'|wc -l
138
~~~
[3] precision=`1m`, period=`30 minutes` for `etcd_request_duration_seconds_bucket`
~~~
[must-gather-wbqs4] POD 2021-06-21T19:47:10.297731294Z INFO: Collecting metric /query_range?query=etcd_request_duration_seconds_bucket
[must-gather-wbqs4] POD 2021-06-21T19:47:15.665498273Z INFO: Must-gather monitoring finished

$ du -sh mgm-30m-1m-etcd/*/monitoring/prometheus/metric-etcd_request_duration_seconds_bucket.json.gz
2.2M	mgm-30m-1m-etcd/*/monitoring/prometheus/metric-etcd_request_duration_seconds_bucket.json.gz

$ zcat mgm-30m-1m-etcd/*/monitoring/prometheus/metric-etcd_request_duration_seconds_bucket.json.gz |jq -r ' .data.result[].metric.__name__'|wc -l
28208
~~~

mtulio: So, for that reason I believe that `GATHER_MONIT_START_DATE_DEFAULT` is already high, I would love to leave it by default `3 days` and if needed it can be chanced by CM.

mtulio: Also for that reason I cut a lot of "features" from initial PR that could impact in performance, like "Grafana metrics discovery". It could be a problem if enabled by default. IMO We need to improve it before become a option to run in CU envs.

collection-scripts/gather_monitoring
```diff
@@ -1,29 +1,208 @@
 #!/bin/bash
```
sthaha, line 1: @mtulio based on [comment from Damien](https://github.com/openshift/must-gather/pull/234#issuecomment-865167716), do you think renaming this script so that its usage would be to gather additional data that can be gathered on demand? 
 
```
oc adm must-gather -- /usr/bin/gather-monitoring-mertics
```
I also wonder if additional parameters can be passed to the script. e.g. 

```
oc adm must-gather -- /usr/bin/gather-monitoring-mertics --start-date='..' --end-date='...' --step='...'
```


cc: @dgrisonnet

collection-scripts/gather_monitoring
```diff
@@ -1,29 +1,208 @@
...
+# - GATHER_MONIT_QUERY_STEP: metric resolution to get from Prometheus. Default: "1m".
+# - GATHER_MONIT_CUSTOM_METRICS: list of additional metrics name to be collected. Default: Undefined
+# - GATHER_MONIT_DISCOVERY_PREFIXES: discovery metrics by prefixes. Default: ''
+# - GATHER_MONIT_TOKEN: optional service account token to gather Prometheus and Grafana API. Default: Undefined
+#
+# To create the ConfigMap:
+# $ echo -e 'GATHER_MONIT_START_DATE="15 days ago"\nGATHER_MONIT_QUERY_STEP="5m"' > env
+# $ echo -e 'GATHER_MONIT_CUSTOM_METRICS="up"' > env
+# $ echo -e 'GATHER_MONIT_DISCOVERY_PREFIXES="apiserver_"' > env
+# $ oc create configmap must-gather-env -n openshift-monitoring --from-file=env=env
```
sthaha, line 20: How does this configmap get injected into the must-gather pod?

sthaha: User will need to remember to clean this up as well don't they? 

It may be less magic if user can point to any configmap rather than create on in `openshift-monitoring` e.g. 

`oc adm must-gather -- /usr/bin/gather-monitoring-data  --use-cm=ns/cm-name`

collection-scripts/gather_monitoring
```diff
@@ -1,29 +1,208 @@
...
+#
+function get_discovery_metrics {
+
+    if [[ -z "${GATHER_MONIT_DISCOVERY_PREFIXES:-}" ]]; then
+        return
+    fi
+    echo "INFO: Collecting metrics by prefixes..."
+
+    prom_get_api "label/__name__/values" \
+        | python2 -c \
```
sthaha, line 163: is it better to just gather the data and do the parsing and extraction later / offline?

collection-scripts/gather_monitoring
```diff
@@ -1,29 +1,208 @@
...
+
+function oc_get {
+    oc get \
+        --certificate-authority="${CA_BUNDLE}" \
+        --token="${SS_TOKEN}" \
+        "$@"
+}
+
+function prom_get_api {
+    local API_URI="${PROM_API}/${1}"; shift
```
sthaha, line 118: NIT: could you please use UPPER for `globals` and snake_case for `locals` ?

collection-scripts/gather_monitoring
mtulio: @sthaha it was used in alternative of `jq`, that I deprecated it in previous review. It used online on the script to discovery the metrics (next loop).

collection-scripts/gather_monitoring
mtulio: > based on comment from Damien, do you think renaming this script so that its usage would be to gather additional data that can be gathered on demand?

@sthaha yes, it could be a good idea rename it to keep the roles separated from default "monitoring" collector. 

> I also wonder if additional parameters can be passed to the script. e.g.

If we will rename the script to run on demand, it will be a great idea, so we can avoid some extra arguments using CM - not sure if we can remove the CM due to long parameters that can be sent with long list of metrics names.

collection-scripts/gather_monitoring
mtulio: > How does this configmap get injected into the must-gather pod?

Upon created, the script load it [here : `function get_env`](https://github.com/openshift/must-gather/pull/214/files#diff-a34d5a93515e38dfc675ecd82be2d0c92c9a384475fc1a6b157b311a262daebeR52-R64)

> User will need to remember to clean this up as well don't they?

Unfortunately if you want to customize the "metrics collector", the CM need to be created before the execution and manually cleaned after it finished.

I didn't think in a life cycle of this CM. Could be a good idea send it as argument, but it still needing to be manually created/removed. Initially I thought in remove this CM after execution (on script) but I not sure if it will break the principle of MG to be a "super-get tool".

sferich888:
@mtulio can you rebase or close (*depending on the direction that this took)?

collection-scripts/gather_monitoring
mtulio: Done https://github.com/mtulio/must-gather/pull/21/commits/e6b6b276402ea92942571ce5461b03266dad8bfb

collection-scripts/gather_monitoring
mtulio: As you suggested, I renamed the script from `gather_monitoring` to `gather-monitoring-data`. I am implementing the flags.
https://github.com/mtulio/must-gather/pull/21/commits/62c19c3f9acd85a655446d0594f8f72103d01002

mtulio:
@dgrisonnet @sthaha ptal? I just finished the review from monitoring team. Now the script is sepparated from standard 
collector, the script will be responsible only to collect metrics. I also added flags to the script to be 
more flexible on runtime (sthaha++).

Please see bellow the Usage help, I leave it verbose to be easier for the user when running standalone script. I also logged some outputs from my local tests [here](https://github.com/mtulio/must-gather/pull/25).

~~~bash
$ ./gather-monitoring-data -h
Usage: ./gather-monitoring-data [options]

Available options:
    -h | --help                 Show this help
    -c | --use-cm  ns/cm_name   Specify the ConfigMap to load environment variables on the
                                format namespace/configMapName. Default is: None (load From variables)
    -s | --start-date date_str  Start date string. Default is: 1 day ago
    -e | --end-date date_str    End date string. Default is: now
    -S | --step duration        Query resolution step width in 'duration' format or float
                                number of seconds. Default is: 1m
    -o | --dest-dir path        Set a specific directory on the local machine to write gathered data to.
    -t | --timeout value        The length of time to gather data, in seconds. Default is: 15s
    --endpoint (API|tsdb)       TODO. Endpoint to be created, API queries or chunks from TSDB. Default: API.
    --query PromQuery           Prometheus Query to collect instant query
    --query-range PromQuery     Prometheus Query to collect time range
    --query-range-prefix Query  Prometheus Query range prefix to be discovered and collected.

Available environments variables to be used on ConfigMap:
    GATHER_BASE_PATH_DEFAULT="/must-gather"

    # Query setup
    GATHER_MONIT_START_DATE="1 day ago"
    GATHER_MONIT_END_DATE="now"
    GATHER_MONIT_QUERY_STEP="1m"

    # Queries
    GATHER_PROM_QUERIES=
    GATHER_PROM_QUERIES_RANGE=
    GATHER_PROM_QUERIES_RANGE_PREFIX=

Examples:
    # Simple collect simple query
    ./gather-monitoring-data --query "up"

    # Simple collect simple query and save to a custom directory
    ./gather-monitoring-data --query "up" --dest-dir ./monitoring-metrics

    # Query range collector from last hour
    ./gather-monitoring-data -s "1 hours ago" -e "now" -o metrics-1h --query-range "up"

    # Query range collector from last hour of metrics with prefix name "etcd_disk_"
    ./gather-monitoring-data -s "1 hours ago" -e "now" -o metrics-1h-prefix --query-range-prefix "etcd_disk_"

    # Create a ConfigMap to store variables
    echo "GATHER_MONIT_START_DATE='12 hours ago'
GATHER_MONIT_QUERY_STEP='1m'
GATHER_PROM_QUERIES='up'
GATHER_PROM_QUERIES_RANGE='instance:node_load1_per_cpu:ratio
changes(etcd_server_leader_changes_seen_total[1h])'
GATHER_PROM_QUERIES_RANGE_PREFIX='etcd_disk_
apiserver_flowcontrol'" > ./env
    oc create configmap must-gather-metrics-env -n mgm --from-file=env=env
    ./gather-monitoring-data --use-cm mgm/must-gather-metrics-env

~~~

mtulio:
The kubectl plugin [`promdump`](https://github.com/ihcsim/promdump) could complement, or override. the work done in this PR, as it can list, download and restore persistent blocks on Prometheus, one alternative of querying API VS download the entire DB. That could be a good choice, and I can write as `--endpoint` option, that I left as `TODO` in my previous development to support flags.

What's your opinion @dgrisonnet @sthaha ? 

Regarding the failed CI job, I am not sure if it's related with the last change or a CI fail due the amount of `timed out` when running the test.

PR #217, opened 2021-03-16
add ovs db per node to gather network logs

msherif1234:
Description
extending must gather network logging to include ovs db dump to enhance debugging.

Type of change
Enhance debugging of openshift cluster.

How Has This Been Tested?
created Openshift cluster then run 
```
oc adm must-gather gather_network_logs
```


msherif1234:
instead of running debug container we will mount ovs db inside kubenode and just copy it from there

PR #225, opened 2021-04-01
Bug 1907353: Add ovs related services to GENERAL_SERVICES

jluhrsen:
- also alpha ordering of the services

Signed-off-by: Jamo Luhrsen <jluhrsen@gmail.com>

### DIFF
--- a/collection-scripts/gather_service_logs
+++ b/collection-scripts/gather_service_logs
@@ -5,7 +5,8 @@ BASE_COLLECTION_PATH="/must-gather"
 ROLES=${1:-master}         ### Defaults to only collecting things from Masters
 
 # Service Lists
-GENERAL_SERVICES=(kubelet crio machine-config-daemon-firstboot machine-config-daemon-host)
+GENERAL_SERVICES=(crio kubelet machine-config-daemon-firstboot machine-config-daemon-host NetworkManager openvswitch
+                  ovs-configuration ovsdb-server ovs-vswitchd)
 MASTER_SERVICES+=(${2:-"${GENERAL_SERVICES[@]}"})
 MASTER_SERVICES+=()        ### Placeholder to extend Master only services
 NODE_SERVICES+=(${2:-"${GENERAL_SERVICES[@]}"})


danwinship:
Background here is that we used to get the OVS logs as part of must-gather because OVS ran in pods. Then in 4.6 and 4.7 we still had "OVS pods", but all they did was "`tail -f /var/log/openvswitch/*.log`". We want to get rid of the useless pods, but we need the OVS logs to still be available in must-gather output by default. They are not large (except sometimes when things are horribly broken, and then they're large-but-important).

collection-scripts/gather_service_logs
```diff
@@ -5,7 +5,8 @@ BASE_COLLECTION_PATH="/must-gather"
 ROLES=${1:-master}         ### Defaults to only collecting things from Masters
 
 # Service Lists
-GENERAL_SERVICES=(kubelet crio machine-config-daemon-firstboot machine-config-daemon-host)
+GENERAL_SERVICES=(crio kubelet machine-config-daemon-firstboot machine-config-daemon-host openvswitch ovs-configuration
```
trozet, line 8: could we also add NetworkManager to this list? With ovs-configuration failures, we usually need to also see the NM logs to debug.

jluhrsen: @trozet it's there now. tested locally and it works.

jluhrsen:
@sferich888, is this ok or is there a different approach I should take?

jluhrsen:
@squeed @trozet , can you confirm that this PR will just be retrieving the same service logs we are getting anyway with
the ovs daemonset?

@sferich888, once we merge https://github.com/openshift/cluster-network-operator/pull/1028 we will not be getting these
same logs, which is why I added this one in must-gather.

squeed:
Yup; this won't add any appreciable size to the payload (or the size) - this is literally the same content as some pod logs that we already always gather (and we'll remove those pods when this merges).

jluhrsen:
/retitle Bug 1907353: Add ovs related services to GENERAL_SERVICES

PR #226, opened 2021-04-06
Add multi-networkpolicy object in must-gather

s1061123:


collection-scripts/gather_network_logs
```diff
@@ -19,6 +19,8 @@ function gather_multus_data {
 	  oc -n openshift-multus logs ${MULTUS_POD} > "${NETWORK_LOG_PATH}"/"${MULTUS_POD}".log 2>&1 & PIDS=($!)
   done
 
+  oc get multi-networkpolicy -o yaml --all-namespaces > "${NETWORK_LOG_PATH}"/multi_networkpolicy 2>&1 & PIDS=($!)
```
sferich888, line 22: @davemulford should we split up some network collection (like this); out of the gatehr_network_logs script (because they are general vs host dependent)? 

@knobunc what is your thought on this?

PR #227, opened 2021-04-20
Add collection-script for sosreports

andreaskaris:
This allows users to gather sosreports on all or on a subset of
nodes of their cluster. This follows the same idea as the
gather_core_dumps scripts. This is not run by default, instead users
must explicitly run this with /usr/bin/gather-sosreports
and they can provide a label selector and overwrite for image
and registry, e.g.:
    -l kubernetes.io/hostname=openshift-worker-0
    -r registry.example.com:5000
    -i support-tools:0.0.2

Signed-off-by: Andreas Karis <ak.karis@gmail.com>

### DIFF
--- a/collection-scripts/gather-sosreports
+++ b/collection-scripts/gather-sosreports
@@ -1,0 +1,184 @@
+#!/bin/bash
+
+BASE_COLLECTION_PATH="/must-gather"
+SOSREPORT_DUMP_PATH=${OUT:-"${BASE_COLLECTION_PATH}/node_sosreports"}
+# 20 minutes timeout for a sosreport pod should be reasonablef
+POD_TIMEOUT=1200
+# Do not run a sosreport if the minimum availble GB is not met on /var/tmp
+MIN_AVAILBLE_GB=20
+
+# Gather a sosreport on a specific node
+# Parameters:
+# $1 - node name: argument to oc debug node in format node/<name>
+# $2 - registry:  URL to custom registry for image overwrite of toolbox image
+# $3 - image:     image name for image overwrite of toolbox image
+function gather_sosreport() {
+	local node=$1
+	local registry=$2
+	local image=$3
+
+	# Because we have to chroot, sudo and start a toolbox container
+	# it is easier to run this in a script which is then copied to
+	# the debug pod
+	local tmp_file=$(mktemp)
+	cat <<'EOF' > $tmp_file
+#!/bin/bash
+
+chroot /host /usr/bin/sudo --preserve-env /bin/bash <<'EOT'
+# Defaults first
+REGISTRY=registry.redhat.io
+IMAGE=rhel8/support-tools
+TOOLBOX_NAME=toolbox-"$(whoami)-sosreport-$(uuidgen | cut -b-8)"
+TOOLBOXRC="/root/.toolboxrc"
+
+# Second if .toolboxrc file exists, take its values
+if [ -f "${TOOLBOXRC}" ]; then
+    echo ".toolboxrc file detected, overriding defaults..."
+    source "${TOOLBOXRC}"
+fi
+
+# Third if user overrides are set, use those
+USER_REGISTRY=""
+USER_IMAGE=""
+if [ "${USER_REGISTRY}" != "" ]; then
+    REGISTRY="${USER_REGISTRY}"
+fi
+if [ "${USER_IMAGE}" != "" ]; then
+    IMAGE="${USER_IMAGE}"
+fi
+
+# Now, concat registry and image
+TOOLBOX_IMAGE="${REGISTRY}"/"${IMAGE}"
+
+podman pull --authfile /var/lib/kubelet/config.json "$TOOLBOX_IMAGE"
+
+podman run \
+  --rm \
+  --hostname $(hostname) \
+  --name "$TOOLBOX_NAME" \
+  --privileged \
+  --net=host \
+  --pid=host \
+  --ipc=host \
+  -e HOST=/host \
+  -e NAME="$TOOLBOX_NAME" \
+  -e IMAGE="$IMAGE" \
+  --security-opt label=disable \
+  --volume /run:/run \
+  --volume /var/log:/var/log \
+  --volume /etc/machine-id:/etc/machine-id \
+  --volume /etc/localtime:/etc/localtime \
+  --volume /:/host \
+   "$TOOLBOX_IMAGE" \
+  sosreport -k crio.all=on -k crio.logs=on --batch --all-logs 2>/dev/null | tee /tmp/must_gather_sosreport.log
+EOT
+EOF
+	# Set overrides if a user provided registry and image are provided
+	sed -i "s#^USER_REGISTRY=#USER_REGISTRY=\"$registry\"#" ${tmp_file}
+	sed -i "s#^USER_IMAGE=#USER_IMAGE=\"$image\"#" ${tmp_file}
+
+	# Get debug pod's name and exit right away
+	local debug_pod=$(oc debug ${node} -o jsonpath='{.metadata.name}')
+
+	# Start Debug pod force it to stay up until POD_TIMEOUT is reached
+	oc debug ${node} -- /bin/bash -c "sleep ${POD_TIMEOUT}" > /dev/null 2>&1 &
+
+	# Wait 300 seconds for this pod to come up
+	for i in {1..10}; do
+	    echo "Waiting for pod ${debug_pod} to be ready. Try $i"
+	    sleep 2
+	    oc wait --for=condition=Ready pod/${debug_pod} --timeout=30s
+	    ret_val="$?"
+	    if [ "$ret_val" != "0" ] && [ "$i" -eq "10" ]; then
+	        echo "Debug pod did not spawn for node ${node}."    
+	        return
+	    elif [ "$ret_val" == "0" ]; then
+	        break
+            fi
+	done
+
+	local available_kb=$(oc exec ${debug_pod} -- df /host/var/tmp | tail -n+2 | awk '{print $4}')
+	local available_gb=$[ $available_kb / 1024 /1024 ]
+	if [ "$available_gb" -lt "$MIN_AVAILBLE_GB" ]; then
+	    echo "Not enough disk space available on ${node}"
+	    echo "Not creating sosreport. Skipping."
+	    oc delete pod "${debug_pod}"
+	    return
+	fi
+
+	# Copy script to the debug pod
+	oc cp --loglevel 1 ${tmp_file} "${debug_pod}":/tmp/sosreport.sh
+
+	# Execute script
+	oc exec ${debug_pod} -- /bin/bash /tmp/sosreport.sh
+
+	# Copy the log output to the local host to track why a sosreport failed, 
+	# if it failed
+	local node_simple="${node/node\//}"
+	local log_dest="${SOSREPORT_DUMP_PATH}/must_gather_sosreport.${node_simple}.log"
+	oc cp --loglevel 1 \
+	  ${debug_pod}:/host/tmp/must_gather_sosreport.log \
+	  ${log_dest}
+
+	# Get sosreport archive name
+	local tmp_sosreport_file=$(
+	    grep "tar.xz" ${log_dest}  | awk '{print $1}'
+        )
+
+	# Abort if the name is empty
+	if [ "$tmp_sosreport_file" == "" ]; then
+		echo "Could not get sosreport file name"
+		return
+	fi
+
+	# Copy the sosreport to the local host
+	oc cp --loglevel 1 \
+	  ${debug_pod}:${tmp_sosreport_file} \
+	  ${SOSREPORT_DUMP_PATH}/$(basename ${tmp_sosreport_file})
+
+	# Delete the original sosreport archive
+	# This is necessary as the sosreport is generated in /host/var/tmp on
+	# the node itself, and not inside the container
+	oc exec ${debug_pod} -- rm -f ${tmp_sosreport_file}
+
+	# Clean up the debug pod
+	oc delete pod "${debug_pod}"
+
+	# Clean up the temp file with the script
+	rm -f ${tmp_file}
+}
+
+# Create a dedicated directory for sosreports
+mkdir -p ${SOSREPORT_DUMP_PATH}
+
+LABEL=""
+IMAGE=""
+REGISTRY=""
+while getopts "l:i:r:" OPTION; do
+    case $OPTION in
+    i)
+        IMAGE="$OPTARG"
+        ;;
+    l)
+        LABEL="-l $OPTARG"
+        ;;
+    r)
+        REGISTRY="$OPTARG"
+        ;;
+    esac
+done
+
+# Gather a list of nodes based on a specified label
+nodes=$(oc get nodes -o name $LABEL)
+
+# Loop through all nodes and gather sosreports for each of them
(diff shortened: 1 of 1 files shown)


andreaskaris:
I'm putting this up here as a suggestion / implementation for: https://issues.redhat.com/browse/RFE-1810

The user documentation is here: https://github.com/andreaskaris/must-gather-sosreport

collection-scripts/gather-sosreports
```diff
@@ -0,0 +1,132 @@
...
+  --volume /:/host \
+   "$TOOLBOX_IMAGE" \
+  sosreport -k crio.all=on -k crio.logs=on --batch --all-logs 2>/dev/null | tee /tmp/must_gather_sosreport.log
+EOT
+EOF
+	sed -i "s#^USER_REGISTRY=#USER_REGISTRY=\"$registry\"#" ${tmp_file}
+	sed -i "s#^USER_IMAGE=#USER_IMAGE=\"$image\"#" ${tmp_file}
+
+	# Get debug pod's name and exit right away
+	debug_pod=$(oc debug --to-namespace=default ${node} -o jsonpath='{.metadata.name}')
```
sferich888, line 68: We don't want to spin this pod up in the 'default namespace'. Every must-gather invocation runs in its own namespaces; and we will want to launch this pod in that namespaces as well.

collection-scripts/gather-sosreports
```diff
@@ -0,0 +1,132 @@
...
+	if [ "$tmp_sosreport_file" == "" ]; then
+		echo "Could not get sosreport file"
+		return
+	fi
+
+	oc cp --loglevel 1 -n default \
+	  ${debug_pod}:${tmp_sosreport_file} \
+	  ${SOSREPORT_DUMP_PATH}/$(basename ${tmp_sosreport_file})
+
+	oc exec -n default ${debug_pod} -- rm -f ${tmp_sosreport_file}
```
sferich888, line 95: Is this necessary? When the must-gather pod is done this will be dropped/deleted as these pods don't have persistent storage. 

In short this adds time to the collection that I don't think it necessary.

collection-scripts/gather-sosreports
```diff
@@ -0,0 +1,132 @@
...
+	oc wait -n "default" --for=condition=Ready pod/${debug_pod} --timeout=30s
+
+	# Copy script to the debug pod
+	oc cp --loglevel 1 -n default $tmp_file "${debug_pod}":/tmp/sosreport.sh
+
+	# Execute script
+	oc exec -n default ${debug_pod} -- /bin/bash /tmp/sosreport.sh
+
+	# get sosreport archive name
+	tmp_sosreport_file=$(oc exec -n default ${debug_pod} -- grep 'tar.xz' /host/tmp/must_gather_sosreport.log  | awk '{print $1}')
```
sferich888, line 83: We should save this archive to the must-gather collection to look at / better understand why sosreport collections fail; should they fail.

collection-scripts/gather-sosreports
```diff
@@ -0,0 +1,132 @@
...
+EOT
+EOF
+	sed -i "s#^USER_REGISTRY=#USER_REGISTRY=\"$registry\"#" ${tmp_file}
+	sed -i "s#^USER_IMAGE=#USER_IMAGE=\"$image\"#" ${tmp_file}
+
+	# Get debug pod's name and exit right away
+	debug_pod=$(oc debug --to-namespace=default ${node} -o jsonpath='{.metadata.name}')
+
+	#Start Debug pod force it to stay up until removed in "default" namespace
+	oc debug --to-namespace=default ${node} -- /bin/bash -c 'sleep infinity' > /dev/null 2>&1 &
```
sferich888, line 71: This seems dangerous; as we are in theory starting a pod; that runs forever, and it might hang the over all collection, if this pod can't be deleted/stopped for some reason.

collection-scripts/gather-sosreports
```diff
@@ -0,0 +1,132 @@
...
+
+	sleep 2
+	oc wait -n "default" --for=condition=Ready pod/${debug_pod} --timeout=30s
```
sferich888, line 74: Does this need to be wrapped in a re-try loop? IE: should we wait for N seconds to get the pod running and if it can't start we should exit with an error reason.

collection-scripts/gather-sosreports
```diff
@@ -0,0 +1,132 @@
...
+        REGISTRY="$OPTARG"
+        ;;
+    esac
+done
+
+nodes=$(oc get nodes -o name $LABEL)
+
+for n in $nodes ; do
+	echo "Gathering sosreport for $n"
+	if [ "$REGISTRY" != "" ]; then
```
sferich888, line 125: how are these values set/passed into to the container/script? I see that https://github.com/openshift/must-gather/pull/227/files#diff-1261786f8e9ac4fc6040be6bbb412a9e5a1204b9cb0c0fe6955c30b274d6ea53R6 uses variables; but I don't see how these variables are passed into this gather script fo setting the values.

collection-scripts/gather-sosreports
```diff
@@ -0,0 +1,132 @@
...
+    l)
+        LABEL="-l $OPTARG"
+        ;;
+    r)
+        REGISTRY="$OPTARG"
+        ;;
+    esac
+done
+
+nodes=$(oc get nodes -o name $LABEL)
```
sferich888, line 121: We want/need an option to limit this to specific nodes. In short if we have 200 'workers' we will want to be able to specify or limit that collection better. 

We do this in several other collectors; so I think we need to move away from just limiting this to a 'label'.

sferich888 (changes requested):
I think we need to review this a bit more; @andreaskaris can you outline the motivation for the need of for sosreports exposed in this manner? 

In general; you should be using must-gather to look at cluster issues; and only using sosreport to get a 'cluster back to functional' (so that you could run a must-gather), or using sosreport in narrowly scoped space (on a node where a particular issue is happening). In general my fear / issue with this is that we are adding a collector/collection of information that we typically reserve for situations where must-gather will not run.  

That and the size of must-gather will expand / grow exponentially as a result of this. 

In general while collecting a sosreport in this way is what we document (one way we expose this), its not the recommended path to get a sosreport (the recommend path is still with ssh and starting a pod).

collection-scripts/gather-sosreports
andreaskaris: Hi,

>> I think we need to review this a bit more; @andreaskaris can you outline the motivation for the need of for sosreports exposed in this manner?

This is a customer ask, coming from a high priority customer.  https://issues.redhat.com/browse/RFE-1810 contains more details and the reason behind the ask.

>> In general; you should be using must-gather to look at cluster issues; and only using sosreport to get a 'cluster back to functional' (so that you could run a must-gather)

In the field, that unfortunately does not work, at least for our enhanced support teams. We often ask our customers to generate sosreports to get node local events and configuration (imagine SR-IOV issues, CNV issues) in additiona to must-gather reports.  So we end up asking our customers to gather a must-gather, then a specific must-gather such as CNV or SR-IOV, and in addition, we need sosreports to analyze a baremetal node's status (imagine kernel bugs ,etc.)
In the enhanced support team, we work with OpenShift in baremetal events and sosreports are the only tool that captures some of the critical data for baremetal nodes. The must-gather tool already contains a script to gather vmcores; that's a highly specialized and lowlevel operation and falls into the same category as gathering  sosreports.
Given that for our team at least it is a very common task to request sosreports on top of must-gathers, the customer must go into every single node and gather and pull a sosreport. That's cumbersome and the best tool to gather a subset of nodes is the must-gather tool.

>> (on a node where a particular issue is happening)

That's what the `-l` label invocation takes care of.

>> In  general my fear / issue with this is that we are adding a collector/collection of information that we typically reserve for situations where must-gather will not run.

As stated above, we have several high profile customers where our custom policy is: generate and gather a must-gather, and generate and gather a sosreport, even if the cluster is running. That might be a shortcoming of the other must-gather scripts, but then I strongly believe that sosreports fill a different purpose, and also one that's beyond breaking issues. Again, NFV, CNV, performance issues, kernel issues, OOM situations, all of these are things that sosreports capture and allow to analyze.

>> That and the size of must-gather will expand / grow exponentially as a result of this.

The same as the core dump collector script, I would like to add this as a standalone script which is not run by default. A user has to explicitly call the script, otherwise no sosreport will be run.

>> In general while collecting a sosreport in this way is what we document (one way we expose this), its not the recommended path to get a sosreport (the recommend path is still with ssh and starting a pod).

In order to automate this and run this via `must-gather`, I need to use `oc debug` and a toolbox container.

andreaskaris: > We don't want to spin this pod up in the 'default namespace'. Every must-gather invocation runs in its own namespaces; and we will want to launch this pod in that namespaces as well.

I will change this. However, then we allso must change:
https://github.com/openshift/must-gather/blob/cbf587fe6e4c97196e84aea68ad7a57210035e92/collection-scripts/gather_core_dumps#L14

Because that's where I got the `default` namespace from, in the first place :-)

collection-scripts/gather-sosreports
andreaskaris: I'll change that.

collection-scripts/gather-sosreports
andreaskaris: I'll change that.

collection-scripts/gather-sosreports
andreaskaris: I'll follow https://github.com/openshift/must-gather/blob/cbf587fe6e4c97196e84aea68ad7a57210035e92/collection-scripts/gather_core_dumps#L14 and I'll set a reasonably high limit, say nodecount * some reasonable timeout.

collection-scripts/gather-sosreports
andreaskaris: Yes, good suggestion. I'l change that.

collection-scripts/gather-sosreports
andreaskaris: https://github.com/andreaskaris/must-gather-sosreport
~~~
oc adm must-gather --timeout=60m  --image=quay.io/akaris/must-gather-sosreport:latest -- /bin/bash /usr/bin/gather-sosreports -l kubernetes.io/hostname=openshift-worker-0 -r registry.example.com:5000 -i support-tools:0.0.2
~~~

collection-scripts/gather-sosreports
andreaskaris: When it comes to scheduling, we can schedule pods on specific nodes with affinity or with nodeSelector. 
The nodeSelector however bypasses the scheduler. 
So, for a decently working kubernetes cluster, it is mandatory that every node has a `kubernetes.io/hostname=(...)` label, otherwise the nodeaffinity feature does not work.

Especially in the context of OpenShift, it is safe to assume that every node has a `kubernetes.io/hostname` label, because so many other features need it.

So, as a matter of simplication, I thought it would be best to forgo another command line parameter and simply go by label:
~~~
-l kubernetes.io/hostname=openshift-worker-0
-l node-role.kubernetes.io/worker=
~~~

That said, I just realized that I do not add tolerations (so I cannot gather sosreports on the masters with the current approach).

I'm open for any other suggestion --- should there just be an `-h` `--host` switch?

Thanks for all your feedback, I'll work on improving this and I'll submit a new PR.

andreaskaris:
I uploaded a new version with most of the points addressed. 

I also only have the script create a sosreport if there is sufficient space available on /host/var/tmp

collection-scripts/gather-sosreports
andreaskaris: Actually, this was necessary. The sosreport is created on /host/var/tmp and so the file must be deleted after copying it out.

andreaskaris:
Ping

andreaskaris:
We can also convert this into a draft and move the conversation there. I'd just like to get some traction for this, I'm happy to change / adjust the script or think and talk about different options.

andreaskaris:
Qq - I tested the script so far only with a dedicated image: https://github.com/andreaskaris/must-gather-sosreport

When I try to build this repository directly, the build fails because I cannot pull:

*  registry.svc.ci.openshift.org/openshift/release:golang-1.12 AS builder
* registry.svc.ci.openshift.org/openshift/origin-v4.0:cli

(from https://github.com/openshift/must-gather/blob/master/Dockerfile)

Is there a decent way for me to test this with another base image?

Thanks!

andreaskaris:
I just talked to the customer and they would prefer to gather sosreports through the API, so either with must-gather or with something similar.

The reason is that they can access the API and run must-gathers from several different places in their org, but ssh access is limited to their bastion node only.

While I personally still love the `sos collect` approach, I will also reopen the pull request.

Do we have any other ideas?

Thanks!

collection-scripts/gather-sosreports
```diff
@@ -0,0 +1,184 @@
...
+  -e NAME="$TOOLBOX_NAME" \
+  -e IMAGE="$IMAGE" \
+  --security-opt label=disable \
+  --volume /run:/run \
+  --volume /var/log:/var/log \
+  --volume /etc/machine-id:/etc/machine-id \
+  --volume /etc/localtime:/etc/localtime \
+  --volume /:/host \
+   "$TOOLBOX_IMAGE" \
+  sosreport -k crio.all=on -k crio.logs=on --batch --all-logs 2>/dev/null | tee /tmp/must_gather_sosreport.log
```
TurboTurtle, line 73: One thing to consider here is varying versions of `sos` available on each node/image, and the ability to have users pass/change options as needed.

For `sos collect` this is handled on a per-node basis, meaning `collect` is aware of the capabilities of sos on each individual node. If plugins, options, etc... are passed by the user but aren't supported by particular nodes, the options are trimmed for those nodes alone to still allow a successful command. Invalid options or more specifically invalid plugin names will otherwise result in a failed collection.

Then you also need to consider how do you allow users to disable arbitrary collections (e.g. `-n $plugin_name` or `--skip-commands=$some_cmd`) that may be causing issues for otherwise "standard" sos runs?

Tangentially, I would very strongly recommend dropping `--all-logs` as a default as that removes all size limiting which will not only balloon the file archive size on disk, but will also significantly increase the memory usage of sos. Further, the combination of `-k {docker|podman|crio}.all=on -k {docker|podman|crio}.logs=on` is something I have been advocating against for some time as a default configuration as that means "give me all logs (everything the container sent to stdout) for every container that exists on this host even if they aren't currently running". Combined with `--all-logs` this can quickly become extremely expensive in terms of disk and memory usage.

palonsoro: From a "field" perspective of almost 3 years in Support and 2 more as an OCP customer, OCP troubleshooting (specially OCP4) implies looking what happens in a number of containers again and again. 

I end up **always** requesting all the logs from all the containers and I do need them 95% of the times (not an exaggeration). Maybe I don't need every single container logs, but asking sosreport and then logs of the relevant containers is usually not an acceptable approach, because the container may have vanished or its logs may have been rotated by the time you ask them separately.

On the other hand, bigger memory consumption and disk stress have been a problem only in very very few scenarios, where the node was so broken that I also needed to request other usually working sosreport modules to be disable with `-n`, i.e. these were very exceptional situations. 

And in what regards sosreport size... must-gathers are usually much bigger than a single sosreport (even after audit logs are not collected by default), so our customers are kind of used to share large data sets already.

So if these options are not made default, I foresee myself and other colleagues just always asking for them, because we are going to need them. If they are not provided at all, maybe we would need to build an alternative and not use this, because in short: We do need them most of the times.

PR #235, opened 2021-06-10
Bug 1958405: add etcd health logs to audit

hexfusion:
https://github.com/openshift/cluster-etcd-operator/pull/596 adds health audit logs for etcd. This PR adds those logs to the audit_logs gather script. This script is already known by support and users and the etcd state should not be large in size. Piggybacking the logs along with the other control-plane components seems logical.

alternative : https://github.com/openshift/must-gather/pull/236

PR #236, opened 2021-06-10
add health audit log collection for etcd

hexfusion:
This PR is an alternative to #235 and introduces the health audit log to must gather. The naming this generic in case we decide to extend the pattern to other components.

/hold


deads2k:
I prefer https://github.com/openshift/must-gather/pull/235.  thanks

/close

PR #243, opened 2021-06-23
Minor fixes for gather_network_logs

npinaeva:
Fix ip route gathering (command should be evaluated inside OVNKUBE_NODE_POD)
Fix evaluating if ipsec is enabled (condition evaluation previously failed, always returning False condition result)
Add How to run section to README.md

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,3 +10,7 @@ Data collection scripts are kept in `./collection-scripts`.  The content of that
 The data collection scripts should only include collection logic for components that are included as part of the OpenShift
 CVO payload.  Outside components are encouraged to produce a similar "must-gather" image, but this is not the spot to be
 included.
+
+## How to run
+You can run `must-gather` on a cluster with `oc adm must-gather`. 
+Use `-h` flag to see available options.
--- a/collection-scripts/gather_network_logs
+++ b/collection-scripts/gather_network_logs
@@ -129,9 +129,6 @@ function gather_ovn_kubernetes_nodes_data {
       "ethtool -k \$(ip route show default | cut -d ' ' -f5)" > "${NETWORK_LOG_PATH}"/"${NODE}"_ethtool_offload &
       PIDS+=($!)
 
-      # collect ovs log files from node
-      collect_node_log_files "${NODE}" openvswitch/ovs-vswitchd.log openvswitch/ovsdb-server.log
-
       # collect_service_logs queries logs from systemd journal (default syslog log level for ovs modules is ERR)
       collect_service_logs "${NODE}" openvswitch ovs-vswitchd ovsdb-server ovs-configuration
   done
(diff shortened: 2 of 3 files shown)


collection-scripts/gather_network_logs
```diff
@@ -237,6 +242,22 @@ function gather_kuryr_nodes_data {
   done
 }
 
+function collect_node_log_files {
+# Call format: NODE [file_paths]
+# Collects log files from given paths,
+# path should be relative to the node's /var/logs/ folder,
+# e.g. openvswitch/ovs-vswitchd.log will be collected to "$NETWORK_LOG_PATH"/"$NODE"_ovs-vswitchd.log
```
trozet, line 249: we already capture ovs-vswitchd.log as part of must-gather now:
https://github.com/openshift/must-gather/blob/cbf587fe6e4c97196e84aea68ad7a57210035e92/collection-scripts/gather_service_logs#L9

The issue is we need to set the logging levels in it, by doing ovs-vsctl vlog/set INFO for example. It looks like we just use to set it to INFO as shown here:
https://github.com/openshift/cluster-network-operator/blob/release-4.6/bindata/network/ovn-kubernetes/006-ovs-node.yaml#L92

We previously ran this OVS container in 4.6 that took care of this, but we no longer have it since 4.7. So log levels are not getting set correctly. I think you will need a PR in CNO to set the log level for OVS in the ovnkube-node template:
https://github.com/openshift/cluster-network-operator/blob/release-4.6/bindata/network/ovn-kubernetes/ovnkube-node.yaml

or alternatively you could put it in MCO:
https://github.com/openshift/machine-config-operator/blob/master/templates/common/_base/files/configure-ovs-network.yaml

npinaeva: Thanks for the comment, I will open different PR for the bug fix, and leave minor fixes in this one

npinaeva:
Reproducing: run `oc adm must-gather -- gather_network_logs`

Before
In the stdout:
/usr/bin/gather_network_logs: line 256: [-z: command not found
netlink error: no device matches name (offset 24)
netlink error: No such device

In gathered files:
network_logs/<node name>_ethtool_driver
network_logs/<node name>_ethtool_offload - empty files

After
No errors in stdout, files are filled with output

npinaeva:
@davemulford, @sferich888: I think this is the final version of the PR, added a comment about testing changes. I will appreciate if you could take a look.

dcbw:
/lgtm
@sferich888 what do you think?

README.md
```diff
@@ -10,3 +10,7 @@ Data collection scripts are kept in `./collection-scripts`.  The content of that
 The data collection scripts should only include collection logic for components that are included as part of the OpenShift
 CVO payload.  Outside components are encouraged to produce a similar "must-gather" image, but this is not the spot to be
 included.
+
+## How to run
```
sferich888, line 14: Should we link to https://docs.okd.io/latest/support/gathering-cluster-data.html#support_gathering_data_gathering-cluster-data (or something similar on how to run advanced commands, later  in these docs)?

collection-scripts/gather_network_logs
```diff
@@ -253,10 +255,10 @@ elif [[ "${NETWORK_TYPE}" == "kuryr" ]]; then
     gather_kuryr_nodes_data
     gather_kuryr_data
 elif [[ "${NETWORK_TYPE}" == "ovnkubernetes" ]]; then
-    if [-z "$IPSEC_PODS"]; then
+    if [ -z "$IPSEC_PODS" ]; then
         echo "INFO: No ovn-ipsec pods exist, tunnel traffic will be unencrypted"
     else 
-        echo "INFO: ovn-ipsec is enabled, tunnel traffic should be encryted"  
+        echo "INFO: ovn-ipsec is enabled, tunnel traffic should be encrypted"
```
sferich888, line 261: With this being true is running `gather_ovn_ipsec_data` valuable? 
Will we be decrypting it to use what we collect? If we do decrypt it do we need to tell the user? It doesn't look like we collect any traffic only how we shape/direct the traffic so these question may not matter, and if so I don't know if the messages/logs are needed (but they are not harmful).

sferich888 (approved):

README.md
npinaeva: Sure, that may be useful

PR #244, opened 2021-06-24
node: Gather rpm-ostreed

cgwalters:
For e.g. https://bugzilla.redhat.com/show_bug.cgi?id=1958812

### DIFF
--- a/collection-scripts/gather_service_logs
+++ b/collection-scripts/gather_service_logs
@@ -6,7 +6,7 @@ ROLES=${1:-master}         ### Defaults to only collecting things from Masters
 
 # Service Lists
 GENERAL_SERVICES=(crio kubelet machine-config-daemon-firstboot machine-config-daemon-host NetworkManager openvswitch
-                  ovs-configuration ovsdb-server ovs-vswitchd)
+                  ovs-configuration ovsdb-server ovs-vswitchd rpm-ostreed)
 MASTER_SERVICES+=(${2:-"${GENERAL_SERVICES[@]}"})
 MASTER_SERVICES+=()        ### Placeholder to extend Master only services
 NODE_SERVICES+=(${2:-"${GENERAL_SERVICES[@]}"})


cgwalters:
Can someone remind me actually why we don't gather the *entire* journal?

sferich888:
> Can someone remind me actually why we don't gather the _entire_ journal?

We don't collect the _entire_ journal for 2 reasons: 

1. Collecting data from a host (to inspect the host) is more of a sosreport (collector) responsibility (separation of duties).
    - See: https://github.com/sosreport/sos/blob/master/sos/report/plugins/logs.py#L66 (with the logs plugin) for how this is done/gated (with options/controls). 
    - **Note**: the os-tree plugin - https://github.com/sosreport/sos/blob/master/sos/report/plugins/ostree.py (doesn't collect any logs)
    - **Note**: The OpenShift plugin does not review/collect anything for os-tree (which I think is ok) - https://github.com/sosreport/sos/blob/master/sos/report/plugins/openshift.py#L160
2. Size considerations (of the must-gather archive).
    - To kee must-gathers (smallish) we have to reduce the logs we collect (because they consistently represent the greatest amount of space used in an archive). The kublet and cri-o are VERY noisy. 
    - To do this we limit the logs collected to [~7 days by default ](https://github.com/openshift/must-gather/blob/master/collection-scripts/gather_service_logs_util#L28)

cc: @control-d @TurboTurtle

My general concern with this is how much data ~7 days of yet another services logs will add (probably not much), but outside of that I don't see any reason not to add this (provided we are using this on the regular in CI/bugs for RHCOS).

cc: @jholecek-rh we may want to consider insights changes (with data collection) to get the the insights operator to collect the rpm-ostree status and a set of logs (on the regular too).

sferich888:
/lgtm
/hold 

I want to put a hold on this for @davemulford and others to review (however I think this is ok - in general; but it may not go far enough for the RHCOS team's need). @cgwalters to ack if he understand the limitations of when/how this is called/run.

davemulford:
This looks fine to me, too, as we only gather the last 7 days of logs. Do we know what sort of size this would be for a busy cluster?

cgwalters:
> This looks fine to me, too, as we only gather the last 7 days of logs. Do we know what sort of size this would be for a busy cluster?

This service only logs when we're actively updating basically so...very little.

cgwalters:
> To do this we limit the logs collected to ~7 days

This is fine.

> , on control plan hosts by default 

This isn't - actually in general what we care about here is the service logs on a failing node.  We've already changed the MCD to include these logs in its pod logs in some cases, but it's clearly more robust if we just gather it directly.  

Perhaps we could change the collection logic here to look especially at nodes which are listed as degraded by the MCO?

sferich888:
> > To do this we limit the logs collected to ~7 days
> 
> This is fine.
> 
> > , on control plan hosts by default
> 
> This isn't - actually in general what we care about here is the service logs on a failing node. We've already changed the MCD to include these logs in its pod logs in some cases, but it's clearly more robust if we just gather it directly.
> 
> Perhaps we could change the collection logic here to look especially at nodes which are listed as degraded by the MCO?

We could; but I think that would mean a new script and new defaults (in the gather script), so if we go this way it means a VASTLY different PR.

cgwalters:
OK.  This PR won't do what I need; may try to look at a PR like that at some point if no one else does it.

cgwalters:
OK ultimately I think https://github.com/openshift/machine-config-operator/pull/2587 will fix this because we'll stop throwing away `stderr` from the rpm-ostree process and get those logs in the MCD pod logs as well as propagate it up into the clusteroperator status.

PR #249, opened 2021-07-19
Added NodeSelection functionality

bpickard22:
Added NodeSelection functionality to gather_network_logs

Users may now pass in nodes that exist on the cluster to the gather and
it will only run on those nodes.
User input is verified against nodes that exist on the cluster
To use run $oc adm must-gather -- gather_network_logs <node/s> 
Signed-off-by: Ben Pickard <bpickard@redhat.com>

### DIFF
--- a/collection-scripts/gather_network_logs
+++ b/collection-scripts/gather_network_logs
@@ -65,7 +65,7 @@ function gather_ovn_ipsec_data {
 
 function gather_openshiftsdn_nodes_data {
   echo "INFO: Gathering Openshift-SDN data"
-  for NODE in ${NODES}; do
+  for NODE in "${NODES[@]}"; do
       SDN_POD=$(oc -n openshift-sdn get pods --no-headers -o custom-columns=":metadata.name" --field-selector spec.nodeName="${NODE}" -l app=sdn)
       oc -n openshift-sdn exec "${SDN_POD}" -c sdn -- bash -c "iptables-save -c" > "${NETWORK_LOG_PATH}"/"${NODE}"_iptables &
       PIDS+=($!)
@@ -81,7 +81,7 @@ function gather_openshiftsdn_nodes_data {
 
 function gather_ovn_kubernetes_nodes_data {
   echo "INFO: Gathering ovn-kubernetes node data"
-  for NODE in ${NODES}; do
+  for NODE in "${NODES[@]}"; do
       OVNKUBE_NODE_POD=$(oc -n openshift-ovn-kubernetes get pods --no-headers -o custom-columns=":metadata.name" --field-selector spec.nodeName="${NODE}" -l app=ovnkube-node)
       oc -n openshift-ovn-kubernetes exec -c ovnkube-node "${OVNKUBE_NODE_POD}" -- bash -c "iptables-save -c" > "${NETWORK_LOG_PATH}"/"${NODE}"_"${OVNKUBE_NODE_POD}"_iptables &
       PIDS+=($!)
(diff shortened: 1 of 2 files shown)


jluhrsen:
nothing stood out to me looking at the code.
Is there any example using this in some doc or maybe we don't need that?

bpickard22:
> nothing stood out to me looking at the code.
> Is there any example using this in some doc or maybe we don't need that?
@jluhrsen 
added use command to comment of pr. Is that a good place for it or would you want a document with some examples?

jluhrsen:
> > nothing stood out to me looking at the code.
> > Is there any example using this in some doc or maybe we don't need that?
> > @jluhrsen
> > added use command to comment of pr. Is that a good place for it or would you want a document with some examples?

yeah sure. I was expecting some other place with some docs about how to use different gather commands, but I'm not finding anything where you'd add this to.

collection-scripts/gather_network_logs
```diff
@@ -228,21 +228,41 @@ function gather_kuryr_data {
...
+TMP_ARR=() #tmp array to place the args that are nodes when they are checked against the cluster nodes
+if [ $# -eq 0 ]; then 
     echo "WARNING: Collecting network logs on ALL linux nodes in your cluster. This could take a long time." >&2
+else
+    for (( i=0; i<ELEMENTS; i++));do
+      for (( j=0; j<NODE_LENGTH; j++));do
+        if [[ "${ARGS[$i]}" == "${NODES[$j]}" ]]; then
+        echo  ${ARGS[$i]} "->node exists in nodes"
+        TMP_ARR[${#TMP_ARR[@]}]+="${NODES[$j]}"
+        fi
```
flavio-fernandes, line 253: NIT: should we have a warning for node names that were provided by caller and are not found from "oc get nodes"?

bpickard22: probably would be a good thing to add

astoycos: +1 here

collection-scripts/gather_network_logs
```diff
@@ -228,21 +228,41 @@ function gather_kuryr_data {
...
+NODE_LENGTH=${#NODES[@]}
+TMP_ARR=() #tmp array to place the args that are nodes when they are checked against the cluster nodes
+if [ $# -eq 0 ]; then 
     echo "WARNING: Collecting network logs on ALL linux nodes in your cluster. This could take a long time." >&2
+else
+    for (( i=0; i<ELEMENTS; i++));do
+      for (( j=0; j<NODE_LENGTH; j++));do
+        if [[ "${ARGS[$i]}" == "${NODES[$j]}" ]]; then
+        echo  ${ARGS[$i]} "->node exists in nodes"
+        TMP_ARR[${#TMP_ARR[@]}]+="${NODES[$j]}"
```
astoycos, line 252: Why do you need the TMP_ARR array here? Just remove it from the `ARGS` array if it doesn't exist in `NODES
`

sferich888:
@bpickard22 have you had a chance to checkup on this?

collection-scripts/gather_network_logs
```diff
@@ -228,21 +228,41 @@ function gather_kuryr_data {
 
 function gather_kuryr_nodes_data {
   echo "INFO: Gathering kuryr nodes data"
-  for NODE in ${NODES}; do
-      CNI_POD=$(oc -n openshift-kuryr get pods --no-headers -o custom-columns=":metadata.name" --field-selector spec.nodeName="${NODE}" -l app=kuryr-cni)
+  for NODE in "${NODES[@]}"; do
+      CNI_POD=($(oc -n openshift-kuryr get pods --no-headers -o custom-columns=":metadata.name" --field-selector spec.nodeName="${NODE}" -l app=kuryr-cni))
```
RickJWagner, line 232: What difference is the extra paragraph making?  
CNI_POD=($(oc -n           << instead of  CNI_POD=$(oc -n
not criticizing, just wondering why this change is necessary.

collection-scripts/gather_network_logs
```diff
@@ -228,21 +228,41 @@ function gather_kuryr_data {
...
           nsenter --net=/host_proc/\$pid/ns/net ip -d addr; \
           done" > "${NETWORK_LOG_PATH}"/"${CNI_POD}"_interfaces & PIDS+=($!)
   done
 }
-
-if [ $# -eq 0 ]; then
+ 
+ARGS=($@) #args passed in by the user
+IFS=" " read -r -a NODES <<< $(oc get nodes -o jsonpath='{range .items[*]}{@.metadata.name} {.status.nodeInfo.operatingSystem==linux}')
+ELEMENTS=${#ARGS[@]} #length of args passed in by the user
```
RickJWagner, line 242: Thanks for leaving the explanatory comments, this makes reviews easier and will be helpful to future readers, too.

RickJWagner (commented):
Left a few comments in-line with code.

bpickard22:
closing this pr due to the deprecation of gather_network_logs. If we decide that we want to revisit this idea in must_gather after the refactor I can take care of it there

PR #251, opened 2021-09-03
introduces a script for collecting kas static pod logs

p0lyn0mial:
the logs are optional and stored at `/var/logs/kube-apiserver/startup.log` and `/var/logs/kube-apiserver/termination.log` on the host.

the script will save it under `must-gather/static-pods/kube-apiserver`

### DIFF
--- a/collection-scripts/gather_kas_startup_termination_logs
+++ b/collection-scripts/gather_kas_startup_termination_logs
@@ -1,0 +1,32 @@
+#!/bin/bash
+# Downloads the startup.log and termination.log (and its rotated copies) from /var/logs/kube-apiserver on each master node
+
+set -o nounset
+set -o errexit
+set -o pipefail
+
+BASE_COLLECTION_PATH="${BASE_COLLECTION_PATH:-/must-gather}"
+
+echo "WARNING: Collecting one or more kube-apiserver related logs on ALL masters in your cluster. This could take a large amount of time." >&2
+
+# the command executed by xargs below expects four parameters:
+# $1 - node path under /var/logs to download
+# $2 - local output path
+# $3 - node name
+# $4 - log file name
+paths=(kube-apiserver)
+for path in "${paths[@]}" ; do
+  output_dir="${BASE_COLLECTION_PATH}/static-pods/$path"
+  mkdir -p "$output_dir"
+  oc adm node-logs --role=master --path="$path" | \
+  grep -E '(startup.*.log|termination.log)' | \
+  sed "s|^|$path $output_dir |"
+done | \
+xargs --max-args=4 --max-procs=45 --no-run-if-empty bash -c \
+  'echo "INFO: Started  downloading $1/$4 from $3";
+  oc adm node-logs $3 --path=$1/$4 | gzip > $2/$3-$4.gz;
+  echo "INFO: Finished downloading $1/$4 from $3"' \
+  bash
+
+# force disk flush to ensure that all data gathered is accessible in the copy container
+sync
--- a/collection-scripts/gather
+++ b/collection-scripts/gather
@@ -70,5 +70,8 @@ oc adm inspect --dest-dir must-gather "${group_resources_text}"
 # Gather HAProxy config files
 /usr/bin/gather_haproxy_config
 
+# Gather kas startup and termination logs
+/usr/bin/gather_kas_startup_termination_logs
+
 # force disk flush to ensure that all data gathered is accessible in the copy container
 sync


collection-scripts/gather_static_pod_monitor_logs
```diff
@@ -0,0 +1,30 @@
...
+echo "WARNING: Collecting one or more startup monitor logs on ALL masters in your cluster. This could take a large amount of time." >&2
+
+# the command executed by xargs below expects four parameters:
+# $1 - node path under /var/logs to download
+# $2 - local output path
+# $3 - node name
+# $4 - log file name
+paths=(kube-apiserver)
+for path in "${paths[@]}" ; do
+  output_dir="${BASE_COLLECTION_PATH}/startup_monitor_logs/$path"
```
sttts, line 19: would rather generalize this to `staticpods/kube-apiserver/{startup,termination}.log`

collection-scripts/gather_static_pod_monitor_logs
```diff
@@ -0,0 +1,30 @@
...
+  mkdir -p "$output_dir"
+  oc adm node-logs --role=master --path="$path" | \
+  grep "startup.log" | \
```
sttts, line 22: also the termination.logs

p0lyn0mial: done

collection-scripts/gather_static_pod_monitor_logs
p0lyn0mial: it is easier to copy to one file, at the moment it will be `must-gather/kube-apiserver/`

p0lyn0mial: we could change it to `must-gather/kube-apiserver/lifecycle_logs`

p0lyn0mial:
maybe we should store `terimination` logs under `namespaces/openshift-kube-apiserver/pod/kube-apiserver-ip-10-0-145-61.us-west-2.compute.internal/kube-apiserver/kube-apiserver`

and `startup.log` in `namespaces/openshift-kube-apiserver/pod/kube-apiserver-ip-10-0-145-61.us-west-2.compute.internal/kube-apiserver/startup-monitor`

collection-scripts/gather_kas_startup_termination_logs
```diff
@@ -0,0 +1,32 @@
...
+for path in "${paths[@]}" ; do
+  output_dir="${BASE_COLLECTION_PATH}/static-pods/$path"
+  mkdir -p "$output_dir"
+  oc adm node-logs --role=master --path="$path" | \
+  grep -E '(startup.*.log|termination.log)' | \
+  sed "s|^|$path $output_dir |"
+done | \
+xargs --max-args=4 --max-procs=45 --no-run-if-empty bash -c \
+  'echo "INFO: Started  downloading $1/$4 from $3";
+  oc adm node-logs $3 --path=$1/$4 | gzip > $2/$3-$4.gz;
```
sferich888, line 27: @mfojtik do users need to do this (compression)? Isn't must-gather now compressing the whole archive when we collect it? 

I fear this complicates reading the archive once initially extracted; for automated tools; like insights, etc

sferich888: I know we use/have this convention with; https://github.com/openshift/must-gather/blob/master/collection-scripts/gather_audit_logs#L45-L50, but audit logs aren't collected by default; thus the format these are 'retained' in doesn't matter as much. 

@p0lyn0mial can we just save the raw files to the directory; vs doing this complicated archive?

collection-scripts/gather_kas_startup_termination_logs
```diff
@@ -0,0 +1,32 @@
...
+# the command executed by xargs below expects four parameters:
+# $1 - node path under /var/logs to download
+# $2 - local output path
+# $3 - node name
+# $4 - log file name
+paths=(kube-apiserver)
+for path in "${paths[@]}" ; do
+  output_dir="${BASE_COLLECTION_PATH}/static-pods/$path"
+  mkdir -p "$output_dir"
+  oc adm node-logs --role=master --path="$path" | \
```
sferich888, line 21: @p0lyn0mial please add a note pointing at https://github.com/openshift/library-go/pull/1196/files#diff-387eea59f06b61f1bf3cce6c2fd89f63150b60bfee719c53efa885a87bbc5be9R180 that explains how much data this should be collecting. 

@sttts and I discussed this on slack; thus it should be ~3 pod logs * ~3 (MaxFiles) * ~100 KB +/- (a bit depending on the math). 

I simply want to capture how much we expected this to retain.

sferich888 (approved):

PR #256, opened 2021-10-08
Bug 2012326: add script for collection debug information specific to IBM Z systems

Prashanth684:
This is to address https://bugzilla.redhat.com/show_bug.cgi?id=2012326. Today as part of the docs for IBM Z installation [1], we ask users to run a script to collect debug information[2] by doing an oc debug node. There should be no need for this as this should be part of the must-gather command.

[1] https://access.redhat.com/documentation/en-us/openshift_container_platform/4.8/html/installing/installing-with-rhel-kvm-on-ibm-z-and-linuxone
[2] https://github.com/ibm-s390-linux/s390-tools/blob/master/scripts/dbginfo.sh

### DIFF
--- a/collection-scripts/gather_dbginfo_z
+++ b/collection-scripts/gather_dbginfo_z
@@ -1,0 +1,56 @@
+#!/bin/bash
+# dbginfo.sh is a script shipped with the s390-utils package on IBM-Z systems that collects system level information for debugging
+BASE_COLLECTION_PATH="must-gather"
+DBGINFO_PATH=${OUT:-"${BASE_COLLECTION_PATH}/node_dbginfo"}
+
+mkdir -p "${DBGINFO_PATH}"/
+
+function get_dbginfo_off_node {
+    local debugPod=""
+
+    #Get debug pod's name
+    debugPod=$(oc debug --to-namespace="default" node/"$1" -o jsonpath='{.metadata.name}')
+
+    #Start Debug pod, execute the dbginfo command to collect the system data and force it to stay up until removed in "default" namespace
+    oc debug --to-namespace="default" node/"$1" -- chroot /host /bin/bash -c 'mkdir -p /tmp/dbginfo && dbginfo.sh -d /tmp/dbginfo && sleep 300' > /dev/null 2>&1 &
+
+    #Mimic a normal oc call, i.e pause between two successive calls to allow pod to register
+    #Wait a minute for the dbginfo script to collect the information
+    sleep 60
+    oc wait -n "default" --for=condition=Ready pod/"$debugPod" --timeout=30s
+
+    if [ -z "$debugPod" ]
+    then
+      echo "Debug pod for node ""$1"" never activated"
+    else
+      #Copy dbginfo out of Nodes suppress Stdout
+      echo "Copying dbginfo on node ""$1"""
+      oc cp  --loglevel 1 -n "default" "$debugPod":/host/tmp/dbginfo/ "${DBGINFO_PATH}"/ > /dev/null 2>&1 && PIDS+=($!)
+
+      #clean up debug pod after we are done using them
+      oc delete pod "$debugPod" -n "default"
+    fi
+}
+
+function gather_dbginfo_data {
+  #Run dbginfo pull function on all nodes in parallel
+  for NODE in ${NODES}; do
+    get_dbginfo_off_node "${NODE}" &
+  done
+}
+
+if [ $# -eq 0 ]; then
+    echo "WARNING: Collecting dbginfo on ALL linux nodes in your cluster."
+fi
+
+PIDS=()
+NODES="${*:-$(oc get nodes --selector=node-role.kubernetes.io/master -o jsonpath='{range .items[*]}{@.metadata.name} {.status.nodeInfo.operatingSystem==linux}')}"
+
+gather_dbginfo_data
+
+echo "INFO: Waiting for node dbginfo collection to complete ..."
+wait "${PIDS[@]}"
+echo "INFO: Node dbginfo collection to complete."
+
+# force disk flush to ensure that all data gathered is accessible in the copy container
+sync
--- a/collection-scripts/gather
+++ b/collection-scripts/gather
@@ -82,5 +82,10 @@ oc adm inspect --dest-dir must-gather --rotated-pod-logs "${group_resources_text
 # Gather metallb logs
 /usr/bin/gather_metallb_logs
 
+arch=$(uname -m)
+if [ $arch == "s390x" ]; then
+    /usr/bin/gather_dbginfo_z
+fi
+
 # force disk flush to ensure that all data gathered is accessible in the copy container
 sync


collection-scripts/gather_dbginfo_z
```diff
@@ -0,0 +1,56 @@
...
+mkdir -p "${DBGINFO_PATH}"/
+
+function get_dbginfo_off_node {
+    local debugPod=""
+
+    #Get debug pod's name
+    debugPod=$(oc debug --to-namespace="default" node/"$1" -o jsonpath='{.metadata.name}')
+
+    #Start Debug pod, execute the dbginfo command to collect the system data and force it to stay up until removed in "default" namespace
+    oc debug --to-namespace="default" node/"$1" -- chroot /host /bin/bash -c 'mkdir -p /tmp/dbginfo && dbginfo.sh -d /tmp/dbginfo && sleep 300' > /dev/null 2>&1 &
```
sferich888, line 15: Is the sleep going to be enough time here to start all these processes and wait for them to complete? 300 seconds is 5 min (thats a long time). 

What happens when the debug script finishes early? Do we just have to wait? 

A better approach here (IMO) is to use a 'signal' to say 'collection is done', then exfiltrate the data you want. 

Consider the following:  

```yaml
apiVersion: v1
kind: Pod
metadata:
  name: sos-collector
  labels:
    name: sosreport-container
spec:
  hostNetwork: true
  nodeName: "NODE NAME"   ## TODO: Template or Change
  hostPID: true
  restartPolicy: Never
  priorityClassName: "system-cluster-critical"
  tolerations:
  - key: node-role.kubernetes.io/master
    operator: "Equal"
    effect: "NoExecute"
    tolerationSeconds: 3600
  - key: node-role.kubernetes.io/master
    operator: "Equal"
    effect: "NoSchedule"
  containers:
  - name: sos-extractor
    image: registry.redhat.io/rhel8/support-tools
    command: ["/bin/bash", "-c", "trap : TERM INT; sleep infinity & wait"]
    volumeMounts:
    - mountPath: /tmp/sos-data
      name: sos-data
  initContainers:
  - name: sos-collector
    image: registry.redhat.io/rhel8/support-tools
    command: ["sosreport", "-s", "/host", "--tmp-dir", "/tmp/sos-data", "-c", "always", "--batch", "-v"]
    securityContext:
      runAsUser: 0
      privileged: True
    volumeMounts:
    - mountPath: /host
      name: host-volume
    - mountPath: /tmp/sos-data
      name: sos-data
  volumes:
  - name: host-volume
    hostPath:
      path: /
      type: Directory
  - name: sos-data
    emptyDir: {}
```
The cli execution then looks like the following. 
```
oc create -f sos_pod.yaml
oc wait --for=condition=Ready pod/sos-collector --timeout 10m && oc cp -c sos-extractor sos-collector:/tmp/sos-data/ .
```

What this does is it 'starts a collection container' and only if that completes it 'starts an exfiltration container' that the `oc wait` (can pick up the signal - start of the pod) and `oc cp/rsync` command can continue driving the script copying data as fast as possible (with out the need for long sleeps or delays).

collection-scripts/gather_dbginfo_z
```diff
@@ -0,0 +1,56 @@
...
+
+      #clean up debug pod after we are done using them
+      oc delete pod "$debugPod" -n "default"
+    fi
+}
+
+function gather_dbginfo_data {
+  #Run dbginfo pull function on all nodes in parallel
+  for NODE in ${NODES}; do
+    get_dbginfo_off_node "${NODE}" &
```
sferich888, line 38: Save this to the pid of this to the PIDS array.

collection-scripts/gather_dbginfo_z
```diff
@@ -0,0 +1,56 @@
...
+BASE_COLLECTION_PATH="must-gather"
+DBGINFO_PATH=${OUT:-"${BASE_COLLECTION_PATH}/node_dbginfo"}
+
+mkdir -p "${DBGINFO_PATH}"/
+
+function get_dbginfo_off_node {
+    local debugPod=""
+
+    #Get debug pod's name
+    debugPod=$(oc debug --to-namespace="default" node/"$1" -o jsonpath='{.metadata.name}')
```
sferich888, line 12: Don't push this pod to the default namespace, must-gathers already run in an isolated namespace; doing this could break (escape) our cleanup processes.

sferich888: You do this below too (multiple places to correct).

collection-scripts/gather_dbginfo_z
```diff
@@ -0,0 +1,56 @@
...
+
+    #Start Debug pod, execute the dbginfo command to collect the system data and force it to stay up until removed in "default" namespace
+    oc debug --to-namespace="default" node/"$1" -- chroot /host /bin/bash -c 'mkdir -p /tmp/dbginfo && dbginfo.sh -d /tmp/dbginfo && sleep 300' > /dev/null 2>&1 &
+
+    #Mimic a normal oc call, i.e pause between two successive calls to allow pod to register
+    #Wait a minute for the dbginfo script to collect the information
+    sleep 60
+    oc wait -n "default" --for=condition=Ready pod/"$debugPod" --timeout=30s
+
+    if [ -z "$debugPod" ]
```
sferich888, line 22: I wonder about this check; what your checking for is that you define a pod name not that the pod is up(checked for by the wait command). 

Should we reconsider how this is done?

sferich888 (changes requested):
I found a few more thing I would like to see changed.

collection-scripts/gather_dbginfo_z
RickJWagner: I agree with @sferich888 on this one.  Waiting a fixed amount of time seems fragile.
- We might have a slow response for some reason.
- On the other hand, perhaps time will go by and 500 seconds will be way too long (because of resource improvements, for example).
- Finding some way to 'signal' that the work is done would be better, if we can find a good way to do it.

collection-scripts/gather_dbginfo_z
RickJWagner: Sleeping for some fixed amount of time can be problematic.  
Slow/resource-taxed environments will have to be the 'high water mark', every other environment will be hobbled by having to wait as long as the slowest cluster requires.  This isn't optimal.
If another way is possible, please consider it.

RickJWagner (commented):
Left review comments as additions to Eric's.

collection-scripts/gather_dbginfo_z
Prashanth684: yes a signal is the cleanest way. is there a way to do it through oc rather than the approach outlined above which is having it as an init container? this would require a container image.

collection-scripts/gather_dbginfo_z
Prashanth684: will correct all the places.

collection-scripts/gather_dbginfo_z
Prashanth684: i could remove the timeout and check if the condition is met.

Prashanth684: i am not sure how to avoid the sleep here or the timeout. we would need one or the other to give the pod some time.

collection-scripts/gather_dbginfo_z
Prashanth684: i believe that is done here: https://github.com/openshift/must-gather/pull/256/files#diff-d1363d2abd16572145a0357dbb642ef1a7a8b5a9087559c62dfb92e94741eaa4R28 ?

collection-scripts/gather_dbginfo_z
```diff
@@ -0,0 +1,56 @@
+#!/bin/bash
+# dbginfo.sh is a script shipped with the s390-utils package on IBM-Z systems that collects system level information for debugging
+BASE_COLLECTION_PATH="must-gather"
+DBGINFO_PATH=${OUT:-"${BASE_COLLECTION_PATH}/node_dbginfo"}
+
+mkdir -p "${DBGINFO_PATH}"/
+
+function get_dbginfo_off_node {
```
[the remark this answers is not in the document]
soltysh: @sferich888 you mean the for-range over the nodes? I agree that this has multiple potential fallouts, both scale-wise (with multiple nodes) and stability-wise (when some nodes are down). Given our recent discussions how node sos-reports are being gathered, wouldn't be more beneficial to push towards that direction for per-node information gathering?

sferich888:
@Prashanth684 can you update this PR or progress it so we can resolve the connected bug?

Prashanth684:
> @Prashanth684 can you update this PR or progress it so we can resolve the connected bug?

there was a recent package change in the s390utils repo which caused this script to be removed . i am working with the RHEL team to get the command added back.

In the meanwhile, from your comments what i understand is that we want to do this the way you mention here: https://github.com/openshift/must-gather/pull/256#discussion_r739509509. i can do this today the same way that you mention for sos-report and it works fine. is there a way to get this automated through must-gather?

sferich888:
@Prashanth684 as we discussed we want to close this in favor of using tools/plugins with sos-4.2 to collect this information. 

/close

PR #257, opened 2021-10-15
Adding soltysh as approver

soltysh:
/assign @sttts @mfojtik


mfojtik:
Adding @soltysh as an emergency approver in case we need to merge a PR and Eric is not around.

Issue #255, opened 2021-09-29
Include insights-operator archive in must-gather


iNecas:
While insights-operator archive is significantly smaller than must-gather, it manages to collect data relevant for the cluster that are not collected by must-gather.

Having insights-operator as part of the must-gather would expand the data available for troubleshooting, either to be used directly, or by the recommendation engine for the support.

sferich888:
@iNecas I added a PR for this; I will want @davemulford to review it; as I am worried that this method may need to be adjusted given the amount of data we can pull back doing this.

PR #263, opened 2021-10-29
Adding automated insights archive collection

sferich888:
This should resolve https://github.com/openshift/must-gather/issues/255

cc: @iNecas

### DIFF
--- a/collection-scripts/gather
+++ b/collection-scripts/gather
@@ -45,6 +45,9 @@ oc adm inspect --dest-dir must-gather "${named_resources[@]}"
 group_resources_text=$(IFS=, ; echo "${group_resources[*]}")
 oc adm inspect --dest-dir must-gather "${group_resources_text}"
 
+# Gather Insights Operator Archives
+/user/bin/gather_insights
+
 # Gather monitoring data from the cluster
 /usr/bin/gather_monitoring
 
--- a/collection-scripts/gather_insights
+++ b/collection-scripts/gather_insights
@@ -1,0 +1,14 @@
+#!/bin/bash
+# Gather the insights archive data from insights operator
+
+BASE_COLLECTION_PATH="${BASE_COLLECTION_PATH:-/must-gather}"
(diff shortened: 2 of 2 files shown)


iNecas:
`/var/lib/insights-operator` keeps some history of archives in there: I guess we would be fine keeping just the last one of it, and perhaps considering unpacking it, so that it would be ready to be used once the must-gather is decompressed.

sferich888:
@iNecas i guess I just need to know what the CCX team wants? To me it's odd to collect this when this is already streamed back to Red Hat.?.?

iNecas:
The reasons:
1. facing reality, all in once is easier than combining the data at end-user level when troubleshooting
2. insights-operator might be disabled to send data, but it's still generating tarballs - allows having same structure of data even for disconnected clusters
3. despite must-gather collecting larger dataset, it's not collecting everything insights-operator does: this could reduce the efforts done on enhancing the data collection - we wouldn't have to update must-gather if insights-operator already collected the data (and the reasons lay in (1) or (2))
4. would make must-gather truly superset of what insights-operator gathers - would allow not having to have two separate conditions in insights rules

The reasons are mostly pragmatic. And given the IO payload is minimal compared to what must-gather has, I'm not seeing that much of downsides here. But if you feel strongly against this, I could live with that.

sferich888:
I have no issues collecting the data; it just wasn't clear how it would be used. 

- Note: need to raise / discuss with @soltysh flags/options to 'oc rsync' that would let us do the filtering of files  (only get latest, etc) vs having to make this multiple calls with 'oc'.

RickJWagner (approved):
Looks good to me.  One item stands out, it's only a problem if it's unintended.  Please make sure the dash (-) is intended in BASE_COLLECTION_PATH.  
-/must-gather          (do we really want the leading dash?)

sferich888:
@RickJWagner that is a bash syntax for defaults ':-' is the separation identifier.

PR #267, opened 2021-11-10
Add bare metal networking logs collection script.

npinaeva:
Since bare metal logs are required from all nodes at the same time, new mg script fits better than collecting sosreports from every node.

Signed-off-by: Nadia Pinaeva <npinaeva@redhat.com>

### DIFF
--- a/collection-scripts/gather_bm_network_logs
+++ b/collection-scripts/gather_bm_network_logs
@@ -1,0 +1,61 @@
+#!/bin/bash
+. gather_service_logs_util
+BASE_COLLECTION_PATH="must-gather"
+NETWORK_LOG_PATH=${OUT:-"${BASE_COLLECTION_PATH}/bm_network_logs"}
+SERVICE_LOG_PATH="${NETWORK_LOG_PATH}"
+
+mkdir -p "${NETWORK_LOG_PATH}"/
+
+function gather_bm_network_logs {
+  NODE=${1}
+  mkdir -p "${NETWORK_LOG_PATH}/${NODE}"
+  HOST_POD=$(oc debug --to-namespace="${MG_NAMESPACE}" node/"${NODE}" -o jsonpath='{.metadata.name}')
+  oc debug --to-namespace="${MG_NAMESPACE}" node/"${NODE}" -- /bin/bash -c \
+    'trap : TERM INT; sleep infinity & wait' > /dev/null 2>&1 &
+
+  # wait for pod creation
+  sleep 1
+
+  oc wait -n="${MG_NAMESPACE}" --for=condition=Ready pod/"${HOST_POD}" --timeout=60s
+  if [ $? -ne 0 ]; then
+    echo "ERROR: Timeout waiting for pod to become Ready"
+    oc delete pod -n="${MG_NAMESPACE}" "${HOST_POD}"
+    return
+  fi
+
+  oc exec -n="${MG_NAMESPACE}" "${HOST_POD}" -- bash -c "ip addr" > "${NETWORK_LOG_PATH}/${NODE}/ip-addr.txt"
+  oc exec -n="${MG_NAMESPACE}" "${HOST_POD}" -- bash -c "ip route" > "${NETWORK_LOG_PATH}/${NODE}/ip-route.txt"
+  oc exec -n="${MG_NAMESPACE}" "${HOST_POD}" -- bash -c "hostname" > "${NETWORK_LOG_PATH}/${NODE}/hostname.txt"
+  HOST_COMMAND='for c in $(nmcli -f uuid con show | tail -n +2); do nmcli con show $c; done'
+  oc exec -n="${MG_NAMESPACE}" "${HOST_POD}" -- bash -c \
+    "chroot /host /bin/bash -c '${HOST_COMMAND}'" \ > "${NETWORK_LOG_PATH}/${NODE}/nmcli.txt"
+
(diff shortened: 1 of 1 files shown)


collection-scripts/gather_bm_network_logs
```diff
@@ -0,0 +1,64 @@
...
+
+  # wait for pod creation
+  sleep 1
+
+  for (( i=0;i<${#NODES[@]};i++ )); do
+    NODE=${NODES[i]}
+    HOST_POD=${HOST_PODS[i]}
+    oc wait -n="${MG_NAMESPACE}" --for=condition=Ready pod/"${HOST_POD}" --timeout=30s
+    if [ $? -ne 0 ]; then
+      echo "ERROR: Timeout waiting for pod to become Ready"
```
sferich888, line 37: In addition to logging this error save this to a file in the archive, so when looking at the archive you can quickly see where you did/didn't collect logs from.

collection-scripts/gather_bm_network_logs
```diff
@@ -0,0 +1,64 @@
...
+  done
+}
+
+PIDS=()
+NODES=(${@:-$(oc get nodes -o jsonpath='{.items[?(@.status.nodeInfo.operatingSystem=="linux")].metadata.name}')})
+
+echo "WARNING: Collecting network logs on ALL linux nodes in your cluster. This could take a long time."
+gather_bm_network_logs
+echo "INFO: BM network log collection complete."
+wait "${PIDS[@]}"
```
sferich888, line 62: We don't take advantage of this and I think, that's an issue (for speed/threading). 

gather_bm_network_logs should be a function that is node specific (IMO) and you should subprocess it's invocation and background it so that you can thread the collection to multiple nodes at once vs it being done in serial (the way it is happening now). 

To do this it will take a bit of rework but the changes are likely small/manageable. Can this be considered?

sferich888 (changes requested):
What we have is fine (personally I hate the the 'oc [debug|exec] && oc wait' practice, but I understand why your doing it). 

That said this is going to be SLOW (as it is now) for any set of nodes let alone a large set of nodes(ie:200), so we should try and improve the way we collect.

collection-scripts/gather_bm_network_logs
npinaeva: Empty <node_name> folder should give a hint
I can add erros.txt file like in sosreport to the archive root, or replace empty folder with <node_name.error> or put error file inside a folder. I don't think we have a common way to log errors in must-gather, wdyt?

collection-scripts/gather_bm_network_logs
npinaeva: Absolutely!

collection-scripts/gather_bm_network_logs
```diff
@@ -0,0 +1,61 @@
...
+  # POD_NAME env variable is passed to must-gather pods since 4.10 version,
+  # if is not defined, command will return the first line of 'get pods' output, which is column name='ns'
+  echo "ERROR: Collection failed, use the latest oc version for it to succeed"
+  exit 1
+fi
+
+for NODE in ${NODES[@]}; do
+  gather_bm_network_logs ${NODE} & PIDS+=($!)
+  # collect_service_logs adds PID to PIDS
+  collect_service_logs "${NODE}" NetworkManager
```
sferich888, line 55: Thread this with the PIDS structure too; if you don't this becomes serial vs threaded.

collection-scripts/gather_bm_network_logs
```diff
@@ -0,0 +1,61 @@
...
+
+  oc cp -n="${MG_NAMESPACE}" "${HOST_POD}":/host/etc/resolv.conf "${NETWORK_LOG_PATH}/${NODE}/resolv.conf"
+
+  oc delete pod -n="${MG_NAMESPACE}" "${HOST_POD}"
+}
+
+PIDS=()
+NODES=(${@:-$(oc get nodes -o jsonpath='{.items[?(@.status.nodeInfo.operatingSystem=="linux")].metadata.name}')})
+
+echo "WARNING: Collecting network logs on ALL linux nodes in your cluster. This could take a long time."
```
sferich888, line 41: Should this only be shown with default argument are used?

collection-scripts/gather_bm_network_logs
```diff
@@ -0,0 +1,61 @@
...
+  # wait for pod creation
+  sleep 1
+
+  oc wait -n="${MG_NAMESPACE}" --for=condition=Ready pod/"${HOST_POD}" --timeout=60s
+  if [ $? -ne 0 ]; then
+    echo "ERROR: Timeout waiting for pod to become Ready"
+    oc delete pod -n="${MG_NAMESPACE}" "${HOST_POD}"
+    return
+  fi
+
```
sferich888, line 25: We should comment that all of this can be / is collectable via sosreport.

collection-scripts/gather_bm_network_logs
npinaeva: It's done in `collect_service_logs` https://github.com/openshift/must-gather/blob/master/collection-scripts/gather_service_logs_util#L31-L32

collection-scripts/gather_bm_network_logs
npinaeva: Not sure I get it, what kind of default arguments you mean?

sferich888: When no node list is provided.

npinaeva:
Decided to collect this information using `sos collect` https://github.com/sosreport/sos/wiki instead of must-gather

PR #264, opened 2021-11-09
Add network logs gathering by default

npinaeva:
Remove "for every node" collection loops from gather_network_logs,
this information is supposed to be collected on a per-node basis
(e.g. using sosreport).
Remove duplicating with /namespaces logs collection.
Add cluster scale information (services, endpoints, pods amount).
Add gather_network_logs to default gather script.
Remove stale gather_network_ovn_trace script.

Signed-off-by: Nadia Pinaeva <npinaeva@redhat.com>

PR #269, opened 2021-11-11
Collect ip route cache data

mccv1r0:
Collect ip route cache data

Signed-off-by: Michael Cambria <mcambria@redhat.com>


mccv1r0:
Fixes https://github.com/openshift/must-gather/issues/270

sferich888:
See #264 I think it's going to collide with this.

PR #273, opened 2021-12-01
Add script to collect kubelet profiling data

fgiudici:
Allow to collect kubelet profiling data when required:
oc adm must-gather -- gather_profiling_kubelet $NODE

This would help diagnosing issues in the kubelet.
(https://issues.redhat.com/browse/OCPNODE-698)

### DIFF
--- a/collection-scripts/gather_profiling_node
+++ b/collection-scripts/gather_profiling_node
@@ -1,0 +1,70 @@
+#!/bin/bash
+
+BASE_COLLECTION_PATH="${BASE_COLLECTION_PATH:-must-gather}"
+PROFILING_NODE_LOG_PATH="${BASE_COLLECTION_PATH}/pprof"
+PROFILING_NODE_KUBELET_LOG_PATH="${PROFILING_NODE_LOG_PATH}/kubelet"
+
+
+PROFILING_NODE_APISERVER="kubernetes.default.svc"
+PROFILING_NODE_SECONDS="${PROFILING_NODE_SECONDS:-30}"
+
+# _gather_profiling_node_node_exists $NODE
+# returns "true" if $NODE is found among the cluster nodes, "false" otherwise.
+_gather_profiling_node_node_exists() {
+    local node="$1"
+
+    oc get node "$node" > /dev/null 2>&1
+    return "$?"
+}
+
+# _gather_profiling_node_is_apiserver_available
+# returns "true" if the kubernetes apiserver can be reached, returns "false" otherwise.
+_gather_profiling_node_is_apiserver_available() {
+    oc get --raw https://${PROFILING_NODE_APISERVER}/api > /dev/null 2>&1
+    return "$?"
+}
+
+_gather_profiling_node_collect_kubelet_data() {
+    local node="$1"
+    local out_kub_base="${PROFILING_NODE_KUBELET_LOG_PATH}/${node}"
+    local out_kub_heap="${out_kub_base}_heap.out"
+    local out_kub_prof="${out_kub_base}_prof.out"
+    local url_base="https://${PROFILING_NODE_APISERVER}/api/v1/nodes/${node}/proxy/debug/pprof"
+
+    oc get --raw "${url_base}/profile?seconds=${PROFILING_NODE_SECONDS}" > "$out_kub_prof"
+    oc get --raw "${url_base}/heap" > "$out_kub_heap"
+}
+
+_gather_profiling_node_collect_data_loop() {
+    local pids=()
+    local node
+
+    mkdir -p "$PROFILING_NODE_KUBELET_LOG_PATH"
+    for node in ${NODES}; do
+        if _gather_profiling_node_node_exists "$node"; then
+            echo "INFO: start kubelet profiling task on node ${node}"
+            _gather_profiling_node_collect_kubelet_data "${node}" & pids+=($!)
+        else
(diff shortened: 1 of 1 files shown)


collection-scripts/gather_profiling_kubelet
```diff
@@ -0,0 +1,83 @@
...
+        fi
+    done
+
+    [ -z "$pids" ] && return
+    echo "INFO: wait for kubelet profiling tasks"
+    wait ${pids[@]}
+}
+
+
+NODES="${@:-$(oc get nodes -o jsonpath='{.items[?(@.status.nodeInfo.operatingSystem=="linux")].metadata.name}')}"
```
sferich888, line 73: If we pass in a windows node will this still work? If yes, then why limit to 'linux' nodes by default?

collection-scripts/gather_profiling_kubelet
```diff
@@ -0,0 +1,83 @@
...
+
+echo "INFO: GATHER KUBELET PROFILING DATA"
+
+if _gather_profiling_kubelet_start_proxy; then
+    _gather_profiling_kubelet_collect_data_loop
+    _gather_profiling_kubelet_stop_proxy
+    echo "INFO: gather kubelet profiling data completed"
+else
+    echo "ERROR: failed connecting to the kubernetes apiserver via proxy"
```
sferich888, line 82: Do we have to connect via a proxy? or can we connect directly?

collection-scripts/gather_profiling_kubelet
```diff
@@ -0,0 +1,83 @@
...
+}
+
+_gather_profiling_kubelet_collect_data() {
+    local node="$1"
+    local output_base="${PROFILING_KUBELET_LOG_PATH}/${node}"
+    local output_heap="${output_base}_heap.out"
+    local output_prof="${output_base}_prof.out"
+    local url_base="http://localhost:8001/api/v1/nodes/${node}/proxy/debug/pprof"
+
+    curl -o "$output_prof" "${url_base}/profile?seconds=${PROFILING_KUBELET_DURATION_SEC}" > /dev/null 2>&1
```
sferich888, line 49: Should we swap this out with `oc raw` calls?

sferich888 (changes requested):
I think there may be more direct ways to get at this profiling data. Can we explore what that might look like?

collection-scripts/gather_profiling_kubelet
fgiudici: Right, should work also on windows nodes. Fixed, thanks.

collection-scripts/gather_profiling_kubelet
fgiudici: yeah, we don't need the proxy inside the must-gather pod... dropped!

collection-scripts/gather_profiling_kubelet
fgiudici: Thanks for the hint, moved to oc get --raw (and so we don't even need the proxy)!

fgiudici:
Thanks for the suggestions and the hints @sferich888, tried to follow them and now I think the script really looks more simple (and efficient)! (I have also rebased it on current master).

collection-scripts/gather_profiling_node
```diff
@@ -0,0 +1,66 @@
...
+        fi
+    done
+
+    [ -z "$pids" ] && return
+    echo "INFO: wait for kubelet profiling tasks"
+    wait ${pids[@]}
+}
+
+
+NODES="${@:-$(oc get nodes -o jsonpath='{.items[?(@.kind=="Node")].metadata.name}')}"
```
sferich888, line 58: Let's print a warning that this could take a large amount of time, if no nod list is passed to the script.

In short when you have 200 nodes doing the default run will take time, and we should warn our users of this.

sferich888 (approved):
I like the changes, one more nit

collection-scripts/gather_profiling_node
fgiudici: Thanks @sferich888 ! Added a warning when no nodes are passed on the command line.

davemulford:
@fgiudici - This looks good and I've run the script on a cluster to see what the outputs are.  One question or request I have is whether this is something a customer or support is likely to use; or whether this is more of a tool for engineering's use.

If we expect customers and/or support to use this functionality, I'd like to see instructions on how to read and interpret the profiling output, even if it's a link to upstream golang debugging docs.

RickJWagner (approved):
Reviewing at Eric's request (he can't make it today).

I'll be interested to see the response to Dave's question, otherwise looks good to me.

fgiudici:
@davemulford , @RickJWagner : thanks for jumping here!

> @fgiudici - This looks good and I've run the script on a cluster to see what the outputs are. One question or request I have is whether this is something a customer or support is likely to use; or whether this is more of a tool for engineering's use.
> If we expect customers and/or support to use this functionality, I'd like to see instructions on how to read and interpret the profiling output, even if it's a link to upstream golang debugging docs.

The usage I envision is to let a customer having issues with one or more Nodes (mainly on the kubelet) to provide easily (and in a familiar way, i.e., the must-gather tool) the profiling data to the Engineering (e.g., https://bugzilla.redhat.com/show_bug.cgi?id=2014073).
So, I expect customers/support to use it but just to provide information to the engineering.

fgiudici:
Hi, jumping back on this PR! Anything that can be improved or that should be done differently here?

PR #277, opened 2021-12-20
[MON-1234] Collect prometheus target details

sthaha:
This patch is related to [MON-1234](https://issues.redhat.com/browse/MON-1234) Jira issue. The 4d36ee0f which
originally addressed the ticket missed collecting Prometheus target
information which this patch addresses.

Additionally, this PR ...
1. Fixes a script error (introduced in d21e69ee) caused due to
   usage of an uninitialized variable - PODS

2. Makes proper use of array to iterate prometheus pods

3. Makes "path" parameter to `prom_get` and `alertmanager_get` functions
   optional to improve  readability. I.E. it now supports both
   *  `prom_get rules`
   *  `prom_get rules?type=alert  rules/alerts`

4. Passes shellcheck once again

See: https://issues.redhat.com/browse/MON-1234

Signed-off-by: Sunil Thaha <sthaha@redhat.com>

### DIFF
--- a/collection-scripts/gather_monitoring
+++ b/collection-scripts/gather_monitoring
@@ -5,11 +5,14 @@ set -o nounset
 set -o errexit
 set -o pipefail
 
+# global readonly constants
 declare -r BASE_COLLECTION_PATH="../must-gather"
 declare -r MONITORING_PATH="${BASE_COLLECTION_PATH}/monitoring"
 declare -r CA_BUNDLE="${MONITORING_PATH}/ca-bundle.crt"
-declare -r PROM_PODS="$(oc get pods -n openshift-monitoring -l prometheus=k8s -o jsonpath='{.items[*].metadata.name}')"
 
+
+# init initializes global variables that need to be computed.
+# E.g. get token of the default ServiceAccount
 init() {
   mkdir -p "${MONITORING_PATH}"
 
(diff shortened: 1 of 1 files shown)


sthaha:
cc: @simonpasquier  @wking

collection-scripts/gather_monitoring
```diff
@@ -101,14 +105,15 @@ monitoring_gather(){
...
-  prom_get alertmanagers alertmanagers  || true
-  prom_get status/config status/config || true
-  prom_get status/flags status/flags || true
-  prom_get status/runtimeinfo status/runtimeinfo || true
-  prom_get_from_replicas status/tsdb status/tsdb || true
+  prom_get rules    || true
+  prom_get alertmanagers   || true
+  prom_get status/config  || true
+  prom_get status/flags  || true
+  prom_get targets  || true
```
simonpasquier, line 112: I'd recommend that the script gathers from both replicas since they may not have the same view. I would also limit the query to active targets.

```suggestion
  prom_get_from_replicas targets?state=active active_targets  || true
```

RickJWagner (approved):
I offer no suggested improvements.
The script seems clean and well-written.

collection-scripts/gather_monitoring
sthaha: @simonpasquier thank you for the review. 

Should `prom_get_from_replicas` be used for all resources that we are collecting? 
There is also `alertmanager_get` does that need to change as well to collect from individual pods?

collection-scripts/gather_monitoring
```diff
@@ -98,17 +108,22 @@ alertmanager_get() {
...
-  prom_get status/runtimeinfo status/runtimeinfo || true
-  prom_get_from_replicas status/tsdb status/tsdb || true
+  prom_get alertmanagers   || true
+  prom_get rules    || true
+  prom_get status/config  || true
+  prom_get status/flags  || true
+  prom_get status/runtimeinfo  || true
+
+  # using prom_get_from_replica as the state differs for each replica
+  prom_get_from_replicas targets?state=active active-targets  || true
```
wking, line 123: nit: `?` is shell-sensitive, so we should probably quote `'targets?state=active'` here.  Although the likelihood of there being a file to match this path expansion is very low, so not a big risk.

sthaha: Good catch :)

PR #279, opened 2022-01-24
Gather machine-config-daemon-pull host service logs

LorbusChris:
Having these will make issues like https://github.com/openshift/machine-config-operator/pull/2925 easier to debug

/cc @vrutkovs

### DIFF
--- a/collection-scripts/gather_service_logs
+++ b/collection-scripts/gather_service_logs
@@ -5,8 +5,8 @@ BASE_COLLECTION_PATH="/must-gather"
 ROLES=${1:-master}         ### Defaults to only collecting things from Masters
 
 # Service Lists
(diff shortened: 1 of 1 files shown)


vrutkovs:
LGTM

sferich888:
How exactly does this help debug this service? As these are systems shell execs is anything even logged?

vrutkovs:
>How exactly does this help debug this service? As these are systems shell execs is anything even logged?

This service is used in OKD only to extract `machine-config-daemon` from MCO image. We need these logs to find out why master/node is stuck, logging which command got stuck and their output

vrutkovs:
Actually, we don't really need logs from `-pull`. If it failed then the node didn't join the cluster and must-gather won't be able to reach it

LorbusChris:
Right, the correct place to gather these logs would be on the bootstrap node in the installer. Tabling this for now to see where the CoreOS layering effort leads us - the mcd-pull service might not be needed anymore with that.

PR #280, opened 2022-01-25
Bug 2041087: Revert MetalLB  speaker component label

liornoy:
changing the MetalLB speaker component label back to the old
version as this change produced bugs on MetalLB and the label
was reverted there as well.

[changed from: app.kubernetes.io/component=speaker to: component=speaker]


fedepaol:
@sferich888 mind approving this? We reverted the label selector to the value we had in 4.10

PR #282, opened 2022-02-08
Adding SR-IOV resource collection

sferich888:
@rjhowe please review this as I know your working an issue currently where you think this would be helpful. 

cc: @trozet @knobunc @mcurry-rh @Deepthidharwar  for review (as this will either need to be treated as a 'bug' (missing functionality for debugging) or an RFE (enhancement to diagnostics collection capabilities), and thus have the proper trackers/paperwork filed to get this pushed though CI/automated builds/releases.

### DIFF
--- a/collection-scripts/gather_sriov
+++ b/collection-scripts/gather_sriov
@@ -1,0 +1,7 @@
+#!/bin/bash
+
+sriov_resources+=(ns/openshift-sriov-network-operator sriovnetworks sriovnetworknodestates sriovnetworknodepolicies sriovnetworks sriovoperatorconfigs network-attachment-definitions)
+
+oc adm inspect --dest-dir must-gather --rotated-pod-logs "${sriov_resources[@]}"
+
+sync
--- a/collection-scripts/gather
+++ b/collection-scripts/gather
@@ -82,5 +82,8 @@ oc adm inspect --dest-dir must-gather --rotated-pod-logs "${group_resources_text
 # Gather metallb logs
 /usr/bin/gather_metallb_logs
 
+# Gather SR-IOV resources
+/usr/bin/gather_sriov
+
 # force disk flush to ensure that all data gathered is accessible in the copy container
 sync


RickJWagner (approved):
Looks good.

collection-scripts/gather_sriov
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash
+
+sriov_resources+=(ns/openshift-sriov-network-operator sriovnetworks sriovnetworknodestates sriovnetworknodepolicies sriovnetworks sriovoperatorconfigs network-attachment-definitions)
```
zshi-redhat, line 3: There are two sriovnetworks.

zshi-redhat: It would be needed to collect some additional sriov resources: sriovibnetworks, sriovnetworkpoolconfigs 
Other related resources are: nodes machineconfigs machineconfigpools featuregates kubeletconfigs ns/openshift-multus ns/openshift-marketplace (I assume these are already collected as part of must-gather).

There is also a must-gather script under sriov-network-operator repo: https://github.com/openshift/sriov-network-operator/blob/master/must-gather/collection-scripts/gather

SchSeba:
@sferich888 thanks for this PR!

please collect the same info we do in the sriov specific bundle https://github.com/openshift/sriov-network-operator/blob/master/must-gather/collection-scripts/gather

also another important artifact to collect is dmesg (journalctl -k) which helps us to identify driver issues on the network nics.

sferich888:
If we already have this in a different image it's better to collect it that way! 

/close

cgoncalves:
Here are some reasons off the top of my head to move the SR-IOV must-gather collector to the must-gather repository:

- No need for us to maintain a separate container image (e.g. rebuild periodically or async to include new bug fixes or CVEs and add build to erratum)
- Users do not need to pull yet another container image
- Simpler documentation (no need to document SR-IOV must-gather container image, see [documentation](https://docs.openshift.com/container-platform/4.9/support/gathering-cluster-data.html#gathering-data-specific-features_gathering-cluster-data))
- Avoids issues like the one fixed in https://github.com/openshift/sriov-network-operator/pull/621
- Users would only need to run must-gather once (as opposed to twice: 1) must-gather built-in scripts to collect network information etc, and 2) SR-IOV must-gather to collect SR-IOV information)
- SR-IOV must-gather code is in [upstream repository](https://github.com/k8snetworkplumbingwg/sriov-network-operator/tree/master/must-gather). Must-gather is OCP-specific so maintaining its code in upstream doesn't make sense. One could argue we could maintain it only in the [downstream repository](https://github.com/openshift/sriov-network-operator/tree/master/must-gather) but we would still have not answered the items above
- The SR-IOV must-gather collector script could be improved to run only if an SR-IOV Operator subscription is present, similarly to what has been done for MetalLB.

PR #283, opened 2022-02-08
Fixing openshift/must-gather#198 and enabling collection

sferich888:
This fixes the resources are in the right array and thus collected by default. 

cc: @soltysh @bcrochet


sferich888:
Linking to openshift/must-gather#198

soltysh (approved):
/lgtm
/approve
/retest

RickJWagner (approved):
Reviewed, looks good.  Network seems more appropriate for 'group' level.

PR #291, opened 2022-03-15
Add time range limit to Windows node logs

jrvaldes:
This PR adds a time range limit to the collection of logs
for Windows nodes.


jrvaldes:
/close

`--since` not available for Windows logs.

PR #292, opened 2022-03-16
Bug 2062355: Collect NMState resources when operator is installed

creydr:
Currently we are missing the NMState resources in the must-gathers. This is especially a problem as the kubernetes-nmstate-operator became GA in 4.10. 
This PR addresses it and adds the NMState resources (namespace resources & CRs) to the must-gathers when the operator is installed.

**Hints for reviewers:**
I was following a similar approach to how MetalLB is collecting its resources ([./collection-scripts/gather_metallb_logs](https://github.com/openshift/must-gather/blob/master/collection-scripts/gather_metallb_logs)). If there is an easier way (e.g. by adding a certain annotation), please let me know.

**How to verify:**
1. Deploy a cluster and install the NMState operator (e.g. via the marketplace)
2. Add the NMState CR:
    ```
    cat <<EOF | oc apply -f -
    apiVersion: nmstate.io/v1
    kind: NMState
    metadata:
      name: nmstate
    EOF
    ```
3. Wait for the handler being deployed on all nodes (e.g. `oc -n openshift-nmstate get po -w`)
4. Add a simple nodenetworkconfigurationpolicy. e.g.:
    ```
    cat <<EOF | oc apply -f -
    apiVersion: nmstate.io/v1
    kind: NodeNetworkConfigurationPolicy
    metadata:
      name: br1-eth0
    spec:
      desiredState:
        interfaces:
        - name: br1
          type: linux-bridge
          state: up
          ipv4:
            dhcp: true
            enabled: true
          bridge:
            port:
            - name: eth0
    EOF
    ```
5. Collect the must-gathers and check for the nmstate resouces (e.g. `omg get nnce`)

### DIFF
--- a/collection-scripts/gather
+++ b/collection-scripts/gather
@@ -82,5 +82,8 @@ oc adm inspect --dest-dir must-gather --rotated-pod-logs "${group_resources_text
 # Gather metallb logs
 /usr/bin/gather_metallb_logs
 
+# Gather NMState
+/usr/bin/gather_nmstate
+
 # force disk flush to ensure that all data gathered is accessible in the copy container
 sync
--- a/collection-scripts/gather_nmstate
+++ b/collection-scripts/gather_nmstate
@@ -1,0 +1,28 @@
+#!/bin/bash
+BASE_COLLECTION_PATH="must-gather"
+NMSTATE_NS="$(oc get subs -A -o template --template '{{range .items}}{{if eq .spec.name "kubernetes-nmstate-operator"}}{{.metadata.namespace}}{{end}}{{end}}')"
+NMSTATE_API_GROUP="nmstate.io"
+
+if [ -z "${NMSTATE_NS}" ]; then
+    echo "INFO: NMState not detected. Skipping."
+    exit 0
+fi
+
+function get_nmstate_cr() {
+    declare -a NMSTATE_CRDS=("nmstates" "nodenetworkconfigurationenactments" "nodenetworkconfigurationpolicies" "nodenetworkstates")
+    for NMSTATE_CRD in "${NMSTATE_CRDS[@]}"; do
+        CR_PATH="${BASE_COLLECTION_PATH}/cluster-scoped-resources/${NMSTATE_API_GROUP}/${NMSTATE_CRD}"
+        mkdir -p ${CR_PATH}
+
+        for CR_NAME in $(oc get ${NMSTATE_CRD} -ojsonpath='{.items[*].metadata.name}'); do
+            oc get ${NMSTATE_CRD} ${CR_NAME} -o yaml > "${CR_PATH}/${CR_NAME}.yaml"
+        done
+    done
+}
+
+oc adm inspect --dest-dir must-gather "ns/${NMSTATE_NS}"
+
+get_nmstate_cr
+
+# force disk flush to ensure that all data gathered is accessible in the copy container
+sync


creydr:
/retitle Bug 2062355: Collect NMState resources when operator is installed

creydr:
@cybertron @qinqon @rhrazdil: fyi

qinqon:
@creydr we have a must-gather bits to retrieve knmstate stuff at openshift virtualization, maybe you can take inspiration from there.

collection-scripts/gather_nmstate_logs
```diff
@@ -0,0 +1,29 @@
+#!/bin/bash
+BASE_COLLECTION_PATH="must-gather"
+NMSTATE_NS="$(oc get subs -A -o template --template '{{range .items}}{{if eq .spec.name "kubernetes-nmstate-operator"}}{{.metadata.namespace}}{{end}}{{end}}')"
+NMSTATE_PODS_PATH="${BASE_COLLECTION_PATH}/namespaces/${NMSTATE_NS}/pods"
```
qinqon, line 4: This is not used

qinqon (changes requested):

collection-scripts/gather_nmstate_logs
creydr: Good catch. Thanks
Updated it

collection-scripts/gather
```diff
@@ -82,5 +82,8 @@ oc adm inspect --dest-dir must-gather --rotated-pod-logs "${group_resources_text
 # Gather metallb logs
 /usr/bin/gather_metallb_logs
 
+# Gather NMState logs
+/usr/bin/gather_nmstate_logs
```
qinqon, line 86:
```suggestion
# Gather NMState CRs
/usr/bin/gather_nmstate_crs
```

collection-scripts/gather_nmstate_logs
```diff
@@ -0,0 +1,28 @@
+#!/bin/bash
```
qinqon, line 1: Do we fail must-gather if one of the script fails ? for tha we need to do `set -e`

qinqon (changes requested):

collection-scripts/gather
creydr: It is not only about CRs (but not only about logs neither). WDYT about `gather_nmstate`?

collection-scripts/gather_nmstate_logs
creydr: That's probably something @RickJWagner or @sferich888 can answer better

collection-scripts/gather
creydr: Updated

RickJWagner (approved):
This looks good to me.

One point to consider:  There's a lot of similarity between gather_nmstate and scripts like gather_metallb_logs.  Maybe consider factoring out the common stuff into a script that could be sourced by other scripts.  (DRY).

collection-scripts/gather_nmstate_logs
RickJWagner: Given that must-gather might run in clusters that are having problems, we should probably tend to be more forgiving of errors.  (And let things run as bullet-proof as possible, even if it means we give extra leniency to components that don't complete their jobs error-free.)  We'd prefer to get as much information as possible, even if it's some parts are unavailable.

cybertron:
Tried it locally, looks like it captured everything we need.
/lgtm

PR #296, opened 2022-04-12
Bug 1999529: Filter disabled resources in inspection to proceed others

ardaguclu:
Some clusters might disable some core resources. This causes
must-gather failing while trying to get these disabled resources and
can not proceed to other ones.

This PR adds filtering for core resources by checking they exist in
cluster or not.

### DIFF
--- a/collection-scripts/gather
+++ b/collection-scripts/gather
@@ -42,7 +42,16 @@ resources+=(nodenetworkstates nodenetworkconfigurationenactments nodenetworkconf
 # Run the Collection of Resources using inspect
 # running across all-namespaces for the few "Autoscaler" resources.
 oc adm inspect --dest-dir must-gather --rotated-pod-logs "${named_resources[@]}"
-group_resources_text=$(IFS=, ; echo "${group_resources[*]}")
+
+filtered_group_resources=()
+for gr in "${group_resources[@]}"
+do
+  oc get "$gr" > /dev/null 2>&1
+  if [[ "$?" -eq 0 ]]; then
+    filtered_group_resources+=("$gr")
+  fi
+done
+group_resources_text=$(IFS=, ; echo "${filtered_group_resources[*]}")
 oc adm inspect --dest-dir must-gather --rotated-pod-logs "${group_resources_text}"
 
 # Gather Insights Operator Archives


ardaguclu:
@davemulford @RickJWagner could you please take a look at this fix?. Thanks.

RickJWagner (approved):
Looks good. No objections, approved.

ardaguclu:
@soltysh would you mind reviewing this change?. Thanks.

collection-scripts/gather
```diff
@@ -42,7 +42,16 @@ resources+=(nodenetworkstates nodenetworkconfigurationenactments nodenetworkconf
...
+
+filtered_group_resources=()
+for gr in "${group_resources[@]}"
+do
+  oc get "$gr" > /dev/null 2>&1
+  if [[ "$?" -eq 0 ]]; then
+    filtered_group_resources+=("$gr")
+  fi
+done
+group_resources_text=$(IFS=, ; echo "${filtered_group_resources[*]}")
```
soltysh, line 54: @ardaguclu I'm inclined to fix `oc adm inspect` such that it will expose a flag `--ignore-not-found` similar to get or delete commands, which will print information about missing resource but will continue working which would be much better approach and will make this script cleaner.

soltysh (changes requested):
/hold

collection-scripts/gather
ardaguclu: After investigating the problem further, I noticed that `--ignore-not-found` might not cover our case. Because this is not an error comes from api server which we can ignore with `IsNotFound`, it is directly returned by [builder](https://github.com/kubernetes/kubernetes/blob/cc85d27a9c07a92e9d013d8972771e77e7855933/staging/src/k8s.io/cli-runtime/pkg/resource/builder.go#L795). This is also the case for `get` command.

For example, I deleted `volumesnapshotclasses` CRD.

```bash
$ oc get crd/volumesnapshotclasses
Error from server (NotFound): customresourcedefinitions.apiextensions.k8s.io "volumesnapshotclasses" not found
$ oc get crd/volumesnapshotclasses --ignore-not-found
results nothing as expected
```
However;
```bash
$ oc get volumesnapshotclasses
error: the server doesn't have a resource type "volumesnapshotclasses"
$ oc get volumesnapshotclasses --ignore-not-found
error: the server doesn't have a resource type "volumesnapshotclasses"
```

`ignore-not-found` handles the cases in `visit`. But if the group resource is not found by builder, it returns error and exits.

Since builder immediately returns error, it does not continue to other group resources and even if we intercept error and filter somehow, we'll still not have others.

If we want to preserve this script clean, maybe we can manually `oc get` each group resources and send filtered resources to builder in `adm inspect`.

What do you think @soltysh ?

soltysh: Two things, we should probably turn on `ignore-not-found` I just tried `oc adm inspect ns/default ns/nonexistent` and that failed, where it shouldn't. But for the non-existent resource type I think you're right we can leave it as is.

soltysh (approved):
/hold cancel
/lgtm
/approve

PR #301, opened 2022-05-09
fix gather_insights collection

npinaeva:
Looks like this collection has never worked. 
This PR fixes it, but I am not sure if we need this collection command, since nobody noticed it's broken.

Signed-off-by: Nadia Pinaeva <npinaeva@redhat.com>


RickJWagner (approved):
lgtm.  As @npinaeva mentioned, it seems likely this isn't vital functionality as nobody has complained about it being broken.  
I don't see anything objectionable, tough, so 'approved'.

PR #315, opened 2022-06-21
Stop collecting Docker logs on Windows instances

saifshaikh48:
With the switch to containerd as default runtime, we no longer want
to collect logs logs for the Docker Windows service.


jrvaldes (approved):
Thanks @saifshaikh48  for working on this.

saifshaikh48:
This change can wait until after 4.11 code freeze ends to merge. Trying to collect docker logs on Windows instances that do not have the docker runtime will not cause must gather to fail; instead the `docker_winevent.log` file will contain an error output from

saifshaikh48:
@alinaryan Yep containerd logs are already collected, see this commit https://github.com/openshift/must-gather/commit/df3666c2f2fc043414963564d4ccf69f81a995cb where it was introduced

PR #316, opened 2022-06-24
adding proxy information

sferich888:
@wking If I read https://docs.openshift.com/container-platform/4.9/networking/enable-cluster-wide-proxy.html#nw-proxy-configure-object_config-cluster-wide-proxy correctly this is how we should be collecting this information.


RickJWagner (approved):
Looks good.  Approved.

PR #317, opened 2022-07-06
Bug 2103283: Add timeout to oc cp command to fix must-gather delays when routers are terminating

gcs278:
This fixes a bug where if router pods are stuck in terminating when doing a must-gather, the must-gather will timeout instead of waiting for the router pods to terminate. More specifically, we have a bug in 4.10 CI in which the e2e-aws-operator job consistently fails on e2e-aws-operator-gather-must-gather container test due to pods stuck in the `terminating` state. 

Two reasons why this happens in CI:
1. If the `oc cp` command is ran on `terminating` pods, it will hang until the pod is deleted.
2. Router pods ignore SIGTERMs sent immediately after starting up and will get "stuck" in terminating state until the `terminationGracePeriodSeconds` sends a SIGKILL. This results in a number of old router pods just hanging out in the terminating state. Backport is pending for 4.10 https://bugzilla.redhat.com/show_bug.cgi?id=2098230.

Either way, seems like we should have a reasonable timeout on the `oc cp` command to protect against situations like this.

### DIFF
--- a/collection-scripts/gather_haproxy_config
+++ b/collection-scripts/gather_haproxy_config
@@ -11,7 +11,7 @@ function gather_haproxy_config {
     for IC in ${INGRESS_CONTROLLERS}; do
         PODS=$(oc get pods -n openshift-ingress --no-headers -o custom-columns=":metadata.name" --selector ingresscontroller.operator.openshift.io/deployment-ingresscontroller="${IC}")
         for POD in ${PODS}; do
-            oc cp openshift-ingress/"${POD}":/var/lib/haproxy/conf/haproxy.config "${HAPROXY_CONFIG_PATH}"/"${IC}"/"${POD}"/haproxy.config &
+            timeout -v 3m oc cp openshift-ingress/"${POD}":/var/lib/haproxy/conf/haproxy.config "${HAPROXY_CONFIG_PATH}"/"${IC}"/"${POD}"/haproxy.config &
             PIDS+=($!)
         done
     done


collection-scripts/gather_haproxy_config
```diff
@@ -11,7 +11,7 @@ function gather_haproxy_config {
     for IC in ${INGRESS_CONTROLLERS}; do
         PODS=$(oc get pods -n openshift-ingress --no-headers -o custom-columns=":metadata.name" --selector ingresscontroller.operator.openshift.io/deployment-ingresscontroller="${IC}")
         for POD in ${PODS}; do
-            oc cp openshift-ingress/"${POD}":/var/lib/haproxy/conf/haproxy.config "${HAPROXY_CONFIG_PATH}"/"${IC}"/"${POD}"/haproxy.config &
+            timeout -v 3m oc cp openshift-ingress/"${POD}":/var/lib/haproxy/conf/haproxy.config "${HAPROXY_CONFIG_PATH}"/"${IC}"/"${POD}"/haproxy.config &
```
sferich888, line 14: Is a 3 min timeout really how long we want to wait with this? 
This copy operation should be seconds; IE 10x (seconds) is likely a 1-1.5 m wait time (can we half this)?

sferich888:
> What is `terminationGracePeriodSeconds` set to for router pods? Is it something in the order of minutes?

terminationGracePeriodSeconds as specified in the router deployment, which is 1 hour.
xref: https://bugzilla.redhat.com/show_bug.cgi?id=2098230

collection-scripts/gather_haproxy_config
sferich888: https://coreos.slack.com/archives/CFDM5CQMN/p1657143478060469?thread_ts=1656698903.916469&cid=CFDM5CQMN discussed this. 

/lgtm

PR #325, opened 2022-07-26
Remove no longer needed gather_admission_webhooks script

nekop:
In 4.10, the webhooks info gathered by a part of kube-apiserver co.

```
$ oc version
Client Version: 4.10.22
Server Version: 4.10.22
Kubernetes Version: v1.23.5+3afdacb

$ oc get co kube-apiserver -o yaml
  - group: admissionregistration.k8s.io
    name: ""
    resource: mutatingwebhookconfigurations
  - group: admissionregistration.k8s.io
    name: ""
    resource: validatingwebhookconfigurations
$ ls ./must-gather/cluster-scoped-resources/admissionregistration.k8s.io/
mutatingwebhookconfigurations  validatingwebhookconfigurations
$ ls ./must-gather/web_hooks/
mutatingwebhookconfigurations.yaml  validatingwebhookconfigurations.yaml # duplicated info
```

### DIFF
--- a/collection-scripts/gather_admission_webhooks
+++ b/collection-scripts/gather_admission_webhooks
@@ -1,7 +1,0 @@
-#!/bin/bash
-BASE_COLLECTION_PATH="/must-gather"
-WEB_HOOK_PATH="${BASE_COLLECTION_PATH}/web_hooks/"
-
-mkdir -p ${WEB_HOOK_PATH}
-oc get mutatingwebhookconfigurations -o yaml > ${WEB_HOOK_PATH}/mutatingwebhookconfigurations.yaml
-oc get validatingwebhookconfigurations -o yaml > ${WEB_HOOK_PATH}/validatingwebhookconfigurations.yaml
--- a/collection-scripts/gather
+++ b/collection-scripts/gather
@@ -69,10 +69,6 @@ oc adm inspect --dest-dir must-gather --rotated-pod-logs "${group_resources_text
 # Gather etcd information
 /usr/bin/gather_etcd
 
-# Gather Validation and Mutating Admission Webhook Objects
-# TODO the kube-apiserver operator should list these as related resources
-/usr/bin/gather_admission_webhooks
-
 # Gather Service Logs (using a supplemental Script); Scoped to Masters.
 /usr/bin/gather_service_logs master
 


RickJWagner (approved):
Less is more, the best code is the code that's not there.  
Approved.

sferich888:
@nekop we would/will need to backport this to 4.10 correct? 
cc: @deads2k @sttts (is this ok to remove; and backport)?

collection-scripts/gather_admission_webhooks
```diff
@@ -1,7 +0,0 @@
-#!/bin/bash
-BASE_COLLECTION_PATH="/must-gather"
-WEB_HOOK_PATH="${BASE_COLLECTION_PATH}/web_hooks/"
-
-mkdir -p ${WEB_HOOK_PATH}
-oc get mutatingwebhookconfigurations -o yaml > ${WEB_HOOK_PATH}/mutatingwebhookconfigurations.yaml
```
sferich888, line 6: This script/code collects all `mutatingwebhookconfigurations` and `validatingwebhookconfigurations` and places them in the `${BASE_COLLECTION_PATH}/web_hooks` directory. 

Thus your [comment](https://github.com/openshift/must-gather/pull/325#issue-1317892153) is accurate in that this information is duplicated; but only because the [kube-apiserver (cluster operator) is defining `.status.relatedObjects` ](https://github.com/openshift/enhancements/blob/master/enhancements/oc/inspect.md)which also collects this information.

Thus its probably best that we only remove the [comment](https://github.com/openshift/must-gather/pull/325/files#diff-bad6eb3f5953b7f47f014fd8b04a48470c1d64d94bbd29566af842f0b60c909cL73) about the kubeapi-server doing this (because it now does this). However the other collections (or duplications) may also be warranted (IMO) as we may want to know about what other  `mutatingwebhookconfigurations` and `validatingwebhookconfigurations` are in place. 

We want to do this for 2 reasons: 

1. we may not be collecting the 'resource' that defines the  `.status.relatedObjects` relationship; and thus would not be able to see / understand if something causing mutating or validation behaviors. 

1. for resources we collect, they may don't properly define this `.status.relatedObjects` relationship, and thus would not be able to see / understand if something causing mutating or validation behaviors, if such a hook was in place. 


Thus it might be prudent for us to keep this script in place.

deads2k: > we may not be collecting the 'resource' that defines the .status.relatedObjects relationship; and thus would not be able to see / understand if something causing mutating or validation behaviors.

This would be an operator bug, so a risk, but if it's not observed in the field, I'd allow this.

> 2\. for resources we collect, they may don't properly define this `.status.relatedObjects` relationship, and thus would not be able to see / understand if something causing mutating or validation behaviors, if such a hook was in place.

This would be an `oc adm inspect` bug, so a risk, but if it's not observed in the field, I'd allow this.


Question to the author: have you considered expanding `oc adm inspect` to add namespaces containing services references by these webhooks?

nekop: > Question to the author: have you considered expanding oc adm inspect to add namespaces containing services references by these webhooks?

That may be useful when we have webhooks related issue. It could be considered an unexpected behavior for users if we gather user namespace info automatically as it may contain other apps and/or sensitive information. I'm not sure if we want this behavior with standard cluster operator must-gather.

nekop:
As this is a minor change, I have no backport preference. Latest only fix would be fine for me.

PR #326, opened 2022-08-22
updating version file to better meet spec.

sferich888:


sferich888:
@mfojtik do we want this version to be the must-gather version or the version of the cluster? 
If we wan this to be the must-gather version then this PR isn't what we want. That said; the fallback mechanisms in this don't meet https://github.com/openshift/enhancements/blob/master/enhancements/oc/must-gather.md#must-gather-images (3) in the spec.

RickJWagner (commented):
No offending statements noted in the PR, but the PR has been closed.  Moot review.

PR #341, opened 2022-12-09
Add Performance Profile Controler must-gather

marioferh:
# Description

Launch Performance Profile Controller data collecting in OCP must-gather

# Goal

Remove a standalone PAO must-gather image and make it part of the core OCP must-gather.

Before, the command for the client to run must-gather and PPC must-gather. 

oc adm must-gather \
 --image-stream=openshift/must-gather \ 
 --image=[...]/performance-addon-operator-must-gather-rhel8:$VERSION
 
Now:
oc adm must-gather --image=openshift/must-gather:$VERSION /usr/bin/gather_performance_profile_controller

# Type of change

- [x]  Add PPC scripts
- [X]  Add two stage build to dockerfile in order to get sysinfo binary needed to collect data by scripts
- [X]  Add PPC configuration yaml
- [ ]  Tests not included in this repo. Go to NTO-PPC repo.

# Needed changes

Change temp ` ARG REPO=github.com/marioferh/gather-sysinfo` to actual repo before merge.

PAO = Performance Addon Operator now is  PPC = Performance Profile Controller after this change: https://github.com/openshift/cluster-node-tuning-operator/pull/322


performance-profile-node-gather/clusterrolebinding.yaml
```diff
@@ -0,0 +1,13 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRoleBinding
+metadata:
+  name: perf-node-gather-pods-reader
+subjects:
+- kind: ServiceAccount
```
sferich888, line 6: Isn't this something we provide in its own file?

performance-profile-node-gather/clusterrolebinding.yaml
```diff
@@ -0,0 +1,13 @@
...
+  name: perf-node-gather
+  namespace: perf-node-gather
+  apiGroup: ""
+roleRef:
+  kind: ClusterRole
```
sferich888, line 11: Isn't this something we are providing int its own file?

Dockerfile.rhel7
```diff
@@ -1,8 +1,27 @@
...
+ARG REPO=github.com/marioferh/gather-sysinfo
+WORKDIR /root
+RUN git clone https://$REPO
+WORKDIR /root/gather-sysinfo 
+RUN make dist-gather-sysinfo
+
 FROM registry.ci.openshift.org/ocp/4.12:cli
 COPY --from=builder /go/src/github.com/openshift/must-gather/collection-scripts/* /usr/bin/
 RUN yum install --setopt=tsflags=nodocs -y jq && yum clean all && rm -rf /var/cache/yum/*
+RUN yum install -y pciutils util-linux hostname rsync tar ethtool
```
sferich888, line 19: This should come before the `yum clean all && rm -rf /var/cache/yum/*` parts above.

collection-scripts/gather_performance_profile_controller
```diff
@@ -0,0 +1,162 @@
...
+POD_NAME=${HOSTNAME}
+MUST_GATHER_IMAGE=$(oc get pod -n $NAMESPACE $POD_NAME -o jsonpath="{.spec.containers[0].image}")
+
+POD_IP=$(hostname -I |  tr -d "[:blank:]" )
+echo "[$NAMESPACE/$POD_IP/$POD_NAME]" >> ${BASE_COLLECTION_PATH}/debug
+oc get pod -n $NAMESPACE $POD_NAME -o json >> ${BASE_COLLECTION_PATH}/debug
+
+sed -i -e "s#MUST_GATHER_IMAGE#$MUST_GATHER_IMAGE#" $DAEMONSET_MANIFEST
+
+oc create -f $NAMESPACE_MANIFEST
```
sferich888, line 42: Can we simply all of the `oc create` commands by using `oc apply`?

collection-scripts/gather_performance_profile_controller
```diff
@@ -0,0 +1,162 @@
...
+oc create -f $SERVICEACCOUNT_MANIFEST
+oc adm policy add-scc-to-user privileged -n perf-node-gather -z perf-node-gather
+oc create -f $CLUSTER_ROLE_MANIFEST
+oc create -f $CLUSTER_ROLE_BINDING_MANIFEST
+oc create -f $DAEMONSET_MANIFEST
+
+COUNTER=0
+until check_node_gather_pods_ready || [ $COUNTER -eq 300 ]; do
+   (( COUNTER++ ))
+   sleep 1
```
sferich888, line 52: This is a 5 minute wait time for the daemon set to start; can we provide some output or visual indicator to the user on why we are waiting?

collection-scripts/gather_performance_profile_controller
```diff
@@ -0,0 +1,162 @@
...
+# https://kubernetes.io/docs/concepts/services-networking/dns-pod-service/#pod-s-hostname-and-subdomain-fields
+POD_NAME=${HOSTNAME}
+MUST_GATHER_IMAGE=$(oc get pod -n $NAMESPACE $POD_NAME -o jsonpath="{.spec.containers[0].image}")
+
+POD_IP=$(hostname -I |  tr -d "[:blank:]" )
+echo "[$NAMESPACE/$POD_IP/$POD_NAME]" >> ${BASE_COLLECTION_PATH}/debug
+oc get pod -n $NAMESPACE $POD_NAME -o json >> ${BASE_COLLECTION_PATH}/debug
+
+sed -i -e "s#MUST_GATHER_IMAGE#$MUST_GATHER_IMAGE#" $DAEMONSET_MANIFEST
+
```
sferich888, line 41: Can we add a check somewhere at the start of this; that check for the node label `perf-node-gather-daemonset` on at least 1 node.

sferich888: We should error out early if we can't even deploy this daemon set to a node. If we don't do this; a user could wait for 5+ minutes before they figure out why this script didn't work/run.

collection-scripts/gather_performance_profile_controller
```diff
@@ -0,0 +1,162 @@
...
+function ppc_nodes(){
+  
+check_node_gather_pods_ready() {
+   line=$(oc get ds perf-node-gather-daemonset -o=custom-columns=DESIRED:.status.desiredNumberScheduled,READY:.status.numberReady --no-headers -n perf-node-gather)
+
+    IFS=$' '
+    read desired ready <<< $line
+    IFS=$'\n'
+
+    if [[ "$desired" != "0" ]] && [[ "$ready" == "$desired" ]]
```
sferich888, line 12: This leave no room for error (IE all pods in the daemonset have to come up and be working). 

I worry about how 'strict this is'. While you can control how many nodes this runs on with `perf-node-gather-daemonset` labels; on systems where you 'expect things to not be working properly' this level of 'stringency' may be problematic. 

That said I understand the chicken & egg situation this creates, we should just look at how we can make this a bit more forgiving (if possible).

collection-scripts/gather_performance_profile_controller
```diff
@@ -0,0 +1,162 @@
...
+oc create -f $CLUSTER_ROLE_BINDING_MANIFEST
+oc create -f $DAEMONSET_MANIFEST
+
+COUNTER=0
+until check_node_gather_pods_ready || [ $COUNTER -eq 300 ]; do
+   (( COUNTER++ ))
+   sleep 1
+done
+
+for line in $(oc get pod -o=custom-columns=NODE:.spec.nodeName --no-headers --field-selector=status.phase!=Running -n perf-node-gather)
```
sferich888, line 55: Will we ever get to this check? the `check_node_gather_pods_ready` function requires that all pods be running in order for the collection to continue?

collection-scripts/gather_performance_profile_controller
sferich888: Note: this is just a startup wait time; this doesn't include the run/collection of data.

collection-scripts/gather_performance_profile_controller
sferich888, line 42: Can we 'reuse' the exsisting must-gather namespace; instead of creating a new namespace? 

By putting all the collection artifacts in the 'must-gather' namespace; when the `oc` command that invokes must-gather finishes; a clean up operation is done (removing all of the pods, and other artifacts that we create). 

By spawning off new namespaces; it's possible that we could leave around lingering artifacts.

collection-scripts/gather_performance_profile_controller
```diff
@@ -0,0 +1,162 @@
...
+  # trying again with the pods, which are _usually_ reliable - but users can change them
+  [ -z "${ns}" ] && ns=$( oc get pods -A -l name='performance-operator' -o=jsonpath='{.items[0].metadata.namespace}{"\n"}' 2> /dev/null )
+  # namespace suggested by the documentation. This is a fancier way to check for its existence
+  [ -z "${ns}" ] && ns=$( oc get ns openshift-performance-addon-operator -o jsonpath='{.metadata.name}{"\n"}' 2> /dev/null )
+  # we should never get there. This is the last resort.
+  [ -z "${ns}" ] && ns="openshift-operators"
+  echo ${ns}
+}
+
+function version() {
```
sferich888, line 118: you shouldn't need to do this as every must-gather file includes a version file specifying exactly what was used as part of the collection.

collection-scripts/gather_performance_profile_controller
```diff
@@ -0,0 +1,162 @@
...
+    NODE_PATH=${NODES_PATH}/$NODE
+    mkdir -p ${NODE_PATH}
+    for UNIT in ${NODE_UNITS[@]}; do
+        timeout -k 5m 30m bash -c "oc adm node-logs $NODE -u $UNIT --since '-8h' | gzip" > ${NODE_PATH}/${NODE}_logs_$UNIT.gz &
+	ADM_PIDS+=($!)
+    done
+done
+wait "${ADM_PIDS[@]}"
+
+oc delete -f $DAEMONSET_MANIFEST
```
sferich888, line 99: I don't think we need to do this; if we rely on the invoking `oc` command to clean things up for us.

collection-scripts/gather_performance_profile_controller
```diff
@@ -0,0 +1,162 @@
...
+  # if version not found, fallback to imageID
+  [ -z "${version}" ] && version=$(oc status | grep '^pod.*runs' | sed -r -e 's/^pod.*runs //')
+
+  # if version still not found, use Unknown
+  [ -z "${version}" ] && version="Unknown"
+
+  echo ${version}
+}
+
+# generate /must-gather/version file
```
sferich888, line 134: Since you are including this 'add-on'script in the main must-gather archive - you don't need to do this version creation piece.

sferich888 (changes requested):
This looks great; however there is quite a bit I would like to change or have you consider.

Dockerfile.rhel7
marioferh: done

collection-scripts/gather_performance_profile_controller
marioferh: We need to do it with oc create to apply policies after the SA creation:
oc adm policy add-scc-to-user privileged -n $NAMESPACE -z perf-node-gather

collection-scripts/gather_performance_profile_controller
marioferh: If we continue with pods not ready we can miss the collection of some pods

collection-scripts/gather_performance_profile_controller
marioferh: I am facing this issue if I try to reuse same namespace:
  Warning  FailedCreate  33s (x15 over 115s)  daemonset-controller  Error creating: pods "perf-node-gather-daemonset-" is forbidden: autoscaling.openshift.io/ManagementCPUsOverride the pod namespace "openshift-must-gather-f2s49" does not allow the workload type management

collection-scripts/gather_performance_profile_controller
marioferh: right

collection-scripts/gather_performance_profile_controller
marioferh: for everyone?

collection-scripts/gather_performance_profile_controller
marioferh: right

marioferh (commented):
The new change create sysinfo binary inside the repo to not depend of external repo as before

.ci-operator.yaml
```diff
@@ -1,4 +1,4 @@
 build_root_image:
   name: release
   namespace: openshift
-  tag: rhel-8-release-golang-1.18-openshift-4.12
+  tag: rhel-8-release-golang-1.19-openshift-4.12
```
soltysh, line 4: rhel-8-release-golang-1.19-openshift-4.13 would be the correct one here and everywhere else too.

Dockerfile.rhel7
```diff
@@ -1,8 +1,20 @@
-FROM registry.ci.openshift.org/ocp/builder:rhel-8-golang-1.18-openshift-4.12 AS builder
+FROM registry.ci.openshift.org/ocp/builder:rhel-8-golang-1.19-openshift-4.12 AS builder
+
+RUN dnf install -y git make go
```
soltysh, line 3: Why you're doing this, all of those tools are already part of the builder image.

Dockerfile.rhel7
soltysh: It's still not correct you should just add your tools into the original line like so:
```
RUN yum install --setopt=tsflags=nodocs -y jq pciutils util-linux hostname rsync tar ethtool && yum clean all && rm -rf /var/cache/yum/*
```

Dockerfile.rhel7
```diff
@@ -1,8 +1,20 @@
...
+RUN yum install --setopt=tsflags=nodocs -y jq 
+RUN yum install -y pciutils util-linux hostname rsync tar ethtool
+RUN yum clean all && rm -rf /var/cache/yum/*
+
+ARG NODE_GATHER_MANIFESTS_DIR=performance-profile-node-gather/
+
+COPY ${NODE_GATHER_MANIFESTS_DIR} /etc/performance-profile-node-gather
+
+# the binary follows the must-gather convention of helper tools called gather_$SOMETHING
+COPY build/_output/bin/gather-sysinfo /usr/bin/gather_sysinfo
```
soltysh, line 20: I don't quite understand the lines above, what are they doing?

Makefile
```diff
@@ -1,5 +1,34 @@
...
+
+.PHONY: build-output-dir
+build-output-dir:
+	mkdir -p $(TOOLS_BIN_DIR) || :
+
+.PHONY: dist-gather-sysinfo
+dist-gather-sysinfo: build-output-dir
+	@if [ ! -x $(TOOLS_BIN_DIR)/gather-sysinfo ]; then\
+		echo "Building gather-sysinfo helper";\
+		env CGO_ENABLED=0 GOOS=$(TARGET_GOOS) GOARCH=$(TARGET_GOARCH) go build -ldflags="-s -w" -mod=vendor -o $(TOOLS_BIN_DIR)/gather-sysinfo ./pkg/tools/gather-sysinfo/;\
```
soltysh, line 28: Why you're building it? From the email thread it looked like you already have a package ready that you just install?

collection-scripts/gather_performance_profile_controller
soltysh: I second @sferich888 here, it's more reasonable to have some wait period before failing.

collection-scripts/gather_performance_profile_controller
```diff
@@ -0,0 +1,143 @@
...
+sed -i -e "s#MUST_GATHER_IMAGE#$MUST_GATHER_IMAGE#" $DAEMONSET_MANIFEST
+sed -i -e "s#NAMESPACE#$NAMESPACE#" $CLUSTER_ROLE_BINDING_MANIFEST
+sed -i -e "s#NAMESPACE#$NAMESPACE#" $DAEMONSET_MANIFEST
+sed -i -e "s#NAMESPACE#$NAMESPACE#" $SERVICEACCOUNT_MANIFEST
+
+oc create -f $SERVICEACCOUNT_MANIFEST
+oc adm policy add-scc-to-user privileged -n $NAMESPACE -z perf-node-gather
+oc create -f $CLUSTER_ROLE_MANIFEST
+oc create -f $CLUSTER_ROLE_BINDING_MANIFEST
+oc create -f $DAEMONSET_MANIFEST
```
soltysh, line 51: You can't create any of these, b/c must-gather CLI won't be able to clean it up. You should rather re-use the namespace and the privileged SA created for must-gather pod.

go.mod
```diff
@@ -2,4 +2,110 @@ module github.com/openshift/must-gather
 
 go 1.17
 
-require github.com/openshift/build-machinery-go v0.0.0-20210423112049-9415d7ebd33e
+require (
```
soltysh, line 5: Like mentioned before, I don't think we want to handle it here, it should be built outside of this repo and only installed here.

soltysh (changes requested):

Dockerfile.rhel7
marioferh: done

collection-scripts/gather_performance_profile_controller
marioferh: done

go.mod
MarSik: That would require an RPM in the package repos. The tool is specific to the must gather data collection though. Also we want the tool to be always in sync with the MG and OCP versions. Building it here is actually much much easier and more reliable.

soltysh: Based on what I was told it is already built as RPM so that shouldn't be the problem.

Makefile
MarSik: I do not think we have ever said that. We are currently building it as part of the separate must gather image we ship, but there never was an RPM as far as I know.

.ci-operator.yaml
marioferh: done

Dockerfile.rhel7
marioferh: done

Dockerfile.rhel7
marioferh: After generate a binary with make we need to copy it to the image

marioferh:
https://github.com/openshift/oc/pull/1316 oc PR to add required annotations to must-gather annotatoins

PR #347, opened 2023-02-14
WINC-977: Update kube-proxy log file name

mansikulkarni96:
This PR updates the kube-proxy log file name to sync with the recent changes made in the kube-proxy log collection implementation.

### DIFF
--- a/collection-scripts/gather_windows_node_logs
+++ b/collection-scripts/gather_windows_node_logs
@@ -3,7 +3,7 @@ BASE_COLLECTION_PATH="/must-gather"
 WINDOWS_NODE_LOGS=$BASE_COLLECTION_PATH/host_service_logs/windows
 
 # Logfile list
-LOGS=(kube-proxy/kube-proxy.exe.INFO kube-proxy/kube-proxy.exe.ERROR kube-proxy/kube-proxy.exe.WARNING)
+LOGS=(kube-proxy/kube-proxy.log)
 LOGS+=(hybrid-overlay/hybrid-overlay.log)
 LOGS+=(kubelet/kubelet.log)
 LOGS+=(containerd/containerd.log)


mansikulkarni96:
Merge after https://github.com/openshift/windows-machine-config-operator/pull/1411

collection-scripts/gather_windows_node_logs
```diff
@@ -3,7 +3,7 @@ BASE_COLLECTION_PATH="/must-gather"
 WINDOWS_NODE_LOGS=$BASE_COLLECTION_PATH/host_service_logs/windows
 
 # Logfile list
-LOGS=(kube-proxy/kube-proxy.exe.INFO kube-proxy/kube-proxy.exe.ERROR kube-proxy/kube-proxy.exe.WARNING)
+LOGS=(kube-proxy/kube-proxy.log)
```
aravindhp, line 6: My question here is if there is ever a situation where a newer version of must-gather can be run against an older cluster? If that can happen we would need to be backward compatible,

mansikulkarni96: The docs mention the need for using the [matching gathering tool version](https://github.com/openshift/must-gather/blob/3af14335d5b30106220643b09349e389a3e07cc0/must-gather.md#motivation)
`Simplify your gathering software by only using a matching version of the gathering tool. This simplifies code and test matrices so that your tool always works. For instance, OpenShift payloads include the exact level of gathering used to debug that payload. ` and ` It is tightly coupled to the openshift payload and only contains logic to gather information from that payload. We have e2e tests that make sure this functions.`
But I guess it is still a possibility to use an older version.

aravindhp: But given the doc, I don't think this is an issue.

davemulford: Customers technically could do this, but a default `oc adm must-gather` command will run what was pulled down when the cluster was created. By default, this should align with the correct cluster version. So no backwards-compatibility issues from my perspective.

PR #348, opened 2023-02-17
OCPNODE-1499: Add CMA gather script

jrvaldes:
This changes adds the gather script for the Custom Metrics Autoscaler (CMA) operator so that the operator resources are included in the must-gather if the operator is present.

Gathering the following:
- kedacontroller,
- scaledjob,
- scaledobject,
- triggerauthentication,
- clustertriggerauthentication and
- subscription namespace, e.g. ns/openshift-keda

PR #350, opened 2023-02-22
Revert "OCPNODE-1499: Add CMA gather script"

jrvaldes:
Reverts openshift/must-gather#348

The data collection scripts should only include collection logic for components that are included as part of the OpenShift
CVO payload. See https://github.com/openshift/must-gather/blob/master/README.md?plain=1#L10-L12

xref: https://redhat-internal.slack.com/archives/CFDM5CQMN/p1677090782806819?thread_ts=1677073218.803299&cid=CFDM5CQMN


RickJWagner (approved):
Approved, based on conversation at https://redhat-internal.slack.com/archives/CFDM5CQMN/p1677073218803299

PR #361, opened 2023-04-26
MCO-608: Gather MCO's on-disk configs from degraded nodes

jkyros:
There have been a lot of Machine Config Operator (MCO)  bugs recently (e.g.  https://issues.redhat.com/browse/OCPBUGS-12741) where the MCO doesn't render the same thing during bootstrap and live operation, usually due to race conditions introduced elsewhere. 

We want these sorts of issues to cause degradation rather than silently succeed so we can catch rendering issues and race conditions, but troubleshooting them hurts a lot right now because there are on-disk files we need to see that we can't. (We often have to ask reporters/customers to manually collect the files for us) 

This PR adds a gather script that:
-  collects the on-disk `currentconfig` and `mcs-machine-config-content.json` from any node the MCO has marked degraded

With this, you end up with something like this in the must-gather for nodes the MCO has labeled with `machineconfiguration.openshift.io/state: Degraded
`  : 
```
must-gather/machine_config_ondisk
must-gather/machine_config_ondisk/ip-10-0-134-89.ec2.internal
must-gather/machine_config_ondisk/ip-10-0-134-89.ec2.internal/currentconfig
must-gather/machine_config_ondisk/ip-10-0-134-89.ec2.internal/mcs-machine-config-content.json
must-gather/machine_config_ondisk/ip-10-0-137-69.ec2.internal
must-gather/machine_config_ondisk/ip-10-0-137-69.ec2.internal/currentconfig
must-gather/machine_config_ondisk/ip-10-0-137-69.ec2.internal/mcs-machine-config-content.json
```
Note: 
- If there is an easier/better place I should do this, or I'm "doing it wrong"  let me know. :smile: 
- I don't have strong opinions on the directory hierarchy, I just want the file in the gather, I'll put it wherever you think is best 

/cc @yuqi-zhang

### DIFF
--- a/collection-scripts/gather_machineconfig_ondisk
+++ b/collection-scripts/gather_machineconfig_ondisk
@@ -1,0 +1,39 @@
+#!/bin/bash 
+
+BASE_COLLECTION_PATH="must-gather"
+MACHINECONFIG_CONFIG_PATH=${OUT:-"${BASE_COLLECTION_PATH}/machine_config_ondisk"}
+
+mkdir -p "${MACHINECONFIG_CONFIG_PATH}"/
+
+# gather_machineconfig_ondisk_currentconfig collects the MCO's "on-disk" MachineConfig files
+function gather_machineconfig_ondisk {
+     
+    echo "INFO: Gathering on-disk MachineConfig from degraded nodes"
+    # we only really need the info from nodes that the MCO has marked as degraded 
+    MACHINECONFIG_DEGRADED_NODES=$(oc get node -o jsonpath='{.items[?(@.metadata.annotations.machineconfiguration\.openshift\.io/state=="Degraded")].metadata.name}')
+    # if kube-system/bootstrap is present, we're done bootstrapping
+    DONE_BOOTSTRAPPING=$(oc get configmap -n kube-system bootstrap)
+    for NODE in ${MACHINECONFIG_DEGRADED_NODES}; do
+         # use the existing MCD pod on the node to get the files
+         DAEMONPOD=$(oc get pods -n openshift-machine-config-operator --field-selector spec.nodeName=${NODE} --selector k8s-app=machine-config-daemon -o custom-columns=:metadata.name --no-headers)
+         if [ -z "$DONE_BOOTSTRAPPING" ]; then 
+            # collect the boostrap config that the Node got from the MCS (useless post-bootstrap)
+            timeout -v 3m oc cp -c machine-config-daemon openshift-machine-config-operator/"${DAEMONPOD}":rootfs/etc/mcs-machine-config-content.json "${MACHINECONFIG_CONFIG_PATH}"/"${NODE}"/mcs-machine-config-content.json &
+         else
+            # collect the Node's currentconfig that it believes it has applied (not present during bootstrap)
+            timeout -v 3m oc cp -c machine-config-daemon openshift-machine-config-operator/"${DAEMONPOD}":rootfs/etc/machine-config-daemon/currentconfig "${MACHINECONFIG_CONFIG_PATH}"/"${NODE}"/currentconfig &
+         fi
+         PIDS+=($!)
+    done
+
+}
+
+PIDS=()
+gather_machineconfig_ondisk
+
+echo "INFO: Waiting for on-disk MachineConfig collection to complete ..."
+wait "${PIDS[@]}"
+echo "INFO: on-disk MachineConfig config collection complete."
+
+# force disk flush to ensure that all data gathered is accessible in the copy container
+sync
--- a/collection-scripts/gather
+++ b/collection-scripts/gather
@@ -105,6 +105,9 @@ oc adm inspect --dest-dir must-gather --rotated-pod-logs "${all_ns_resources[@]}
 # Gather Network Observerability logs
 /usr/bin/gather_network_observability
 
+# Gather On-Disk MachineConfig files
+/usr/bin/gather_machineconfig_ondisk
+
 # Gather vSphere resources. This is NOOP on non-vSphere platform.
 /usr/bin/gather_vsphere
 


collection-scripts/gather_machineconfig_ondisk
```diff
@@ -0,0 +1,34 @@
...
+function gather_machineconfig_ondisk {
+     
+    echo "INFO: Gathering on-disk MachineConfig from degraded nodes"
+    # we only really need the info from nodes that the MCO has marked as degraded 
+    MACHINECONFIG_DEGRADED_NODES=$(oc get node -o jsonpath='{.items[?(@.metadata.annotations.machineconfiguration\.openshift\.io/state=="Degraded")].metadata.name}')
+    for NODE in ${MACHINECONFIG_DEGRADED_NODES}; do
+         # use the existing MCD pod on the node to get the files
+         DAEMONPOD=$(oc get pods -n openshift-machine-config-operator --field-selector spec.nodeName=${NODE} --selector k8s-app=machine-config-daemon -o custom-columns=:metadata.name --no-headers)
+         # collect the Node's currentconfig that it believes it has applied 
+         timeout -v 3m oc cp -c machine-config-daemon openshift-machine-config-operator/"${DAEMONPOD}":rootfs/etc/machine-config-daemon/currentconfig "${MACHINECONFIG_CONFIG_PATH}"/"${NODE}"/currentconfig &
```
yuqi-zhang, line 18: Hmm, I am looking at the codepaths, and it seems like in the failure mode we care about here (the node had a drift from the bootstrap time config) we wouldn't have written this. We should be able to access this via the gathered MachineConfig in the general case. Is there a specific scenario/use case you were thinking of for this?

yuqi-zhang: also, if this fails (timeouts), does it affect the rest of the must-gather at all or would must-gather continue and ignore it?

yuqi-zhang (commented):
I think this is generally fine (without knowing must-gather in too much detail). Some questions:

collection-scripts/gather_machineconfig_ondisk
jkyros: > Is there a specific scenario/use case you were thinking of for this?

There were multiple bugs in 1q this year that were like:

- customer's nodes are degraded and
- they deleted the rendered config while trying to fix the problem and
- the config the broken node applied caused issues but we don't know what it was

And in all those cases they "forcefiled" out and couldn't easily do root cause after the fact. ( And in my brain at least these were similar to the bootstrap cases in that they were "times when the node had a different state than the live cluster and we wanted to see the difference". ) 

I can take it out if it's contentious/the juice isn't worth the squeeze, not trying to be wasteful. I'm not *that* attached to it. I was just the one making the grocery list so I added the snacks I wanted  :smile: 

> also, if this fails (timeouts), does it affect the rest of the must-gather at all or would must-gather continue and ignore it? 

It will log that it failed but otherwise continue, the rest of the gather won't be impacted. If you want to try it, I have this one set to collect a file that doesn't exist from every node: 
```
oc adm must-gather --image quay.io/jkyros/must-gather:broken
```

you should get something like this: 
```
[must-gather-dk7s6] POD 2023-04-28T19:52:35.565600285Z INFO: Gathering on-disk MachineConfig from degraded nodes
[must-gather-dk7s6] POD 2023-04-28T19:52:35.906042203Z tar: rootfs/jkyros/nope: Cannot stat: No such file or directory
[must-gather-dk7s6] POD 2023-04-28T19:52:35.906042203Z tar: Exiting with failure status due to previous errors
[must-gather-dk7s6] POD 2023-04-28T19:52:36.001738289Z tar: rootfs/jkyros/nope: Cannot stat: No such file or directory
[must-gather-dk7s6] POD 2023-04-28T19:52:36.001738289Z tar: Exiting with failure status due to previous errors
[must-gather-dk7s6] POD 2023-04-28T19:52:36.285140754Z tar: rootfs/jkyros/nope: Cannot stat: No such file or directory
[must-gather-dk7s6] POD 2023-04-28T19:52:36.285140754Z tar: Exiting with failure status due to previous errors
[must-gather-dk7s6] POD 2023-04-28T19:52:36.383517740Z tar: rootfs/jkyros/nope: Cannot stat: No such file or directory
[must-gather-dk7s6] POD 2023-04-28T19:52:36.383517740Z tar: Exiting with failure status due to previous errors
[must-gather-dk7s6] POD 2023-04-28T19:52:36.468223813Z INFO: Waiting for on-disk MachineConfig collection to complete ...
[must-gather-dk7s6] POD 2023-04-28T19:52:36.526198943Z tar: rootfs/jkyros/nope: Cannot stat: No such file or directory
[must-gather-dk7s6] POD 2023-04-28T19:52:36.526198943Z tar: Exiting with failure status due to previous errors
[must-gather-dk7s6] POD 2023-04-28T19:52:36.647033091Z tar: rootfs/jkyros/nope: Cannot stat: No such file or directory
[must-gather-dk7s6] POD 2023-04-28T19:52:36.647033091Z tar: Exiting with failure status due to previous errors
[must-gather-dk7s6] POD 2023-04-28T19:52:36.649267872Z INFO: on-disk MachineConfig config collection complete.

```

collection-scripts/gather_machineconfig_ondisk
```diff
@@ -0,0 +1,34 @@
...
+MACHINECONFIG_CONFIG_PATH=${OUT:-"${BASE_COLLECTION_PATH}/machine_config_ondisk"}
+
+mkdir -p "${MACHINECONFIG_CONFIG_PATH}"/
+
+# gather_machineconfig_ondisk_currentconfig collects the MCO's "on-disk" MachineConfig files
+function gather_machineconfig_ondisk {
+     
+    echo "INFO: Gathering on-disk MachineConfig from degraded nodes"
+    # we only really need the info from nodes that the MCO has marked as degraded 
+    MACHINECONFIG_DEGRADED_NODES=$(oc get node -o jsonpath='{.items[?(@.metadata.annotations.machineconfiguration\.openshift\.io/state=="Degraded")].metadata.name}')
```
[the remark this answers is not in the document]
jkyros: I think...I can tell it to collect only if the `kube-system/bootstrap` configmap is absent?

I do worry slightly about making this too narrow because I can't anticipate all the bugs we'll get and what signals will be "dependable" in those scenarios.

jkyros:
Updated per Jerry's concerns such that it only collects `mcs-machine-config-content.json` during bootstrap and only collects `currentconfig` during live operation.  (I also added it to `gather` so now it actually runs :smile:)

collection-scripts/gather_machineconfig_ondisk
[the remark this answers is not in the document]
sferich888: Are we moving forward with this suggestion (is the PR still in draft)?

collection-scripts/gather_machineconfig_ondisk
```diff
@@ -0,0 +1,39 @@
...
+    # if kube-system/bootstrap is present, we're done bootstrapping
+    DONE_BOOTSTRAPPING=$(oc get configmap -n kube-system bootstrap)
+    for NODE in ${MACHINECONFIG_DEGRADED_NODES}; do
+         # use the existing MCD pod on the node to get the files
+         DAEMONPOD=$(oc get pods -n openshift-machine-config-operator --field-selector spec.nodeName=${NODE} --selector k8s-app=machine-config-daemon -o custom-columns=:metadata.name --no-headers)
+         if [ -z "$DONE_BOOTSTRAPPING" ]; then 
+            # collect the boostrap config that the Node got from the MCS (useless post-bootstrap)
+            timeout -v 3m oc cp -c machine-config-daemon openshift-machine-config-operator/"${DAEMONPOD}":rootfs/etc/mcs-machine-config-content.json "${MACHINECONFIG_CONFIG_PATH}"/"${NODE}"/mcs-machine-config-content.json &
```
sferich888, line 21: Why do we wait for up to 3 minutes for this command to timeout? Depending on how many nodes this is that could be a VERY long wait time (to denote a failure to collect across the platform).

sferich888 (changes requested):

yuqi-zhang (approved):
Thanks for the context John, I think I am fine with this, at worst we are collecting a bit more MachineConfig data that we don't need, but for some bugs this will be very handy in helping the customers, so I am fine with this as it stands.

I defer to the team maintaining must-gather to see if this is something they would like to add

sinnykumari:
@sferich888 PTAL

PR #366, opened 2023-06-18
Collect leases.coordination.k8s.io from each namesapce

### LINKED ISSUE #364: Add the leases.coordination.k8s.io as one of the collected resources
The leases resource is mainly used for Node heartbeats and Leader election, 
which both have a key role in some mechanisms of different operators, for exmple in the sriov-operator.
it makes sense to make ocp collect this as part of the must-gather.

liornoy:
This commit adds the leases resource to the list of resources to collect from all namespaces.

fixes #364


liornoy:
hi @RickJWagner can I please ask for your review on this?

PR #367, opened 2023-06-22
Assert that only one subscription exists

liornoy:
Users that mistakenly misconfigure and create two instances (or more)
of the subscription resource causes some gather scripts to fail.

For example, with 2 different sriov subscriptions, the SRIOV_NS will equal
"openshift-sriov-network-operatoropenshift-sriov-network-operator" which is invalid.

Here we add "{{\n}}" and with that, check whether we found more than
one subscription, and if so, exit with an error.

This fix is applied to all of the gather scripts that use this method:
ingress_node_firewall, metallb, network_observability, nmstate and sriov

### DIFF
--- a/collection-scripts/gather_ingress_node_firewall
+++ b/collection-scripts/gather_ingress_node_firewall
@@ -1,10 +1,15 @@
 #!/bin/bash
 BASE_COLLECTION_PATH="must-gather"
-INGRESS_NODE_FIREWALL_NS="$(oc get subs -A -o template --template '{{range .items}}{{if eq .spec.name "ingress-node-firewall"}}{{.metadata.namespace}}{{end}}{{end}}')"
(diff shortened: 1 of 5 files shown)


liornoy:
To test this commit locally you can apply 2 different metallb subs:
```
apiVersion: operators.coreos.com/v1alpha1
kind: Subscription
metadata:
  name: metallb-operator-sub1
  namespace: default
spec:
  name: metallb-operator
  channel: alpha
  source: metallb-operator
  sourceNamespace: mymetallb
  ```
  And:
  ```
apiVersion: operators.coreos.com/v1alpha1
kind: Subscription
metadata:
  name: metallb-operator-sub2
  namespace: default
spec:
  name: metallb-operator
  channel: alpha
  source: metallb-operator
  sourceNamespace: mymetallb
  ```
  And invoke: `collection-scripts/gather_metallb`
  and expect to get: `ERROR: found more than one metallb-operator subscription. Exiting.`

vwalek:
LGTM

sferich888:
@soltysh can you please review; my only nit with this is that we seem to have code duplicated in ~4-5 files (it may be time to put it in a shared utility, to possibly make this simpler to maintain).

soltysh (approved):
/lgtm
/approve

PR #372, opened 2023-07-14
Collect etcd object count

kenneth-dsouza:
Etcd health/performance can often be degraded due high object count.  In this pull request, I am adding command to collect etcd object count.  This information will help support.

### DIFF
--- a/collection-scripts/gather_etcd
+++ b/collection-scripts/gather_etcd
@@ -56,6 +56,11 @@ echo "INFO: Getting etcdctl alarm list"
 ocp4etcdctl alarm list -w json > ${ETCD_LOG_PATH}/alarm_list.json &
 PIDS+=($!)
 
+# etcd object count
+echo "INFO: Getting etcd object count"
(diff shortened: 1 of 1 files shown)


sferich888:
@tjungblu do we want to collect etcd information this way or should we collect it via metrics? 
cc: @soltysh

RickJWagner (commented):
Looks good.  Thanks for providing this information, which seems it will be useful to support.

soltysh (approved):
After discussion with etcd team, this is useful information
/approve

soltysh:
/lgtm cancel
talking with Eric about potential options here
/hold

soltysh (changes requested):
@kenneth-dsouza have a look at my comment in https://github.com/openshift/must-gather/pull/382#pullrequestreview-1649070628, where I argue about how to best address the data you're trying to add to must-gather here. Sync with the authors of that other PR, to help deliver the fix which will also address your use-case.

PR #373, opened 2023-08-02
Added gathering script for SNOs with workload partitioning

rbaturov:
This PR adds support for collecting data needed for validating SNO's with workload partitioning enabled. 

The scripts collects:
1. systemctl list-unit-files and systemctl list-unit, needed to check whether certain services are active or inactive.
2. CRIO directory to verify [workload partitioning is properly configured](https://github.com/openshift/enhancements/blob/1b69d46ae85f1e2952d4193fb77a9bb131d6907d/enhancements/workload-partitioning/management-workload-partitioning.md#cri-o-changes).

The script starts to collect these files only after checking [workload partitioning flags](https://github.com/openshift/api/blob/release-4.13/config/v1/types_infrastructure.go#L146) indicating the SNO has workload partitioning enabled.

https://issues.redhat.com/browse/TELCOSTRAT-156

### DIFF
--- a/collection-scripts/gather
+++ b/collection-scripts/gather
@@ -117,5 +117,8 @@ oc adm inspect --dest-dir must-gather --rotated-pod-logs "${all_ns_resources[@]}
 # Gather Performance profile information
 /usr/bin/gather_ppc
 
+# Gather SNO resources
+/usr/bin/gather_sno
+
 # force disk flush to ensure that all data gathered is accessible in the copy container
 sync
--- a/collection-scripts/gather_sno
+++ b/collection-scripts/gather_sno
@@ -1,0 +1,74 @@
+#!/bin/bash
+BASE_COLLECTION_PATH="must-gather"
+NODES_PATH=${OUT:-"${BASE_COLLECTION_PATH}/nodes"}
+
+function get_system_data_off_sno {
+
+    #Create a path for each node
+    mkdir -p "$NODES_PATH"/"$1"
+    
+    local DEBUG_POD_NAME_PREFIX=""
+
+    #Start Debug pod force it to stay up until removed in "default" namespace
+    oc debug --to-namespace="default" node/"$1" -- chroot /host /bin/bash -c 'sleep 300' > /dev/null 2>&1 &
+
+    #Debug pod name start with the name of the node (after removing all "." for the node name)
+    DEBUG_POD_NAME_PREFIX=${1//./}
+
+    #Mimic a normal oc call, i.e pause between two successive calls to allow pod to register
+    sleep 2
+
+    #Get the debug pod that is deployed in the default namespace on the specified node
+    debugPod=$(oc get pods --namespace=default -o jsonpath='{range .items[?(@.spec.nodeName=="'"$1"'")]}{.metadata.name}{"\n"}{end}' | grep "^${DEBUG_POD_NAME_PREFIX}-" | tail -n 1)
+
+    oc wait -n "default" --for=condition=Ready pod/"$debugPod"  --timeout=30s
+
+    if [ -z "$debugPod" ]
+    then
+      echo "Debug pod for node ""$1"" never activated"
+    else
+
+      #Collect systemctl list unit files
+      echo "Collecting list-unit-files"
+      oc exec "$debugPod" -n "default" -- chroot /host /bin/bash -c "systemctl list-unit-files" > "$NODES_PATH"/"$1"/list_units_files 
+      
+      echo "Collecting list-units"
+      oc exec "$debugPod" -n "default" -- chroot /host /bin/bash -c "systemctl list-units" > "$NODES_PATH"/"$1"/list_units 
+
+      #Collect /etc/crio directory
+      echo "Collecting /etc/crio"
+      oc cp  -n "default" "$debugPod":/host/etc/crio "$NODES_PATH"/"$1"/crio > /dev/null 2>&1
+
+      #clean up debug pod after we are done using them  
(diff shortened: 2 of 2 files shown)


RickJWagner (commented):
Looks good.  Thank you for the informative opening comment (including link to workload partitioning, etc.)

davemulford:
Echoing Rick's words about an informative opening comment. That was very helpful for me in reviewing this PR.
/lgtm

collection-scripts/gather
```diff
@@ -117,5 +117,8 @@ oc adm inspect --dest-dir must-gather --rotated-pod-logs "${all_ns_resources[@]}
 # Gather Performance profile information
 /usr/bin/gather_ppc
 
+# Gather SNO resources
+/usr/bin/gather_sno
```
sferich888, line 121: Do we want to run this all the time (even on non SNO clusters)?

collection-scripts/gather_sno
```diff
@@ -0,0 +1,74 @@
...
+NODES_PATH=${OUT:-"${BASE_COLLECTION_PATH}/nodes"}
+
+function get_system_data_off_sno {
+
+    #Create a path for each node
+    mkdir -p "$NODES_PATH"/"$1"
+    
+    local DEBUG_POD_NAME_PREFIX=""
+
+    #Start Debug pod force it to stay up until removed in "default" namespace
```
sferich888, line 12: We should not start a debug pod in the default namespace; must-gather already runs in its own namespace (and has admin privileges).

collection-scripts/gather_sno
```diff
@@ -0,0 +1,74 @@
...
+    oc debug --to-namespace="default" node/"$1" -- chroot /host /bin/bash -c 'sleep 300' > /dev/null 2>&1 &
+
+    #Debug pod name start with the name of the node (after removing all "." for the node name)
+    DEBUG_POD_NAME_PREFIX=${1//./}
+
+    #Mimic a normal oc call, i.e pause between two successive calls to allow pod to register
+    sleep 2
```
sferich888, line 19: Can't you use a wait for a condition (vs a sleep)? 

https://kubernetes.io/docs/reference/generated/kubectl/kubectl-commands#wait see `--for condition=available`

sferich888: You do this below; why is this sleep needed?

collection-scripts/gather_sno
```diff
@@ -0,0 +1,74 @@
...
+      
+      echo "Collecting list-units"
+      oc exec "$debugPod" -n "default" -- chroot /host /bin/bash -c "systemctl list-units" > "$NODES_PATH"/"$1"/list_units 
+
+      #Collect /etc/crio directory
+      echo "Collecting /etc/crio"
+      oc cp  -n "default" "$debugPod":/host/etc/crio "$NODES_PATH"/"$1"/crio > /dev/null 2>&1
+
+      #clean up debug pod after we are done using them  
+      oc delete pod "$debugPod" -n "default"
```
sferich888, line 43: This isn't needed if the pod; is started in the must-gather namespace.

collection-scripts/gather_sno
```diff
@@ -0,0 +1,74 @@
...
+      oc delete pod "$debugPod" -n "default"  
+    fi
+}
+
+function gather_system_data {
+  #Run system data collection on all nodes in parallel
+
+  mkdir -p "${NODES_PATH}"/
+
+  for NODE in ${NODES}; do
```
sferich888, line 52: How do we limit the nodes that are supplied here (to the control plane)? What happens if we have 500 nodes? How long will this collection take? 

If we limit this (do we need to collect this by default)?

sferich888 (changes requested):
I think there are a few items we want to followup on before we push for testing on this.

collection-scripts/gather_sno
rbaturov: I was following this template when writing my rule:
https://github.com/openshift/must-gather/blob/master/collection-scripts/gather_core_dumps
Could you please guide me then how to do this properly?

collection-scripts/gather_sno
rbaturov: If there was no sleep then the pod will not register on time, and the 'oc wait' command will resolve with an error.
This what will happen if we omit the "sleep 2" command:
`[rbaturov@rbaturov must-gather]$ source collection-scripts/gather_sno
INFO: Waiting for SNO system data collection to complete ...
error: arguments in resource/name form must have a single resource and name
Debug pod for node sno2.r207-sno2.r207.lab.eng.cert.redhat.com never activated
[1]+  Done                    get_system_data_off_sno "${NODE}"
INFO: SNO system data collection complete.
[rbaturov@rbaturov must-gather]$ 
`

collection-scripts/gather
rbaturov: I thought that workload partitioning is an SNO's only feature. but now I understand it being [introduced for multi-node clusters from OCP 4.13](https://cloud.redhat.com/blog/your-guide-to-workload-partitioning-for-multi-node-clusters-in-openshift-4.13). so maybe the "gather_sno" naming should be changed.
Anyhow, we want to run this script for each node, on a cluster that has workload partitioning enabled.

ingvagabund:
@davemulford looks like your lgtm accidentally added approve label as well. Also, I don't see any `revert` button here so we can quickly open a revert pr to undo the changes and resume the review.

@rbaturov given you authored the PR, it will be better if you revert the PR by hand and re-introduce the changes in a new PR so the review can resume.

PR #378, opened 2023-09-04
Use oc get daemonset to identify the NTO image

MarSik:
The previous method of using oc adm release info has been problematic for CI and restricted network environment clusters.

The issue was always the authentication to the registry and access to the release image from the inside of must-gather.

Also the oc adm release info tool runs inside a container and does not respect the node level ImageContentSourcePolicy.

For all these reasons I want to retrieve the image name from the configured NTO deployment instead.

### DIFF
--- a/collection-scripts/gather_ppc
+++ b/collection-scripts/gather_ppc
@@ -39,7 +39,7 @@ function ppc_nodes() {
 
   # Find the NTO image reference
   # NTO contains all the tools needed here
-  NTO=$(oc adm release info --image-for=cluster-node-tuning-operator)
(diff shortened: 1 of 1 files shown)


MarSik:
@sferich888 @mrniranjan Folks, this should fix the issues we are having with oc adm release info in our CI and restricted deployments. I tested it locally and it seems to work for me (quay.io/msivak/must-gather:v11).

MarSik:
You might also be interested @soltysh since you helped me with the original method.

sferich888 (approved):

PR #379, opened 2023-09-06
NO-JIRA: Update gather_metallb

yoyz:
Add a timeout in "oc exec -c frr -- vtysh -c ${COMMAND}" This command can block if vtysh is not responding properly and can put customer into a situation where mustgather can not be grabbed.

### DIFF
--- a/collection-scripts/gather_metallb
+++ b/collection-scripts/gather_metallb
@@ -24,7 +24,7 @@ function gather_frr_logs() {
 
     for COMMAND in "${COMMANDS[@]}"; do
         echo "###### ${COMMAND}" >> ${LOGS_DIR}/dump_frr
-        echo "$( timeout -v 60s oc -n ${operator_ns} exec ${1} -c frr -- vtysh -c "${COMMAND}")" >> ${LOGS_DIR}/dump_frr
+        echo "$( timeout -v 20s oc -n ${operator_ns} exec ${1} -c frr -- vtysh -c "${COMMAND}")" >> ${LOGS_DIR}/dump_frr
     done
 }
 
--- a/collection-scripts/gather_metallb
+++ b/collection-scripts/gather_metallb
@@ -24,7 +24,7 @@ function gather_frr_logs() {
 
     for COMMAND in "${COMMANDS[@]}"; do
         echo "###### ${COMMAND}" >> ${LOGS_DIR}/dump_frr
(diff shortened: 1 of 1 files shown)


yoyz:
Might be interesting to redirect STDERR as well !

RickJWagner (approved):
Approved.

yoyz:
@davemulford could you review it ?

collection-scripts/gather_metallb
```diff
@@ -24,7 +24,7 @@ function gather_frr_logs() {
 
     for COMMAND in "${COMMANDS[@]}"; do
         echo "###### ${COMMAND}" >> ${LOGS_DIR}/dump_frr
-        echo "$( oc -n ${operator_ns} exec ${1} -c frr -- vtysh -c "${COMMAND}")" >> ${LOGS_DIR}/dump_frr
+        echo "$( timeout -v 60s oc -n ${operator_ns} exec ${1} -c frr -- vtysh -c "${COMMAND}")" >> ${LOGS_DIR}/dump_frr
```
sferich888, line 27: can this timeout be reduced to 10,20 or 30 seconds? 60 seconds is a long time. What is the typical runtime for a command like this? really what is the typical 'error identification time' for a command like this?

sferich888 (changes requested):

collection-scripts/gather_metallb
yoyz: On a 6 year old server running 4xVM of OCP 4.10 with cpufreq decrease at it's minimum frequency we got the following figures : 

[root@helper ~]# declare -a COMMANDS=("show running-config" "show bgp ipv4" "show bgp ipv6" "show bgp neighbor" "show bfd peer")
[root@helper ~]# for COMMAND in "${COMMANDS[@]}"; do time oc -n metallb-system exec speaker-27wtk  -c frr -- vtysh -c "show running-config" ; done 2>&1 |grep ^real 
real    0m1.462s
real    0m1.135s
real    0m0.751s
real    0m0.685s
real    0m0.842s

[root@helper ~]# for COMMAND in "${COMMANDS[@]}"; do time oc -n metallb-system exec speaker-27wtk  -c frr -- vtysh -c "show running-config" ; done 2>&1 |grep ^real 
real    0m1.021s
real    0m1.450s
real    0m1.034s
real    0m0.593s
real    0m1.112s

60s is high, but would not decrease it to 10s here for being sure to capture the output in a really overloaded system.
20s or 30s should be totally fine.

After 20s of "nothing" display, I'm already convinced we won't have any output. And all vtysh are stuck inside the pod.

PR #380, opened 2023-09-13
machineconfig: always gather MCS-served config

yuqi-zhang:
The current logic collects either the MCS served config or currentconfig for installation failures. This doesn't really help since the currentconfig gets written by a firstboot service, and contains only a very small encapsulated MachineConfig generally speaking.

Change the script to always collect the MCS served config since this helps with rendered-master-xxx not found bugs, which otherwise we can't easily get.

### DIFF
--- a/collection-scripts/gather_machineconfig_ondisk
+++ b/collection-scripts/gather_machineconfig_ondisk
@@ -11,18 +11,11 @@ function gather_machineconfig_ondisk {
     echo "INFO: Gathering on-disk MachineConfig from degraded nodes"
     # we only really need the info from nodes that the MCO has marked as degraded 
     MACHINECONFIG_DEGRADED_NODES=$(oc get node -o jsonpath='{.items[?(@.metadata.annotations.machineconfiguration\.openshift\.io/state=="Degraded")].metadata.name}')
-    # if kube-system/bootstrap is present, we're done bootstrapping
-    DONE_BOOTSTRAPPING=$(oc get configmap -n kube-system bootstrap)
     for NODE in ${MACHINECONFIG_DEGRADED_NODES}; do
          # use the existing MCD pod on the node to get the files
          DAEMONPOD=$(oc get pods -n openshift-machine-config-operator --field-selector spec.nodeName=${NODE} --selector k8s-app=machine-config-daemon -o custom-columns=:metadata.name --no-headers)
-         if [ -z "$DONE_BOOTSTRAPPING" ]; then 
-            # collect the boostrap config that the Node got from the MCS (useless post-bootstrap)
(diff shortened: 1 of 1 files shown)


yuqi-zhang:
cc @jkyros

yuqi-zhang:
/hold

I think... this doesn't always get us what we want either, will need to think about what determines "MCDaemon never finished"

yuqi-zhang:
/hold cancel

After discussing with John I think the value of this is higher than currentconfig, since if the user deleted a rendered-config, we don't really need to know what the contents are to recover the node (forcefile) vs install time drifts, which cannot be easily directly rectified.

RickJWagner (commented):
Looks good.  The script seems to only pull MCS served config, never current config, as discussed by John and Yu Qi in comments.

collection-scripts/gather_machineconfig_ondisk
```diff
@@ -11,18 +11,11 @@ function gather_machineconfig_ondisk {
...
-         if [ -z "$DONE_BOOTSTRAPPING" ]; then 
-            # collect the boostrap config that the Node got from the MCS (useless post-bootstrap)
-            timeout -v 3m oc cp -c machine-config-daemon openshift-machine-config-operator/"${DAEMONPOD}":rootfs/etc/mcs-machine-config-content.json "${MACHINECONFIG_CONFIG_PATH}"/"${NODE}"/mcs-machine-config-content.json &
-         else
-            # collect the Node's currentconfig that it believes it has applied (not present during bootstrap)
-            timeout -v 3m oc cp -c machine-config-daemon openshift-machine-config-operator/"${DAEMONPOD}":rootfs/etc/machine-config-daemon/currentconfig "${MACHINECONFIG_CONFIG_PATH}"/"${NODE}"/currentconfig &
-         fi
+         # collect the boostrap config that the Node got from the MCS
+         timeout -v 3m oc cp -c machine-config-daemon openshift-machine-config-operator/"${DAEMONPOD}":rootfs/etc/mcs-machine-config-content.json "${MACHINECONFIG_CONFIG_PATH}"/"${NODE}"/mcs-machine-config-content.json &
```
sferich888, line 18: can this really take ~3 minutes to collect? Can we shorten this timeout to 30s? we are simply copying a file (how big it this file) what size is this going to add to the archive?

sferich888 (changes requested):

collection-scripts/gather_machineconfig_ondisk
yuqi-zhang: Shorted to ~1 min to account for slowness.

This file is essentially the size of 1 rendered machineconfig object, and only collected when the node is degraded, so it should add negligible size increase (a rendered machineconfig is around 100Kb in size)

PR #382, opened 2023-09-18
Collect running VMs count

assafad:
If Openshift Virtualization operator is installed, collect the number of running virtual machines in the cluster, across all namespaces. 

Jira-ticket: https://issues.redhat.com/browse/CNV-32855

/cc @sradco, @soltysh

### DIFF
--- a/collection-scripts/gather
+++ b/collection-scripts/gather
@@ -117,5 +117,8 @@ oc adm inspect --dest-dir must-gather --rotated-pod-logs "${all_ns_resources[@]}
 # Gather Performance profile information
 /usr/bin/gather_ppc
 
+# Gather Openshift Virtualization information
+/usr/bin/gather_virtualization
+
 # force disk flush to ensure that all data gathered is accessible in the copy container
 sync
--- a/collection-scripts/gather_virtualization
+++ b/collection-scripts/gather_virtualization
@@ -1,0 +1,13 @@
+#!/bin/bash
+
+source $(dirname "$0")/common.sh
+BASE_COLLECTION_PATH="must-gather"
+get_operator_ns "kubevirt-hyperconverged"
+VIRTUALIZATION_PATH="${BASE_COLLECTION_PATH}/virtualization"
+mkdir -p "${VIRTUALIZATION_PATH}"
+
+function collect_running_vms_count() {
+  oc get vmi --all-namespaces -o=jsonpath='{.items[?(@.status.phase=="Running")].metadata.name}' | wc -w > "${VIRTUALIZATION_PATH}/running_vms_count.txt"
+}
+
+collect_running_vms_count


collection-scripts/gather_running_vms
```diff
@@ -0,0 +1,13 @@
+#!/bin/bash
+
+source $(dirname "$0")/common.sh
+BASE_COLLECTION_PATH="must-gather"
+get_operator_ns "kubevirt-hyperconverged"
+OPENSHIFT_CNV_PATH="${BASE_COLLECTION_PATH}/namespaces/${operator_ns}"
+
+function collect_running_vms_count() {
+  mkdir -p "${OPENSHIFT_CNV_PATH}"
+  oc get vmi --all-namespaces --no-headers | wc -l > "${OPENSHIFT_CNV_PATH}/running_vms_count.txt"
```
assafad, line 10: @nunnatsa, @sradco Do you think it's misleading to put this under `openshift-cnv` dir? 
As it's gathered from `--all-namespaces`. 
We can put it under `cluster-scoped-resources` dir instead, but that's also misleading IMO, since VMs are not cluster scoped.

collection-scripts/gather
```diff
@@ -117,5 +117,8 @@ oc adm inspect --dest-dir must-gather --rotated-pod-logs "${all_ns_resources[@]}
 # Gather Performance profile information
 /usr/bin/gather_ppc
 
+# Gather Number of running VMs
+/usr/bin/gather_running_vms
```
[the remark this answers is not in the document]
dankenigsberg: no, at least not now. I would like to add minimal computation effort on the main must-gather effort. Anything more complex should go to the cnv-specific image.

dankenigsberg, line 121:
```suggestion
/usr/bin/gather_running_vms
```
this comment is redundant. The name of the function should tell the reader everything she needs to know about what the function is going to do. Let us avoid future comment rot.

collection-scripts/gather_running_vms
dankenigsberg: I know that we collect and report a metric named `cnv:vmi_status_running:count`  about currently-running VMs. Can we pull it instead of listing all vmis on the must-gather pod and counting them?

assafad: @sradco mentioned that we can't assume Prometheus is installed and available, and in order to get `cnv:vmi_status_running:count` value, we need to query Prometheus, so I'm not sure it's possible.

collection-scripts/gather
assafad: I don't mind removing it, I just tried the follow the pattern in this [file](https://github.com/openshift/must-gather/blob/197b7c6051deca66c76ecddc69c6b726d0d0f092/collection-scripts/gather#L67), which has comments above every execution of a script.

collection-scripts/gather_running_vms
sradco: @dankenigsberg @assafad  Actually, We can try to collect this from Prometheus, but I think using the oc command would be more reliable. OCP MG does have a a collection for Prometheus alerts (That I would like to add to CNV MG) and we can try to get the metric from it, But it can potentially be unreachable for different reasons.
Do you think it would be "good enough"? What would be the advantage?

collection-scripts/gather
dankenigsberg: I don't like many of those comments too, but I was not their reviewer...

collection-scripts/gather_running_vms
dankenigsberg: > Do you think it would be "good enough"? What would be the advantage?

The code we are adding here would be executed by every OpenShift customer looking for support, if they have Virt installed or not. I would like to ensure we burden them as little as possible. I suspected that collecting a single metric would be faster, simpler and more reliable than listing all VMs. I accept your verdict if this is not the case.

assafad: Just mentioning that the command is skipped when Virt is not installed, using:
 https://github.com/openshift/must-gather/blob/197b7c6051deca66c76ecddc69c6b726d0d0f092/collection-scripts/common.sh#L11

collection-scripts/gather_virtualization
```diff
@@ -0,0 +1,13 @@
+#!/bin/bash
+
+source $(dirname "$0")/common.sh
+BASE_COLLECTION_PATH="must-gather"
+get_operator_ns "kubevirt-hyperconverged"
+VIRTUALIZATION_PATH="${BASE_COLLECTION_PATH}/virtualization"
+mkdir -p "${VIRTUALIZATION_PATH}"
+
+function collect_running_vms_count() {
+  oc get vmi --all-namespaces -o=jsonpath='{.items[?(@.status.phase=="Running")].metadata.name}' | wc -w > "${VIRTUALIZATION_PATH}/running_vms_count.txt"
```
sferich888, line 10: must-gather is primarily for gathering diagnostics information; if you looking to send a metric back to Red Hat; you should use telemetry. 

In speaking with @djfjeff about this; the CNV team already has telemetry reporting for this and the purpose of this gather data (in this fashion) is to account for disconnected cluster too. With that being the case this data should be collected; via `registry.redhat.io/container-native-virtualization/cnv-must-gather-rhel9` 

However as I understand it the concern here (with including this collection in the cnv must-gather image) is this isn't always collected (by support or customers) so we don't have all the data we want (all the time). 

However with the work happening in https://issues.redhat.com/browse/CNV-29411 this data collection would duplicate what is already in the cnv must-gather image; so I am not in favor of including this (and am in favor of waiting for https://issues.redhat.com/browse/CNV-29411 to complete the offical solution).

sferich888: Note if https://github.com/kubevirt/must-gather/pull/191/ also merges; when a user collects both an OCP must-gather and a CNV must-gather you will spend time and data (size of the archive) collecting duplicate information. 

Its not a good ideal to implement collections in this manner.

sradco: Hi, 

I understand that there could be duplication here, but this is a extremely important metric for us that we need to be collected at this time also from OCP for grater exposure. 
Once https://issues.redhat.com/browse/CNV-29411 will be implemented and we can relay on the SRE team to ask for the all the MGs, I promise we will remove this duplication.
I can add a story that will depend on https://issues.redhat.com/browse/CNV-29411 to remove it for tracking.

sradco: @soltysh Please review.

sferich888: @sradco  I am not arguing the importance of the metric (to be clear; I am all for gathering all the metrics/data we can). What I am objecting to is the manner[s] in which the metric is being collected. 

Again; must-gather archives are a 'diagnostics tool'; a 'count' doesn't assist with diagnostics. It provides business insights. While we can (and do) gather 'business insights' from our diagnostics reports (to improve the product); these data points are extrapolated from the 'diagnostics information'. Ie: If you collect all the virtual machine instances (vmi's) object references as configurations a count can be extrapolated from that (for business insights purposes), but the data is also useful for 'diagnostics purposes'. 

Because this data is already collected: https://github.com/kubevirt/must-gather/blob/main/collection-scripts/gather_vms_details I don't see a strong justification to duplicate it. 

- Note: an argument to 'collect it in the primary must-gather image' is because the CNV image isn't always collected (and this requires additional back and forth); which can be true - again  https://issues.redhat.com/browse/CNV-29411 aims to address that problem (and given its focused on that problem it should be allowed to solve that problem (not this PR). 
    - Note: you can collect multiple archives at once with must-gather today; so the 'back and forth' issue while a valid usability issue; is manageable to day with training and education (on the SRE/Support/Customer side of the collection conversation). At least that is my opinion.

soltysh (changes requested):
/hold

After a lengthy discussion with Eric about the value of this PR vs the cost of maintaining it we both agree that the best possible option would be to figure out a way to dump the telemetry reported data (based on the discussion in this PR we already collect that information there) . There's also a similar PR (https://github.com/openshift/must-gather/pull/372) which adds something similar, but which could be solved with this dump. 

@assafad @sradco do you think you could implement that approach rather than the above one?

sradco:
> After a lengthy discussion with Eric about the value of this PR vs the cost of maintaining it we both agree that the best possible option would be to figure out a way to dump the telemetry reported data (based on the discussion in this PR we already collect that information there) . There's also a similar PR (#372) which adds something similar, but which could be solved with this dump.

Hi @soltysh, We don't already collect this information.

assafad:
After discussing the correct way to collect this information, it was decided to query Prometheus and collect the query result to MG. Closing in favor of https://github.com/openshift/must-gather/pull/384.

PR #384, opened 2023-10-03
Add Prometheus queries collection

assafad:
This PR enhances `gather_monitoring` script to query Prometheus for requested queries, and adds these queries results to the collection. The suggested structure for these results in the MG dump is as follows:
```
-- monitoring 
---- queries
------ <query_description>.txt
```

**Example:** 
This PR adds the first query to be collected - `sum(cnv:vmi_status_running:count)`, and its result is collected to `monitoring/queries/running_vms_count.txt` as:
```
Query: sum(cnv:vmi_status_running:count)
{
  "status": "success",
  "data": {
    "resultType": "vector",
    "result": [
      {
        "metric": {},
        "value": [
          1696335577.967,
          "2"
        ]
      }
    ]
  }
}
```

**Notes:**
- `/monitoring` dir is not an addition of this PR.
- The script adds a `Query:  <query>` headline to each file. This is important because when we request a custom query (and not a metric/recording rule), the query is not part of the output. 
- The script prints the full output, and not only values, since queries results can be vectors with many different values, based on labels. Thus, the full output is useful.

Jira-ticket: https://issues.redhat.com/browse/CNV-33642

/cc @soltysh, @sradco, @dankenigsberg

### DIFF
--- a/collection-scripts/gather_monitoring
+++ b/collection-scripts/gather_monitoring
@@ -104,6 +104,23 @@ alertmanager_get() {
       2> "$result_path.stderr"
 }
 
+prom_query() {
+  local query="$1"; shift
+  local path="${1:-$query}"; shift || true
+
+  local result_path="$MONITORING_PATH/queries/$path"
+  mkdir -p "$(dirname "$result_path")"
+
+  query_result="$(oc exec "${PROM_PODS[0]}" \
+    -c prometheus \
+    -n openshift-monitoring \
+    -- curl --silent --data-urlencode \
+    "query=${query}" \
+    http://localhost:9090/api/v1/query? | jq .)"
+
+  echo -e "Query: ${query}\n${query_result}" > "${result_path}.txt"
+}
+
 
 monitoring_gather(){
   init
@@ -124,6 +141,8 @@ monitoring_gather(){
   prom_get_from_replicas status/tsdb  || true
 
   alertmanager_get status  || true
+
+  prom_query "sum(cnv:vmi_status_running:count)" running_vms_count || true
 
   # force disk flush to ensure that all data gathered are written
   sync


collection-scripts/gather_monitoring
```diff
@@ -125,6 +142,8 @@ monitoring_gather(){
 
   alertmanager_get status  || true
 
+  prom_query "sum(cnv:vmi_status_running:count)" running_vms_count || true
```
sradco, line 145: We might want to add a prefix like:
cnv_running_vms_count 
so it will be clear who created the file and if we add additional metrics they will be sorted together.

Can you please show an example of how the metric will be printed in the file?

assafad: Other example than the one in the description?

soltysh (changes requested):
I was more thinking about a full metrics dump rather than this, something like `curl -sG http://localhost:9090/metrics` but I don't have access to a cluster with virtual machines so I can't verify if the data you want is available there. I'd still encourage to reach out to folks from telemetry to ask them how you could provide a dump of the data they are gathering to be send further. 

What you're doing here isn't much different from #382, the only difference is where you gather the data from. Whereas my request was to gather a broader set of metrics to address more issues with a single object dump.

assafad:
> I was more thinking about a full metrics dump rather than this, something like `curl -sG http://localhost:9090/metrics` but I don't have access to a cluster with virtual machines so I can't verify if the data you want is available there. I'd still encourage to reach out to folks from telemetry to ask them how you could provide a dump of the data they are gathering to be send further.
> 
> What you're doing here isn't much different from #382, the only difference is where you gather the data from. Whereas my request was to gather a broader set of metrics to address more issues with a single object dump.

@soltysh `oc exec -n openshift-monitoring prometheus-k8s-0 -c prometheus -- curl http://127.0.0.1:9090/metrics ` doesn't give all the metrics in the cluster, IIUC only metrics that are exposed by the Prometheus pod itself. I'm not aware of a command that can give all the metrics in cluster, but I will search. 

And I think this PR is much different than #382, as it doesn't add additional scripts to be maintained.

assafad:
@soltysh Thinking about it, dumping all the metrics would not be a useful solution, and will not answer the requirements. This kind of solution will basically dump all metrics values at a certain point of time, and thus we will lose the ability to use these metrics for gaining insights we may need. For example, counter metrics are usually used in max/median/avg across periods of time queries, as they can show trends. Their granular values at specific points of time are usually not very useful. 
In addition, I don’t think there is a way to get “all” metrics in the cluster, as each pod (in our case operator) expose its metrics, so we’ll need to look for multiple metrics endpoints, and this brings complexity of identifying which layered products are installed and where they expose their metrics. 

@sradco @machadovilaca WDYT?

machine424:
Hello,


FYI, as I mentioned in https://issues.redhat.com/browse/MON-3108?focusedId=23008065&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-23008065, we're considering using, via must-gather, the promtool binary on the Prometheus container images to dump some chosen metrics and re-load them on another Prometheus.

With promtool, we hope to have something easier to implement than https://github.com/openshift/must-gather/pull/214. and more generic than what is proposed in this PR.

Here is a PoC https://github.com/openshift/must-gather/pull/389, I'll share the proposal soon for discussion.

sferich888:
I think this should be closed in favor of https://github.com/openshift/must-gather/pull/389

sradco:
Hi  @sferich888 @soltysh, After talking with  @machine424 It seems that there is no expected timeline for #389.
Therefore, since this is a high priority metric us, I would appreciate it if you can accept this PR until there is a better approach. Once there is, we will make the relevant adjustments and remove the metric.

sradco:
@machine424 Please review this PR

PR #386, opened 2023-10-05
ppc: explicitly add RuntimeClass to gathered resources

jlojosnegros:
PerformanceProfile creates a RuntimeClass.
It would be nice to have that in the must-gather information so lets explicitly gather that information in the script


jlojosnegros:
Hi @davemulford and @sferich888 
Please take a look whenever you have some spare time.

Thanks in advance.

PR #392, opened 2023-11-07
NO-JIRA: Add ARO support to must-gather client

cadenmarchese:
Azure Red Hat OpenShift (ARO) contains its own operator and custom resource, as well as a few additional namespaces on top of the base OCP. If accepted, this PR would collect this data if present or skip if not present, so that ARO customers can collect must-gathers that will help MCS and SRE debug ARO-specific issues.

Currently, this change is available in a custom must-gather image at `quay.io/cmarches/aro-must-gather`, but I wanted to propose adding the change upstream since ARO is an official OCP flavor and not all Azure customers will be able to pull a custom image from quay for compliance reasons.

### DIFF
--- a/collection-scripts/gather_aro
+++ b/collection-scripts/gather_aro
@@ -1,0 +1,22 @@
+#!/bin/bash
+BASE_COLLECTION_PATH="must-gather"
+ARO_CR_NAME="clusters.aro.openshift.io"
+
+IS_ARO=$(oc get ${ARO_CR_NAME} --ignore-not-found=true )
+if [ -z "$IS_ARO" ]; then
+    exit 0
+fi
+
+function gather_aro() {
+    echo "INFO: Collecting ARO Cluster Data"
+    declare -a INSPECT_TARGETS=(${ARO_CR_NAME} "ns/openshift-azure-operator" "ns/openshift-azure-logging")
+
+    for TARGET in "${INSPECT_TARGETS[@]}"; do
+        oc adm inspect --dest-dir "${BASE_COLLECTION_PATH}" "${TARGET}"
+    done
+}
+
+gather_aro
+
+# force disk flush to ensure that all data gathered are written
+sync
--- a/collection-scripts/gather
+++ b/collection-scripts/gather
@@ -114,5 +114,8 @@ oc adm inspect --dest-dir must-gather --rotated-pod-logs "${all_ns_resources[@]}
 # Gather Performance profile information
 /usr/bin/gather_ppc
 
+# Gather ARO information
+/usr/bin/gather_aro
+
 # force disk flush to ensure that all data gathered is accessible in the copy container
 sync


RickJWagner (commented):
Scripting looks ok.  How does the Quay image relate to this PR, though?

cadenmarchese:
@RickJWagner the quay.io image in the description is just a custom must-gather image built with these changes. I've updated the description to hopefully be more clear about that.

sferich888:
@soltysh @ingvagabund I think this needs your review; I don't see any issues with this; but this may be something best/better suited for its own collection/collector image (your call; I would `/lgtm` this). 

cc: @davemulford

ingvagabund:
@sferich888 what percentage of clusters that are scraped are ARO? How much MBs/KBs get scraped with this change?

cadenmarchese:
@ingvagabund I can't speak to which percentage of OCP clusters are ARO, but in my testing with an ARO cluster, a regular must-gather collects about 903.9 MB of data (uncompressed), whereas the must-gather with this change in pace collects about 904.9 MB (uncompressed). In the case of a non-ARO cluster, there's no additional ARO data to collect, so this change wouldn't impact the size of the must-gather.

It's certainly possible to collect inspect data for the two NSs by hand, but it might not be as nice of a customer experience to have to run three commands instead of one, especially for Azure customers who aren't usually familiar with these collection scripts. 

If anyone's interested in trying out this change, you can run `oc adm must-gather --image quay.io/cmarches/aro-must-gather:20231030.00`

ingvagabund:
> whereas the must-gather with this change in pace collects about 904.9 MB (uncompressed).

This will vary on the length of container logs.

In general it's ok to use the generic must-gather to collect the minimum data about a specific feature and use corresponding custom must-gather to collect more detailed data. I'd say let's accept this change and observe how much data is collected in the wild. We can always revert/modify the change in case too much data is pulled in addition.

ingvagabund:
@cadenmarchese is there a jira card tracking this work item?

ingvagabund:
/retitle NO-JIRA: Add ARO support to must-gather client

sferich888:
/lgtm 
cc: @davemulford  hope your ok adding this.

collection-scripts/gather_aro
```diff
@@ -0,0 +1,22 @@
...
+
+    for TARGET in "${INSPECT_TARGETS[@]}"; do
+        oc adm inspect --dest-dir "${BASE_COLLECTION_PATH}" "${TARGET}"
+    done
+}
+
+gather_aro
+
+# force disk flush to ensure that all data gathered are written
+sync
```
davemulford, line 22: Do we need a sync here since one already happens in the main gather script?

davemulford:
Other than the one question about having a sync in the ARO script, I'm okay with this change.

PR #400, opened 2024-01-08
WRKLDS-950: use must-gather since and since time env vars on inspect cmds

flavianmissi:


collection-scripts/gather
```diff
@@ -51,7 +51,7 @@ all_ns_resources+=(leases)
 
 # Run the Collection of Resources using inspect
 # running across all-namespaces for the few "Autoscaler" resources.
-oc adm inspect --dest-dir must-gather --rotated-pod-logs "${named_resources[@]}" &
+oc adm inspect --since="${MUST_GATHER_SINCE}" --since-time="${MUST_GATHER_SINCE_TIME}" --dest-dir must-gather --rotated-pod-logs "${named_resources[@]}" &
```
ardaguclu, line 54: Should we handle the cases where `MUST_GATHER_SINCE` or `MUST_GATHER_SINCE_TIME` are unset such as?;

```sh
oc adm inspect --since= --since-time= --dest-dir must-gather
```

flavianmissi: When `--since` and `--since-time` are unset, `oc adm must-gather` will still set both envs with Go's empty value for each type, so if the client doesn't specify either flag they will be `0s` and an empty string respectively. My manual testing show that this works.
This approach isn't the most clear though... WDYT?

ardaguclu: Yeah, I agree that this isn't clear. Besides if there is any change in defaults, these commands would fail. I think, we can only pass these flags when they are set.

flavianmissi: Latest push addresses your comment @ardaguclu, PTAL.

collection-scripts/common.sh
```diff
@@ -15,3 +15,28 @@ function get_operator_ns() {
         exit 1
     fi
 }
+
+get_log_collection_args() {
+	LOG_COLLECTION_ARGS=""
+
+	if ! [ -z "${MUST_GATHER_SINCE+x}" ]; then
+		LOG_COLLECTION_ARGS=--since="${MUST_GATHER_SINCE}"
```
ardaguclu, line 23: Since must-gather command in oc is the caller and validation is done here, we don't have to validate that MUST_GATHER_SINCE and MUST_GATHER_SINCE_TIME are both set.

ardaguclu (commented):
I have a question, other than that it looks good to me. Thanks.

collection-scripts/common.sh
ardaguclu: Just a note for the future reviewers :)

flavianmissi: Ah yes, that is a good note to make. Added a comment to the code (will push soon)

collection-scripts/gather_service_logs_util
```diff
@@ -16,7 +16,7 @@ function collect_service_logs {
         DIR_PATH="${SERVICE_LOG_PATH}/masters"
     elif [[ ${selector} == '--role=worker' ]]; then
         DIR_PATH="${SERVICE_LOG_PATH}/workers"
-        echo "WARNING: Collecting one or more service logs on ALL linux $1 workers in your cluster. This could take a large amount of time." >&2
+        echo "WARNING: Collecting one or more service logs on ALL linux $2 workers in your cluster. This could take a large amount of time." >&2
```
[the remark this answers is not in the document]
flavianmissi: huh I must have fat fingered this, good catch - rolling it back to $1.

ardaguclu:
LGTM. I'm deferring to @ingvagabund for another set of eyes.

flavianmissi:
something is still off, holding until I figure it out
/hold

flavianmissi:
Checked that `--since-time` works for both node logs (under `host_service_logs`) and pod logs (under `namespaces/.../pod/.../logs/current.log`.
`--since` is a little more complicated - when a duration is given from via the cli as `2h`, after calling `duration.String()` it is formatted as `2h0m0s` and `oc adm node-logs` does takes a duration matching the `(+|-)[0-9]+(s|m|h|d)` regex. I'm going to have to truncate the duration string to keep its left most part, so in the case of `2h0m0s` only `2h` would be given as an argument to `oc adm node-logs`.

flavianmissi:
fixed formatted of duration given to `oc adm node-logs`.
also using lowercase now for the non-env-vars, which seem to be the best practice.
will be pushing the updates in a moment.

flavianmissi:
Both flags now work as they should! 
/hold cancel

flavianmissi:
@ardaguclu this one is ready to go in!

ardaguclu:
Thanks @flavianmissi 
/lgtm

collection-scripts/gather_audit_logs
```diff
@@ -36,15 +40,15 @@ paths=(openshift-apiserver kube-apiserver oauth-apiserver etcd oauth-server)
...
+  oc adm node-logs ${node_log_collection_args:+"$node_log_collection_args"} --role=master --path="$path" | \
   tee "${BASE_COLLECTION_PATH}/audit_logs/$path.audit_logs_listing" | \
   grep -v ".terminating" | \
   grep -v ".lock" | \
   sed "s|^|$path $output_dir |"
 done | \
 xargs --max-args=4 --max-procs=45 bash -c \
   'echo "INFO: Started  downloading $1/$4 from $3";
-  oc adm node-logs $3 --path=$1/$4 | gzip > $2/$3-$4.gz;
+  oc adm node-logs ${node_log_collection_args} $3 --path=$1/$4 | gzip > $2/$3-$4.gz;
```
flavianmissi, line 51: this doesn't actually work - `xargs` does not have access to env vars, so `node_log_collection_args` will always be empty here. right now it doesn't make a difference because the `--since` flag only affects jounal logs for `oc adm node-logs`. 
I'm fixing it anyway so we're future proof.

flavianmissi: last push fixes this - @ardaguclu can you take another look? This is the last one, I promise 😅

ardaguclu: No worries :). Tagged.

soltysh (approved):
/approve

PR #404, opened 2024-02-20
OPNET-470: Collect host networking logs

mkowalski:
It happens very often that must-gather is the only bundle provided to the networking team when asking to help debugging various issues. In reality, in majority of those cases we would rather get sosreport and not must-gather because the former contains useful outputs from `ip` command.

With this PR we are trying to get the most basic networking informations already into must-gather to speed up debugging.

Contributes-to: [OPNET-470](https://issues.redhat.com//browse/OPNET-470)
Contributes-to: OCPBUGS-26217
Contributes-to: OCPBUGS-29624

### DIFF
--- a/collection-scripts/gather_host_network_logs
+++ b/collection-scripts/gather_host_network_logs
@@ -1,0 +1,36 @@
+#!/bin/bash
+
+source $(dirname "$0")/common.sh
+get_log_collection_args
+. gather_service_logs_util
+
+BASE_COLLECTION_PATH="/must-gather"
+HOST_NETWORK_NODE_LOGS=$BASE_COLLECTION_PATH/host_network_logs
+
+mkdir -p "${HOST_NETWORK_NODE_LOGS}"/
+
+PIDS=()
+echo "INFO: Collecting host networking logs"
+
+collect_service_logs --role=master "on-prem-resolv-prepender"
+collect_service_logs --role=worker "on-prem-resolv-prepender"
+
+collect_service_logs --role=master "nodeip-configuration"
+collect_service_logs --role=worker "nodeip-configuration"
+
+CLUSTER_NODES="${@:-$(oc get node -l node-role.kubernetes.io/master -oname)}"
+for NODE in ${CLUSTER_NODES}; do
+    nodefilename=$(echo "$NODE" | sed -e 's|node/||')
+    out=$(oc debug "${NODE}" -- \
+    /bin/bash -c "ip address" 2>/dev/null) && echo "$out" 1> "${HOST_NETWORK_NODE_LOGS}/ip-address-$nodefilename.log"
+    out=$(oc debug "${NODE}" -- \
+    /bin/bash -c "ip -4 route" 2>/dev/null) && echo "$out" 1> "${HOST_NETWORK_NODE_LOGS}/ip-route4-$nodefilename.log"
+    out=$(oc debug "${NODE}" -- \
+    /bin/bash -c "ip -6 route" 2>/dev/null) && echo "$out" 1> "${HOST_NETWORK_NODE_LOGS}/ip-route6-$nodefilename.log"
+done
+
+wait ${PIDS[@]}
+echo "INFO: Host networking log collection complete."
+
+# force disk flush to ensure that all data gathered is accessible in the copy container
+sync
--- a/collection-scripts/gather
+++ b/collection-scripts/gather
@@ -112,6 +112,10 @@ pids+=($!)
 /usr/bin/gather_network_logs_basics &
 pids+=($!)
 
+# Gather host networking logs
+/usr/bin/gather_host_network_logs &
+pids+=($!)
+
 # Gather metallb logs
 /usr/bin/gather_metallb &
 pids+=($!)


mkowalski:
/cc @cybertron 

/cc @dgoodwin @stbenjam 
This is a closure on the last issues we were debugging. If we had this from the very beginning (i.e. inside must-gather and not only on the live system), we would have saved a bit of time

RickJWagner (commented):
Looks good.

collection-scripts/gather_host_network_logs
```diff
@@ -0,0 +1,36 @@
...
+BASE_COLLECTION_PATH="/must-gather"
+HOST_NETWORK_NODE_LOGS=$BASE_COLLECTION_PATH/host_network_logs
+
+mkdir -p "${HOST_NETWORK_NODE_LOGS}"/
+
+PIDS=()
+echo "INFO: Collecting host networking logs"
+
+collect_service_logs --role=master "on-prem-resolv-prepender"
+collect_service_logs --role=worker "on-prem-resolv-prepender"
```
sferich888, line 16: Gathering logs from the workers (when there are 500 nodes) might be a problem is there any way to limit this ?

collection-scripts/gather_host_network_logs
```diff
@@ -0,0 +1,36 @@
...
+
+collect_service_logs --role=master "nodeip-configuration"
+collect_service_logs --role=worker "nodeip-configuration"
```
sferich888, line 19: Gathering logs from the workers (when there are 500 nodes) might be a problem is there any way to limit this ?

collection-scripts/gather_host_network_logs
```diff
@@ -0,0 +1,36 @@
...
+
+CLUSTER_NODES="${@:-$(oc get node -l node-role.kubernetes.io/master -oname)}"
```
sferich888, line 21: Gathering logs from the workers (when there are 500 nodes) might be a problem is there any way to limit this ?

sferich888, line 21:
```suggestion
CLUSTER_NODES="${@:-$(oc get node -l node-role.kubernetes.io/master -o name)}"
```

sferich888: Can we make it so that this is something a user pass in (the list of nodes to collect this data from)?

sferich888 (changes requested):

collection-scripts/gather_host_network_logs
mkowalski: I feel like this would defeat the purpose. Maybe I never defined the problem we have correctly - what happens today is that people with "some networking problem" (as they feel) come to our team with the most basic must-gather and ask for help. At this moment we start playing ping-pong, because our first reply is "please bring sosreport from all the nodes". Only based on sosreport we can say if the problem is really on us, or not at all. If yes, we ask for the next set of logs (networkmanager with `level=TRACE`).

With this PR I aimed to at least bring must-gather and sosreport a bit closer, so that customer coming with must-gather collected using this PR would receive as a next instruction "yes, it's networking, please bring sosreport and NM trace logs" or "no, it's not networking" already.

If now we make it configurable, I do not believe any customer will ever come with a correct must-gather in the first place (as in, they will first give us must-gather without logs from any worker what renders useless).

collection-scripts/gather_host_network_logs
mkowalski: Are clusters with 500 nodes a real thing? From which perspective it may be a problem, are we now talking about the `time oc must-gather [...]` or about number of files collected?

collection-scripts/gather_host_network_logs
mkowalski: This service should run only once per boot. No matter how many nodes customer has, unless they reboot once-per-minute, we don't expect it to be a big log bundle

collection-scripts/gather_host_network_logs
sferich888: @mkowalski we need to watch for both (we don't want our collection tool to take too long to run or to collect GB's of data; that are impossible to review/process).

collection-scripts/gather_host_network_logs
sferich888: Yes but we still have to reachout to N nodes; and copy files off of them (from the service); that process takes both time and space; and that is the worry here.

collection-scripts/gather_host_network_logs
sferich888: I think we need to time box this (you only get N (<5) minutes to collect all this (or this script exits), or find some way to potentially limit or target this collection at a list of nodes.  

IMO; the exposure risk of trying to collect all of this on hundreds of nodes; add too much to the collection and/or takes too much time (so we have to address that).

mkowalski: As a trade-off I am okay to limit this only to master nodes, is this acceptable for you? It would still put us in a better situation than we are today and if after some months we still feel that what support&customers bring us does not make sense, we can reiterate on this discussion

sferich888:
@mkowalski are you still working this issue? Have you had a chance to look at my suggestions/recommendations and are we able to limit this in some way? 

@ingvagabund this is another example where your comments (in https://github.com/openshift/must-gather/pull/416#issuecomment-2306584908) apply to more diligence in what we collect/pull in.

PR #413, opened 2024-04-09
MON-3754: Enable metrics-server audit log collection

slashpai:
audit-logs are enabled by default on metrics-server in openshift/cluster-monitoring-operator#2280.

This change is to enable collection of metrics-server audit-log in must-gather.

cc: @simonpasquier @jan--f


slashpai:
Is this good to merge?

PR #414, opened 2024-04-11
NO-JIRA: Correct typo.

RickJWagner:
README has a sentence that doesn't read correctly, this PR corrects it.


ardaguclu:
@sferich888 could you PTAL?. Thanks.

PR #416, opened 2024-04-12
OTA-1177: Gather OSUS data

oarribas:
Collect data from OSUS operator if installed in the cluster.

### DIFF
--- a/collection-scripts/gather_osus
+++ b/collection-scripts/gather_osus
@@ -1,0 +1,26 @@
+#!/bin/bash
+
+source $(dirname "$0")/common.sh
+BASE_COLLECTION_PATH="must-gather"
+OSUS_OPERATOR_NAME="update-service-operator"
+get_log_collection_args
+
+HAS_OSUS=$(oc get csv -A --no-headers -o custom-columns=NS:.metadata.namespace,OPERATOR:.metadata.name --ignore-not-found=true | awk '/'${OSUS_OPERATOR_NAME}'/ {print $1}')
+if [ -z "$HAS_OSUS" ]; then
+    exit 0
+fi
+
+function gather_osus() {
+    echo "INFO: Collecting OSUS Data"
+    declare -a OSUS_NAMESPACES=(${HAS_OSUS})
+
+    for OSUS_NAMESPACE in "${OSUS_NAMESPACES[@]}"; do
+        oc adm inspect ${log_collection_args} --dest-dir "${BASE_COLLECTION_PATH}" "ns/${OSUS_NAMESPACE}"
+        oc adm inspect --dest-dir "${BASE_COLLECTION_PATH}" updateservices --namespace "${OSUS_NAMESPACE}"
+    done
+}
+
+gather_osus
+
+# force disk flush to ensure that all data gathered are written
+sync
--- a/collection-scripts/gather
+++ b/collection-scripts/gather
@@ -140,6 +140,10 @@ pids+=($!)
 /usr/bin/gather_ppc &
 pids+=($!)
 
+# Gather OSUS information
+/usr/bin/gather_osus &
+pids+=($!)
+
 # Check if PID array has any values, if so, wait for them to finish
 if [ ${#pids[@]} -ne 0 ]; then
     echo "Waiting on subprocesses to finish execution."


collection-scripts/gather_osus
```diff
@@ -0,0 +1,26 @@
+#!/bin/bash
+
+source $(dirname "$0")/common.sh
+BASE_COLLECTION_PATH="must-gather"
+OSUS_OPERATOR_NAME="update-service-operator"
+get_log_collection_args
+
+HAS_OSUS=$(oc get csv -A --no-headers -o custom-columns=NS:.metadata.namespace,OPERATOR:.metadata.name --ignore-not-found=true | awk '/'${OSUS_OPERATOR_NAME}'/ {print $1}')
```
wking, line 8: This seems like a hard direction for an OCP-central must-gather to support.  I thought the OLM-installed operator pattern was for each operator to ship a separate image with the must-gather-for-them logic (as described in [the KCS][1] you'd linked from [OTA-1177](https://issues.redhat.com/browse/OTA-1177)), so the central must-gather maintainers didn't have to be bothered reviewing gather logic for all the many, many, OLM-installed operators that could possibly be present?

[1]: https://access.redhat.com/solutions/5459251

oarribas: @wking , while I agree with your comment for most of the operators that can be installed in OpenShift, I think creating a full image for this operator is excessive.
Other operators provide extra capacities to the cluster, and usually involve several different CRDs and even several namespaces. This operator helps with one of the core capacities of OpenShift, which is the upgrade (in this case, for disconnected clusters), and lots of cases already includes a must-gather proactively. The info for this operator (when installed) shouldn't increase a lot the size of the must-gather, and will avoid asking for a new different must-gather.

oarribas: @wking , any thoughts on the above?

sferich888: @oarribas / @wking the only thing this collects are the details from a given namespace[s]. I think this is fine (given its related to update); but @soltysh or @ingvagabund should have the final say if this is provided by this image or an independent image.

ingvagabund: @oarribas how big are the extra manifests? Can you share an example of running both commands?

oarribas: @ingvagabund , checking data from some cases, it depends a lot on the volume of the logs from the pods.
Largest inspect of the namespace I have seen, without compression is 30MB, and compressed between 2-3MB. The size of the `updateservice` resource is few KB.

oarribas: @ingvagabund , @soltysh , any thoughts based on the above?

ingvagabund: For some time we have been merging extra collection with a promise of collecting extra "small" data that help to avoid asking customers to run yet another must gather image. More relevant in the disconnected environment. Yet, we don't track how much of these extra collections increase the overall must-gather size on average. Before going further I'd like to see when all the extra collections are collected. E.g. what's the wort case extra bump in the collected data. A new section under https://github.com/openshift/must-gather/blob/master/README.md will do. E.g.

## Extra collections

| script location | short description       | condition | estimated size |
| ------------ | ----------------- | ----------- | ------ |
| /usr/bin/gather_aro | Gather ARO Cluster Data | `ns/openshift-azure-operator` or `ns/openshift-azure-logging` present | ?? |
| /usr/bin/gather_vsphere | Gather vSphere resources | vSphere CSI driver is installed | ?? |

oarribas: @ingvagabund , the estimated size for this one (for the OSUS operator), when compressed, is around 3MB (usually less as per several inspect of the `openshift-update-service` namespace I have reviewed). And it's only collected if OSUS is installed.

ingvagabund:
@kasturinarra just for curiosity do we have any test case/jobs monitoring how much an average must-gather image grows in time in context of various installations?

@oarribas are there any statistics about must-gather size? E.g. a matrix of which operators are installed -> how much data can be collected. Or, what are the variable parts that can significantly increase the size?

@sferich888 is it possible to make a matrix of all flavors of an OCP cluster? Including layered products? To see how complex a must-gather gathering can be? I am quite blind in here. Would like to extend my perspective so we can make better decisions when reviewing this kind of additions.

oarribas:
@ingvagabund , checking in OTA-1177

sferich888:
@ingvagabund by my count (before we add on layered products) the matrix your looking at is has 87k+ combinations in it. 

```
>>> versions = ['4.12', '4.13', '4.14', '4.15', '4.16']
>>> IaaS_providers = ['Alibaba', 'AWS', 'Azure', 'Azure Stack Hub', 'GCP', 'IBM', 'Nutanix', 'BareMetal', 'OpenStack', 'Vsphere', 'OCI']
>>> Install_Method = ['IPI', 'UPI', 'Assisted Installer']
>>> Install_Mode = ['Connected', 'Disconnected']
>>> Deployment_Pattern = ['SingleNode', 'SingleNode+', '3C2W', '3C3I2W', '3CW', '3CW3I']
>>> ### C = Control Plane, W = Worker, I = Infrastructure, += Added workers
>>> Arch = ['x86_64', 'S390', 'ARM', 'Power']
>>> 
>>> from itertools import product
>>> 
>>> m_lists = [versions, IaaS_providers, Install_Method, Install_Mode, Deployment_Pattern, Arch, IaaS_providers]
>>> cp = list(product(*m_lists))
>>> len(cp)
87120
```

However when it come to must-gather and testing are we building a tool that works for the majority of our user base; I think the more important thing to consider is/are only about 9k of those combinations (or 10% of that matrix).

The biggest issues I have seen are related to operating at specific sizes and scales! IE: with our Deployment Patterns (combinations). We see the biggest challenges when must-gather can't find a host to run on (SingleNode Clusters or Clusters with scheduleable control planes (that are loaded with work), or has to crowd out a workload to start (people really don't like this; but its necessary). Or when we try and operate at large scales (500+ nodes; with workloads).

The biggest issues we see are with 'time to collect' data, and with how much data we collect (Note - we don't automatically compress archives (RFE for this; that hasn't been auctioned yet) - so we probably shouldn't make collection estimates based on compression). The size of our 'archive' is an issue for most customers; because they have to, in a lot of situations move the data from one system to another, just so that they can upload it to Red Hat, that is 2+ data transfers for many customers (mostly customers in disconnected or restricted network environments). Pared with the time to collect a must-gather (20+ min in some situations), we could have a customer collecting and transferring data for up to 30 to 40 min (based on some estimates).

ingvagabund:
@sferich888 `IaaS_providers` is mentioned twice in `m_lists`. Is that on purpose?

PR #417, opened 2024-04-15
OCPBUGS-32210: Collect Assisted Installer Namespace Logs

cptmorgan-rh:
Adds the namespace for the Assisted Installer to collect the Assisted Installer Operator logs when a must-gather is run.


RickJWagner (commented):
/lgtm

PR #418, opened 2024-04-29
NO-JIRA: Use oc rhel8 binary explicitly until base image is rhel9

ardaguclu:
must-gather includes oc invocations to gather some data and must-gather gets this oc from cli image. However, when we try to bump cli image to RHEL9, since must-gather is still on RHEL8, all oc invocations failed. This PR changes to use cli-artifacts image instead of cli. Thanks to that change, must-gather can explicitly uses oc.rhel8 and that decouples the cli and must-gather images for the future similar migrations.

example failures;

*  https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_oc/1752/pull-ci-openshift-oc-master-e2e-aws-ovn/1784837264291926016
* https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_oc/1752/pull-ci-openshift-oc-master-e2e-agnostic-ovn-cmd/1784837260940677120

PR #419, opened 2024-05-01
NO-JIRA: copy oc binary from cli-artifacts

jeffdyoung:
additional work for https://github.com/openshift/must-gather/pull/418 to copy multi-arch oc binaries.

### DIFF
--- a/Dockerfile.rhel7
+++ b/Dockerfile.rhel7
@@ -7,5 +7,6 @@ FROM registry.ci.openshift.org/ocp/4.16:cli-artifacts as cli-artifacts
 
 FROM registry.ci.openshift.org/ocp/4.16:cli
 COPY --from=builder /go/src/github.com/openshift/must-gather/collection-scripts/* /usr/bin/
-COPY --from=cli-artifacts /usr/share/openshift/linux_amd64/oc.rhel8 /usr/bin/oc
+COPY --from=cli-artifacts /usr/bin/oc /usr/bin/oc
+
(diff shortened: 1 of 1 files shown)


Dockerfile.rhel7
```diff
@@ -7,5 +7,6 @@ FROM registry.ci.openshift.org/ocp/4.16:cli-artifacts as cli-artifacts
 
 FROM registry.ci.openshift.org/ocp/4.16:cli
 COPY --from=builder /go/src/github.com/openshift/must-gather/collection-scripts/* /usr/bin/
-COPY --from=cli-artifacts /usr/share/openshift/linux_amd64/oc.rhel8 /usr/bin/oc
+COPY --from=cli-artifacts /usr/bin/oc /usr/bin/oc
```
jeffdyoung, line 10: @sferich888 we don't have to copy a specific oc arch here. The cli-artifacts container has the rhel8 variant of oc in /usr/bin/oc. Since the build system builds on native HW, we will always get the right binary arch.

https://github.com/openshift/oc/blob/master/images/cli-artifacts/Dockerfile.rhel#L15-L21

ardaguclu: We are trying to change the default oc's base image to RHEL9 and default oc in cli-artifacts will also not work on must-gather.
/hold

ardaguclu: Proposed this https://github.com/openshift/must-gather/pull/420 instead

jeffdyoung:
closing in favor of https://github.com/openshift/must-gather/pull/420

PR #420, opened 2024-05-02
WRKLDS-1251: Use oc.rhel8 based on the target arch of image

ardaguclu:
This PR uses the correct oc.rhel8 binary based on the image's architecture. This is the follow up of https://github.com/openshift/must-gather/pull/418


sferich888 (approved):

jaypoulz:
Should this have included s390x as well?

PR #424, opened 2024-05-27
OCPBUGS-34360: Run ppc node collection in parallel

MarSik:
The PPC node data collection was running in serial and that increased the time needed to collect the whole cluster. This change executes each per-node sequence in parallel and waits for all the sub tasks.

The only disadvantage is having a per-node process (potentially hundreds in huge clusters), however this case has always been lower in the code as well. The load on the controlling must-gather pod should be low as it only executes remote commands and waits for results.

This can be tested (at least while the PR is open) by executing:

```
oc adm must-gather --image quay.io/msivak/must-gather:test416c
```

Thanks to our scale testing colleagues we have a comparison between the stock and the parallelized runs using OCP 4.16.0-rc.3 on a cluster with 120 worker nodes.

stock:

```
$ time oc adm must-gather
real    48m39.426s
user    0m30.998s
sys     0m14.527s
file size:5969MB
```

with PR applied:

```
$ time oc adm must-gather --image quay.io/msivak/must-gather:test416c
[must-gather      ] OUT 2024-06-06T00:55:35.965950772Z Using must-gather plug-in image: quay.io/msivak/must-gather:test416c
real    8m16.947s
user    0m34.009s
sys     0m16.852s
file size:6993MB (I checked, the size is fine, one extra hour of run time generated some extra logs on the cluster)
```

### DIFF
--- a/collection-scripts/gather_ppc
+++ b/collection-scripts/gather_ppc
@@ -139,6 +139,7 @@ EOF
   done
 
   COLLECTABLE_NODES=()
+  NODE_PIDS=()
   for line in $(oc get pod -o=custom-columns=NODE:.spec.nodeName,NAME:.metadata.name --no-headers -l name=perf-node-gather-daemonset --field-selector=status.phase=Running -n $NAMESPACE)
   do
       node=$(echo $line | awk -F ' ' '{print $1}')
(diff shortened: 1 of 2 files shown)


MarSik:
@ffromani Even the simple change makes quite a difference.

RickJWagner (commented):
Looks good.

MarSik:
@sferich888 @davemulford Folks, who can provide the last missing approve here?

PR #425, opened 2024-05-29
OCPBUGS-33959: Network logs collection: skip unready nodes

martinkennelly:
Previous to this change, debug pods will be launched to node that are unready and deployment will not complete.
Also update control plane label selector.

/cc
~/hold~

### DIFF
--- a/collection-scripts/gather_network_logs_basics
+++ b/collection-scripts/gather_network_logs_basics
@@ -19,8 +19,9 @@ function gather_multus_data {
   done
 }
 
-CLUSTER_NODES="${@:-$(oc get node -l node-role.kubernetes.io/master -oname)}"
-/usr/bin/gather_multus_logs $CLUSTER_NODES
+READY_CONTROL_PLANE_NODES="${@:-$(oc get node -l node-role.kubernetes.io/control-plane -o json | jq -r '.items[] | select(.status.conditions[] | select(.type=="Ready" and .status=="True")).metadata.name')}"
+
+/usr/bin/gather_multus_logs $READY_CONTROL_PLANE_NODES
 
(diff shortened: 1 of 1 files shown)


martinkennelly:
/unhold

testing complete

martinkennelly:
@soltysh PTAL - fix speeds up data collection for MGs when there `NotReady` CP nodes.

soltysh (approved):
/lgtm
/approve

collection-scripts/gather_network_logs_basics
```diff
@@ -19,8 +19,9 @@ function gather_multus_data {
   done
 }
 
-CLUSTER_NODES="${@:-$(oc get node -l node-role.kubernetes.io/master -oname)}"
-/usr/bin/gather_multus_logs $CLUSTER_NODES
+READY_CONTROL_PLANE_NODES="${@:-$(oc get node -l node-role.kubernetes.io/control-plane -o json | jq -r '.items[] | select(.status.conditions[] | select(.type=="Ready" and .status=="True")).metadata.name')}"
```
sferich888, line 22: We may want to do this filtering with just oc (to reduce dependencies on jq). 
That said jq is used in other parts of must-gather - https://github.com/openshift/must-gather/blob/3a90e581047a71a74fd9babb1ada3ad5f54c7a25/collection-scripts/gather_apirequestcounts#L11 

cc: @soltysh

vwalek: @sferich888 unfortunately, `jsonpath` nor `template` can list the nodes based on their readiness. I was working on something similar - using the `jq` is way simpler and better.

PR #428, opened 2024-06-17
OCPBUGS-34323: Collection-scripts as go package

dfajmon:
Related jira ticket https://issues.redhat.com/browse/OCPBUGS-34323

[Local-storage-operator](https://github.com/openshift/local-storage-operator/tree/master/must-gather) and [secrets-store-csi-driver-operator](https://github.com/openshift/secrets-store-csi-driver-operator/tree/main/must-gather) must-gather plugins now have a copy  of [common.sh from must-gather repo](https://github.com/openshift/must-gather/blob/master/collection-scripts/common.sh). If there is a bugfix in that script, we must copy it manually. 

We decided the most optimal way for us to have the script up to date is using go vendoring and making the collection-scripts as a go package. This means we need to add go file, marking it as a package. 

We are already using this trick in [Build Machinery Go](https://github.com/openshift/cluster-storage-operator/tree/master/vendor/github.com/openshift/build-machinery-go) where we add it to go.mod and import it in dir [Dependency magnet](https://github.com/openshift/cluster-storage-operator/blob/master/pkg/dependencymagnet/dependencymagnet.go) and the scripts (in this example makefiles) are downloaded.

### DIFF
--- a/collection-scripts/doc.go
+++ b/collection-scripts/doc.go
@@ -1,0 +1,3 @@
+// required for gomod to pull in packages.
+
+package alpha_must_gather


soltysh (changes requested):
/hold

The general guidelines for building your own must-gather images, like in this case is to base your image on the default must-gather one. This way you have all the scripts already available. 

Also based on what I was looking on the usage, there's only a single function that is currently updated rarely, so this approach seems a bit like an overkill. 

I'm open to revisit this once the size of the deps will be bigger or there will be more consumers. For now I'm against this.

dfajmon:
we decided to use must-gather image instead.

PR #429, opened 2024-06-18
SDN-5066:Remove needless multus log collection

martinkennelly:
Multus is now a Pod and will be captured by normal oc adm must-gather command.

The multus.log file is removed since 4.16 and doesn't exist anymore.

/hold

/cc 

@dougbtv


martinkennelly:
Holding until I get multus team sign-off first.

dougbtv:
This is indeed the correct course of action to take here. The gist is that the architecture of Multus CNI changed (in 4.14) to what we call a "thick plugin" architecture. Previously, Multus CNI, in operation, is a process that runs on the host, a one-shot process from a binary executed on the host (we call this a "thin plugin" architecture). Now, there is both that process, but additionally a daemon that runs in a Kubernetes pod. This pod has the log output, so the previous log collection is obsolete.

Thanks Martin!

bpickard22:
going to attempt to close and re-open to get the approve label to apply

bpickard22:
ahh @martinkennelly we need someone from the mg owners to approve, @dougbtv and I cannot give that label

sferich888 (commented):
Is this still needed @bpickard22

PR #430, opened 2024-07-01
OCPBUGS-34360: Run ppc node collection in parallel

MarSik:
The PPC node data collection was running in serial and that increased the time needed to collect the whole cluster. This change executes each per-node sequence in parallel and waits for all the sub tasks.

The only disadvantage is having a per-node process (potentially hundreds in huge clusters), however this case has always been in the code as well. The load on the controlling must-gather pod should be low as it only executes remote commands and waits for results.

This can be tested (at least while the PR is open) by executing:

oc adm must-gather --image quay.io/msivak/must-gather:test416c

### DIFF
--- a/collection-scripts/gather_ppc
+++ b/collection-scripts/gather_ppc
@@ -136,6 +136,7 @@ EOF
   done
 
   COLLECTABLE_NODES=()
+  NODE_PIDS=()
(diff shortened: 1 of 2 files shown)


sferich888 (approved):
/lgtm

yanirq:
@MarSik @sferich888  who should add the cherry-pick label here ?  this is long standing

PR #434, opened 2024-07-10
MON-3108: Add gather_metrics for Prometheus fetching

machine424:
This introduces a new `gather_metrics` script. The script leverages the new `promtool tsdb dump-openmetrics` command to extract metrics in OpenMetrics format.

The script accepts some arguments, run:
  $ ./collection-scripts/gather_metrics
for more details.

The dumpoed metrics can be imported into a Prometheus instance using:
  $ promtool tsdb create-blocks-from openmetrics
For more information, refer to https://prometheus.io/docs/prometheus/latest/command-line/promtool/#promtool-tsdb-create-blocks-from

Note that in some cases, using `increase()` or `{i}rate()` functions on the imported data may return negative or inaccurate results. This issue is under investigation and will be fixed upstream.

Recommendations:

- Import these metrics into a blank Prometheus instance.
- Ensure that the major version of the target Prometheus instance matches the major version of the dumped instance.
- If the involved metrics are old, consider adjusting the --storage.tsdb.retention.time setting in the host instance to prevent their removal (especially if the receiving Prometheus is to configured to scrape new targets)

---

the next step would be to add some out-of-the box metrics in `gather_monitoring`, see https://github.com/openshift/must-gather/pull/389

### DIFF
--- a/collection-scripts/gather_monitoring
+++ b/collection-scripts/gather_monitoring
@@ -10,6 +10,7 @@ declare -r BASE_COLLECTION_PATH="../must-gather"
 declare -r MONITORING_PATH="${BASE_COLLECTION_PATH}/monitoring"
 declare -r CA_BUNDLE="${MONITORING_PATH}/ca-bundle.crt"
 
+source $(dirname "$0")/monitoring_common.sh
 
 # init initializes global variables that need to be computed.
 # E.g. get token of the default ServiceAccount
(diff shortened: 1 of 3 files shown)


collection-scripts/gather_monitoring
RickJWagner, line 1: What's the purpose of leaving the TODO and commented line here?
Is it a reminder?

machine424: As I mentioned in the description, https://github.com/openshift/must-gather/pull/389 is the second step.
It's kind of a placeholder, I can remove that if you want.

RickJWagner: No, it's fine.  I was just making sure it shouldn't be 'live'.
Thanks.

PR #436, opened 2024-08-01
CNF-13837: gather_ppc: fallback option when NTO image not found

Tal-or:
gather_ppc script is looking for NTO image, becuase it contains all the tools that needed.

On Hypershift platfrom, the control-plane components are located on a different cluster, so the script can't find NTO image since the controller is running on the remote (management) cluster.

This commit suggests to add a fallback method:
In case NTO deployment was not found, the script tries to get NTO image name from the release image.

Signed-off-by: Talor Itzhak <titzhak@redhat.com>

### DIFF
--- a/collection-scripts/gather_ppc
+++ b/collection-scripts/gather_ppc
@@ -44,9 +44,13 @@ function ppc_nodes() {
   # NTO contains all the tools needed here
   NTO=$(oc get deployment -n openshift-cluster-node-tuning-operator cluster-node-tuning-operator -o jsonpath="{.spec.template.spec.containers[0].image}")
   if [ $? -ne 0 ]; then
-    echo "ERROR: Failed to identify the container image with node tools."
-    echo "INFO: Node performance data collection will not contain node level data."
-    return
+    echo "INFO: Fallback to identify the container image from release info"
+    NTO=$(oc adm release info --image-for=cluster-node-tuning-operator)
+    if [ $? -ne 0 ]; then
+      echo "ERROR: Failed to identify the container image with node tools."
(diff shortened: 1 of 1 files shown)


collection-scripts/gather_ppc
```diff
@@ -45,8 +45,12 @@ function ppc_nodes() {
   NTO=$(oc get deployment -n openshift-cluster-node-tuning-operator cluster-node-tuning-operator -o jsonpath="{.spec.template.spec.containers[0].image}")
   if [ $? -ne 0 ]; then
     echo "ERROR: Failed to identify the container image with node tools."
-    echo "INFO: Node performance data collection will not contain node level data."
-    return
+    echo "INFO: Fallback to identify the container image from release info"
+    NTO=$(oc adm release info --image-for=cluster-node-tuning-operator)
+    if [ $? -ne 0 ]; then
+      echo "INFO: Node performance data collection will not contain node level data."
```
MarSik, line 51: You should probably log the second error here as well so we can tell from the logs which method passed / failed.

MarSik: Or just move the ERROR here

MarSik:
Apart from a tiny comment related to logging this is /lgtm

collection-scripts/gather_ppc
Tal-or: Done, moved the error inside the nested `if`

MarSik:
@sferich888 @RickJWagner Could you please take a look folks?

RickJWagner (commented):
/lgtm

PR #440, opened 2024-08-27
OCPBUGS-10957: avoid monitoring routes and TLS

jan--f:


simonpasquier (commented):
/lgtm

zhouying7780:
@jan--f  Since we have remove the define for 'CA_BUNDLE', we also need to remove it from `cleanup()` , because when I tested I hit : /usr/bin/gather_monitoring: line 26: CA_BUNDLE: unbound variable

jan--f:
> @jan--f Since we have remove the define for 'CA_BUNDLE', we also need to remove it from `cleanup()` , because when I tested I hit : /usr/bin/gather_monitoring: line 26: CA_BUNDLE: unbound variable

Fixed.

PR #460, opened 2024-11-01
[HOLD] Add olmv1 operators/v1 OLM resource to must-gather

perdasilva:
Adds openshift api operators OLM resource to must-gather

### DIFF
--- a/collection-scripts/gather_olm_v1
+++ b/collection-scripts/gather_olm_v1
@@ -1,0 +1,6 @@
+#!/usr/bin/env bash
+
+source $(dirname "$0")/common.sh
+get_log_collection_args
+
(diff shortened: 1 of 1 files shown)


davemulford:
Is this a new resource that doesn't exist on clusters? I get the following error when trying the `oc adm inspect` on a 4.17 cluster.

```text
$ oc adm inspect -A olms.operator.openshift.io
error: the server doesn't have a resource type "olms"
```

cc @sferich888

perdasilva:
> Is this a new resource that doesn't exist on clusters? I get the following error when trying the `oc adm inspect` on a 4.17 cluster.
> 
> ```
> $ oc adm inspect -A olms.operator.openshift.io
> error: the server doesn't have a resource type "olms"
> ```
> 
> cc @sferich888

Yeah, it is. We're working through getting it promoted now =D

PR #464, opened 2024-11-15
OCPBUGS-44264: Collect all clustercsidriver resources

mpatlasov:
ClusterCSIDriver objects are small, let's collect them all. We expect that there 4 ClusterCSIDriver instances in a cluster at max (e.g. Azure Disk, Azure File, SMB and SecretStore) and the instances are small.

Gathering these objects is important because we already gather Deployment, DaemonSets and other objects of CSI drivers by generic must-gather, and ClusterCSIDriver objects complement those info with managementState, cipherSuites, and various condition statuses for each CSI driver.

### DIFF
--- a/collection-scripts/gather
+++ b/collection-scripts/gather
@@ -37,7 +37,7 @@ group_resources+=(nodes)
 named_resources+=(ns/default ns/openshift ns/kube-system ns/openshift-etcd)
 
 # Storage Resources
-group_resources+=(storageclasses persistentvolumes volumeattachments csidrivers csinodes volumesnapshotclasses volumesnapshotcontents)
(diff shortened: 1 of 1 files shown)


jsafrane:
/lgtm
from storage POV. We find it useful to capture also ClusterCSIDriver instances that are handled by OLM-based operators, such as AWS EFS, GCP Filestore or Secret Store, without needing operator based must-gather images (such as aws-efs-csi-driver-operator-must-gather). A generic must-gather already contains Deployment, DaemonSets and other objects of these OLM based CSI drivers.

We expect that there 4 ClusterCSIDriver instances  in a cluster at max (e.g. Azure Disk, Azure File, SMB and SecretStore) and the instances are small.

mpatlasov:
@RickJWagner can you approve this PR, please.

davemulford:
@sferich888, can you take a look at this? Also, we need to remove Rick from the list of approvers.

sferich888 (approved):

PR #468, opened 2024-12-12
NO-ISSUE: gather additional MCO debug artifacts

djoshy:
This PR gathers a few debug artifacts pertaining to the machine-config-daemon implemented by openshift/machine-config-operator#4723 and openshift/machine-config-operator#4731. It also gathers the MachineConfiguration object, which is an singleton operator level object used to configure the MCO.

### DIFF
--- a/collection-scripts/gather_machineconfigdaemon_termination_logs
+++ b/collection-scripts/gather_machineconfigdaemon_termination_logs
@@ -1,0 +1,38 @@
+#!/bin/bash 
+
+BASE_COLLECTION_PATH="must-gather"
+MACHINECONFIG_DAEMON_LOG_PATH=${OUT:-"${BASE_COLLECTION_PATH}/machine_config_termination_logs"}
+
+mkdir -p "${MACHINECONFIG_DAEMON_LOG_PATH}"/
+
+# gather_machineconfigdaemon_termination_logs collects the daemon logs before it was last terminated
+function gather_machineconfigdaemon_termination_logs {
+     
+    echo "INFO: Gathering machine config daemon's old logs from all nodes"
+    NODES=$(oc get nodes -o jsonpath='{.items[?(@.kind=="Node")].metadata.name}')
+    for NODE in ${NODES}; do
+         # use the existing MCD pod on the node to get the files
+         DAEMONPOD=$(oc get pods -n openshift-machine-config-operator --field-selector spec.nodeName=${NODE} --selector k8s-app=machine-config-daemon -o custom-columns=:metadata.name --no-headers)
+         # Check if the 'previous-logs' folder exists on the MCD pod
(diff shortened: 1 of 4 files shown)


collection-scripts/gather_machineconfigdaemon_termination_logs
```diff
@@ -0,0 +1,38 @@
...
+PIDS=()
+gather_machineconfigdaemon_termination_logs
+
+echo "INFO: Waiting for Machine Config Daemon termination log collection to complete ..."
+wait "${PIDS[@]}"
+echo "INFO: Machine Config Daemon termination log collection complete."
+
+# force disk flush to ensure that all data gathered is accessible in the copy container
+sync
+
```
sferich888, line 38: This fill will not execute by default is that OK?

collection-scripts/gather_machineconfigdaemon_termination_logs
```diff
@@ -0,0 +1,38 @@
...
+           PIDS+=($!)
+         else
+           echo "INFO: 'previous-logs' folder not found on ${NODE}, skipping..."
+         fi
+    done
+
+}
+
+PIDS=()
+gather_machineconfigdaemon_termination_logs
```
sferich888, line 30: Shouldn't these logs be collected by collecting the log resources related to the pods/deployments/daemonsets (issued by oc inspect upon them)?

sferich888 (changes requested):
I think these logs should already be collected or improvements to collect them should be worked into 'oc inspect'

collection-scripts/gather_machineconfigdaemon_termination_logs
djoshy: Unfortunately, no. These daemonsets get recreated when the MCO rolls out an update so we tend to lose all logs related to those pods. This work involves an MCO side change which preserves the last set of daemon logs on the node to disk when the node is shutdown. We've had many customer cases where there are no "previous" logs because at the time of the must gather collection (typically post upgrade failure), the old daemonset no longer exists.

collection-scripts/gather_machineconfigdaemon_termination_logs
djoshy: Sorry, could you clarify what you mean by fill? 🙂

sferich888: Will typo from my phone. In short this script has to be called explicitly not implicitly by invoking a must-gather collection with this image.

djoshy: Ah, I see! Will update the main script to include this one. Thanks!

djoshy: Updated to explicitly call from the main script! Let me know if I'm still missing something!

PR #488, opened 2025-04-29
RFE-7515: Add support for gathering node resources requests, limits, ovn subnets and resource consumption

marioabajo:
we tried to mimic what we get from an "oc describe nodes", but specifically the resources section, and "oc adm top nodes" to better understand the node resource usage and what the node reports about those pods.


marioabajo:
Closing as new PR with changes already requested: #496

PR #489, opened 2025-05-22
WIP: Add fio benchmark suite to gather_etcd

natamm4:
WIP testing PR to trigger CI and test a Must-Gather image w/ fio integration for etcd diagnostics.

Please ignore for now -- not intended for review/merging.

PR #498, opened 2025-07-16
ETCD-738: Partial Revert of #489

natamm4:
Addressing comments concerning breaking API changes from #489.


machine424:
/lgtm
validated with https://github.com/openshift/origin/pull/30054/files
thanks!

machine424:
I think this needs to be attached to a JIRA ticket.
Do you want me to attach it to one of mine?

PR #502, opened 2025-07-30
add loki logs

avlitman:
Add loki logs collection to must gather by default and also add --only-loki to run only loki logs collection.

the script return logs by batch per log type, each batch is 5000 logs. the default is 100 batches (0.5 mil logs per log type) and each 250k logs are in a separate file. 

Jira-ticket: https://issues.redhat.com/browse/CNV-59376

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,3 +35,23 @@ oc adm must-gather --image=registry/username/your-built-image
 
 ## Obfuscate confidential information
 There is a dedicated effort to obfuscate and omit confidential information. Head over to [openshift/must-gather-clean](https://github.com/openshift/must-gather-clean) for more information.
+
+## Loki Log Collection
+
+### Options
+- `--only-loki` - Collect only Loki logs (skip traditional must-gather)
+- `--logs-namespace=NAME` - Target specific namespace
+- `LOKI_MAX_BATCHES=N` - Batches per log type (default: 100, each batch contain 5000 logs)
+
+### Output Structure
+```
+loki-logs/
+├── application_logs_part1.txt     (250K records)
+├── application_logs_part2.txt     (remaining records)
+├── infrastructure_logs_part1.txt  (250K records)
+├── audit_logs_part1.txt           (250K records)
+└── metadata/
+    ├── clusterlogging.yaml
+    ├── loki_routes.yaml
+    └── loki_pods.yaml
+```
--- a/collection-scripts/gather
+++ b/collection-scripts/gather
@@ -8,8 +8,37 @@ version >> /must-gather/version
 source $(dirname "$0")/common.sh
 get_log_collection_args
 
+# Parse command line arguments for special flags
+LOKI_ONLY=false
+while [[ $# -gt 0 ]]; do
+    case $1 in
+        --only-loki)
+            LOKI_ONLY=true
+            export LOKI_ONLY_MODE=true
+            shift
+            ;;
+        --logs-namespace=*)
+            export LOKI_LOGS_NAMESPACE="${1#*=}"
+            shift
+            ;;
+        *)
+            # Pass through other arguments
+            shift
+            ;;
+    esac
+done
+
 # Store PIDs of all the subprocesses
 pids=()
+
+# Check if only loki collection is requested
+if [[ "$LOKI_ONLY" == "true" ]]; then
+    echo "Running Loki-only collection..."
+    # Gather Loki logs from OpenShift Logging (environment variables are already set)
+    /usr/bin/gather_loki
+    echo "Loki collection completed."
+    exit 0
+fi
 
 # Named resource list, eg. ns/openshift-config
 named_resources=()
@@ -82,6 +111,11 @@ all_ns_resources_text=$(IFS=, ; echo "${all_ns_resources[*]}")
 oc adm inspect ${log_collection_args} --dest-dir must-gather --rotated-pod-logs "${all_ns_resources_text}" --all-namespaces &
 pids+=($!)
 
+# Gather Loki logs from OpenShift Logging FIRST (before any other scripts that might fail)
+echo "Starting Loki log collection first..."
+/usr/bin/gather_loki
+echo "Loki log collection completed. Starting other collectors..."
+
 # Gather Insights Operator Archives
 /usr/bin/gather_insights &
 pids+=($!)
@@ -153,6 +187,8 @@ pids+=($!)
 # Gather vSphere resources. This is NOOP on non-vSphere platform.
 /usr/bin/gather_vsphere &
 pids+=($!)
+
+
 
 # Gather Performance profile information
 /usr/bin/gather_ppc &
--- a/collection-scripts/gather_loki
+++ b/collection-scripts/gather_loki
@@ -1,0 +1,237 @@
+#!/bin/bash
+
+source $(dirname "$0")/common.sh
+BASE_COLLECTION_PATH="${BASE_COLLECTION_PATH:-/must-gather}"
+get_log_collection_args
+
+LOKI_COLLECTION_PATH="${BASE_COLLECTION_PATH}/loki-logs"
+
+# Get standard must-gather parameters (like other gather scripts)
+get_log_collection_args
+
+# Use environment variables for our custom parameters
+ONLY_LOKI_MODE="${LOKI_ONLY_MODE:-false}"
+LOGS_NAMESPACE="${LOKI_LOGS_NAMESPACE:-}"
+
+# Batching configuration (configurable via environment variables)
+LOKI_MAX_BATCHES="${LOKI_MAX_BATCHES:-100}"          # Default 100 batches (500K records total)
+
+function check_loki_installed() {
+    if oc get clusterlogging -n openshift-logging >/dev/null 2>&1; then
+        echo "Found ClusterLogging CR"
+        return 0
+    fi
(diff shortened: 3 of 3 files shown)


avlitman:
cc @ingvagabund  @sferich888 will appricate your review (:

README.md
```diff
@@ -35,3 +35,23 @@ oc adm must-gather --image=registry/username/your-built-image
...
+
+### Options
+- `--only-loki` - Collect only Loki logs (skip traditional must-gather)
+- `--logs-namespace=NAME` - Target specific namespace
+- `LOKI_MAX_BATCHES=N` - Batches per log type (default: 100, each batch contain 5000 logs)
+
+### Output Structure
+```
+loki-logs/
+├── application_logs_part1.txt     (250K records)
```
sradco, line 49: Based on what I looked I think we need to download the logs to a NDJson(new line delimited Json) files, that will be easier for external tools to consume.
How hard would is be to convert them to this format.

In addition, I think we should continue downloading and creating files and generate a gz file to compress them, so that at the end we get 1Gb of data.

avlitman:
> @avlitman I don't think we can enable loki collection by default. What would be the average collected size per node/pod? We can accept changes that adds extra few KBs (sometimes few MBs) per cluster. Though, many MBs is too much.

@ingvagabund At the moment it's max to 0.5gb

README.md
avlitman: @sradco 
You mean by default remove all other must gather script if loki stack exist? 
we can use --only-loki option for that and by default collect only few MB..

about the json - also in txt files you can use grep but I can try to convert this no problem.

sradco: @avlitman  Must Gather doesnt collect only Logs.
I mean to remove only the logs collection.
@ingvagabund  would that make sense to you?

oarribas:
@avlitman , I can see this has been requested in CNV-59376. If this request is specific for CNV, I don't think it should be included in the "default" must-gather, but in the CNV one, as I don't think it is something usually required for the "default" must-gather collection, and it can increase the time and size of the collection.

README.md
sradco: @ ingvagabund What is the format current logs are collect in?

sradco:
@oarribas This is not a CNV specific PR. Collecting log for both existing resources and the ones that were already deleted would be very valuable for all OCP users. Many times the issue cant bee seen or investigated since the resource not longer exist in the cluster.

oarribas:
@sradco , it is not always needed to have the logs for deleted resources (and to be honest, from my experience, only "few" times). And in the cases it is required, I think it is better to ask for those specific logs (and only those logs). Collecting everything only for few cases is not the purpose of must-gather, and when specific data is needed (and specially if it's a big amount of data, or very slow to collect), it is better to ask for it apart of the must-gather.

Also, if I understand the PR correctly, it is collecting latest X logs:

> the script return logs by batch per log type, each batch is 5000 logs. the default is 100 batches (0.5 mil logs per log type) and each 250k logs are in a separate file.

If there is any application flooding Loki, there could be not useful logs at all recollected by this PR.

> At the moment it's max to 0.5gb

How is that calculated? If what is collected is a number of log lines, there are log lines bigger than 1MB, so depending of the number of those, the size could be bigger than that 0.5gb.


On the other hand, note that a must-gather doesn't collect logs and other data from custom namespaces in the cluster, and collecting logs that way could collect logs from any custom application.

sradco:
@r2d2rnd, If OpenShift Logging is our supported logging stack it should include the relevant logs that are needed for troubleshooting.
I don't understand why would it make sense to collect only local logs over collecting them from a central logging store.

>1.It's said that if it's collected the logs from Loki it can be deleted the logs collected from the pods
> This is not correct.
> The logs collected now for the pods with a normal must-gather, they are all the logs generated by the pod/container at  least that the log was rotated by crio. Then, these logs can be from days or months.
> The logs obtained from Loki will be requested from 1 hour - 2 hours? Then, they are not the same logs, the requested to  Loki will be only a small part of the available in the must-gather

Are you saying that logs for recently deleted pods are still collected today with must-gather?
@vladikr can you please comment about what your expirience?

BTW, We want to collect 1 day of logs or up or limit by size, not 1-2 hours, which I agree is insufficient.

If logs that we collect to day with must gather are not also gathered to the logging stack, I see this as an issue that should be addressed by the logging team.

> 2.Nothing ensure that the logs in Loki are all the produced in the system
This is a consequence that logs can be pruned, filtered, rejected to be injested with 500 error or 429 errors, or even never read as the collector has a limitation in the number of logs that able to read by second and this limitation in the Logging stack is the reason that we are clear that the Logging stack doesn't guarantee log sink

If this happens frequently I consider it a bug. The Logging stack should provide this functionality reliably.
@jcantrill do you agree with this?

> We limitate the logs and metrics collected to be sure that we treat the data with the minimum and security principles.

How do you do that today. We can work to address that as well.

> 4.The estimated size for the logs is probably not correct

We can check this when we have create the files.

> 5.Performance issues

We can always fallback to local logs after a specific time period

> As recommendation for the Logging issues, we recommend always to only ask the sorter time as it could be 1 hour, 2 hours and trying to indicate the specific namespace/pod/container. This PR goes against the request that we do to the OpenShift Admin for how to use the Red Hat Loki.

@avlitman did you already add the support to collect specific namespaces? Can you pass the MG parameters when you collect the data?

> At the same time, as Loki needs to get the data from an object storage bucket, the network is also impacted.

All the data collection for MG impacts the network. I dont think this is special. 
@jcantrill do you agree?

> 6.Size of the query?
With the description here and the estimation of the number of batches + streams for each batch, I'm not sure how it's possible to know that you'll get all the logs that you need for doing the troubleshooting as it's pretty common that a normal cluster with low log production can produce between 100 - 200GB per day compressed, the compressed factor in Loki is between 4x and 7x. A normal cluster can produce easily 500GB and with medium-high volume exceed the 1TB per day reaching sometimes 2TB (this includes application logs).

I was not familiar with these numbers, how do you handle the logs collection in that case today? How is that different? We do aim to limit the size of the total log files we collect.

> I could continue with the list, and even when I understand the request and I would like the same if I was an Openshift Admin, I feel that the must-gather should continue as it's now for all the details above and if it was needed in one specific situation to obtain logs from a deleted resource, in that specific situation, ask if a Log store exists (loki or other solution used) and ask for the specific logs needed.

I dont think this is a one customer issue and this aims to address a broad use case. The logging stack is supposed to be what users should use to troubleshoot, so why dont we use it as well when possible for our SREs? Its should provide a more complete picture.

oarribas:
@sradco ,

> If OpenShift Logging is our supported logging stack it should include the relevant logs that are needed for troubleshooting.

That OpenShift Loggin is the supported logging stack doesn't mean all customers are using it. There are also customers using only the LogForwarding feature to send logs to external Log stores and not Loki.

>  I don't understand why would it make sense to collect only local logs over collecting them from a central logging store.

Not all customers uses Loki from OpenShift Logging as log store. Also, collecting logs from Loki will cause a high impact in Loki and even the network as already explained above.

> Are you saying that logs for recently deleted pods are still collected today with must-gather?

Not if the resources are deleted.
And as commented before, in most of the cases, that is not really needed. If only needed in very few cases, they should be requested only for those few cases.

> BTW, We want to collect 1 day of logs or up or limit by size, not 1-2 hours, which I agree is insufficient.

1 day of logs could be several TB for many clusters, depending on lots of things, specially custom applications behavior.

> If logs that we collect to day with must gather are not also gathered to the logging stack, I see this as an issue that should be addressed by the logging team.

As commented before, the logs collected by must-gather by default are enough in most of the cases, and that is the purpose of the must-gather, not collecting everything.
For specific cases, it is needed to request specific things.

> If this happens frequently I consider it a bug. The Logging stack should provide this functionality reliably.

As it is explained in the Loggin docs [1]:
>>Logging for Red Hat OpenShift is an opinionated collector and normalizer of application, infrastructure, and audit logs. It is intended to be used for forwarding logs to various supported systems. 

Check also the "*Logging is not*" in the same document.

> I was not familiar with these numbers, how do you handle the logs collection in that case today? How is that different? We do aim to limit the size of the total log files we collect.

It is not the same collect and store the logs as they are produced, than while doing that, retrieve a very big amount of logs again from the Log Storage. That is needed in very few cases, and should be requested only in those few cases. So not something for the standard must-gather.

> I dont think this is a one customer issue and this aims to address a broad use case. The logging stack is supposed to be what users should use to troubleshoot, so why dont we use it as well when possible for our SREs? Its should provide a more complete picture.

As said before, that is only needed in very few cases. And for only those few cases, it should be enough to request that data as needed, and not in the standard must-gather.

> Maybe use the namespace supplied by the user when it runs the Must Gather. From what I understand the recommendation is to run MG for the relevant namespace. So if it is supplied, you can use it also here.

No, the standard must-gather doesn't run against a namespace.


Note also that a lot of the logs collected with this feature will be duplicated for all the non-deleted resources already collected by the standard must-gather. Also, standard must-gather doesn't collect customer's custom namespaces, which could be required for your use case (and will be collected if no namespace is provided to your command).


[1] https://docs.redhat.com/en/documentation/openshift_container_platform/4.16/html/logging/support

oarribas:
>Are you saying that logs for recently deleted pods are still collected today with must-gather?
>@vladikr can you please comment about what your expirience?

@sradco , @vladikr , could you please clarify the exact logs (from which deleted namespace(s)/pod(s)) you needed and were not in the must-gather?

PR #524, opened 2026-02-17
no-jira: chore: ensure scripts are properly formatted

vprashar2929:
This commit addresses the formatting issues in the must gather collection scripts. It only changes the formatting of the scripts and does not change the functionality of the scripts.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -33,13 +33,16 @@ $(call build-image,ocp-must-gather,$(IMAGE_REGISTRY)/ocp/4.17:ocp-must-gather, .
 
 $(call verify-golang-versions,Dockerfile.ocp)
 
+# Directories containing shell scripts to lint and format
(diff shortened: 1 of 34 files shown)


oarribas:
@ingvagabund , @vprashar2929 , could this PR be backported to previous versions to allow backporting other new features to the must-gather collection (like #527)? As it was discussed in https://github.com/openshift/must-gather/pull/453#issuecomment-2462951443 and also in previous discussions, as the must-gather is a diagnostics tool for troubleshooting, some backports are allowed.

ingvagabund:
I advise against. The goal of this PR are improvements. Which we generally do not backport unless there's a meaningful reason for it. Yet, if your goal is to ease backport of fixes I suggest to port only selected changes from this PR.

PR #531, opened 2026-03-10
OCPBUGS-66983: Fix race condition in gather_core_dumps pod name retrieval

shivprakashmuley:
As per the discussion with @anuragthehatter, raising this PR with his original changes in https://github.com/openshift/must-gather/pull/517.

Current logic found failed on one of prow CI run where node indeed had a core dump present but dump was not collected by oc adm must-gather -- gather_core_dumps command underneath a prow chain. It seems we may have missed core dumps collection since long time due to this race issue.

Failed logs on prod builds [here](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/openshift_release/72130/rehearse-72130-pull-ci-openshift-ovn-kubernetes-release-4.15-e2e-aws-ovn-fdp-qe/1996418222236110848/artifacts/e2e-aws-ovn-fdp-qe/gather-core-dump/build-log.txt)

Saw a race condition where dump was tried to be copied but debug pod was already removed.

### DIFF
--- a/collection-scripts/gather_core_dumps
+++ b/collection-scripts/gather_core_dumps
@@ -7,22 +7,44 @@ mkdir -p "${CORE_DUMP_PATH}"/
 function get_dump_off_node {
 	local debugPod=""
 
-	#Get debug pod's name
-	debugPod=$(oc debug --to-namespace="default" node/"$1" -o jsonpath='{.metadata.name}')
+	#Start Debug pod in background and capture output to get pod name
+	local tmpfile=$(mktemp)
+	oc debug --to-namespace="default" node/"$1" -- /bin/bash -c 'sleep 300' > "$tmpfile" 2>&1 &
 
-	#Start Debug pod force it to stay up until removed in "default" namespace
-	oc debug --to-namespace="default" node/"$1" -- /bin/bash -c 'sleep 300' >/dev/null 2>&1 &
+	#Wait for the debug pod to be created and extract its name with exponential backoff
+	local max_attempts=10  # Fewer attempts needed with exponential backoff
+	local attempt=0
+	local base_delay=0.1  # Starting delay in seconds
+	local max_delay=2.0   # Cap the maximum delay
 
-	#Mimic a normal oc call, i.e pause between two successive calls to allow pod to register
-	sleep 2
-	oc wait -n "default" --for=condition=Ready pod/"$debugPod" --timeout=30s
+	# Initial delay to allow pod creation to start
+	sleep 0.5
+
+	while [ -z "$debugPod" ] && [ $attempt -lt $max_attempts ]; do
+		debugPod=$(sed -n 's/.*pod\/\([^ ]*\).*/\1/p' "$tmpfile" 2>/dev/null | head -1)
+		if [ -z "$debugPod" ]; then
+			# Calculate exponential backoff: base_delay * 2^attempt
+			local delay=$(awk -v base="$base_delay" -v exponent="$attempt" -v max="$max_delay" \
+				'BEGIN {d = base * (2 ^ exponent); print (d > max) ? max : d}')
+			sleep "$delay"
+			attempt=$((attempt + 1))
+		fi
+	done
+	rm -f "$tmpfile"
+
+	#Wait for pod to be ready
+	if [ -n "$debugPod" ]; then
+		oc wait -n "default" --for=condition=Ready pod/"$debugPod" --timeout=30s > /dev/null 2>&1
+	fi
 
 	if [ -z "$debugPod" ]; then
 		echo "Debug pod for node ""$1"" never activated"
 	else
 		#Copy Core Dumps out of Nodes suppress Stdout
 		echo "Copying core dumps on node ""$1"""
-		oc cp --loglevel 1 -n "default" "$debugPod":/host/var/lib/systemd/coredump "${CORE_DUMP_PATH}"/"$1"_core_dump >/dev/null 2>&1 && PIDS+=($!)
+		if ! oc cp  --loglevel 1 -n "default" "$debugPod":/host/var/lib/systemd/coredump "${CORE_DUMP_PATH}"/"$1"_core_dump > /dev/null 2>&1; then
+			echo "Warning: Failed to copy core dumps from node $1"
+		fi
 
 		#clean up debug pod after we are done using them
 		oc delete pod "$debugPod" -n "default"
@@ -33,6 +55,7 @@ function gather_core_dump_data {
 	#Run coredump pull function on all nodes in parallel
 	for NODE in ${NODES}; do
 		get_dump_off_node "${NODE}" &
+		PIDS+=($!)
 	done
 }
 
--- a/collection-scripts/gather_core_dumps
+++ b/collection-scripts/gather_core_dumps
@@ -5,11 +5,17 @@ CORE_DUMP_PATH=${OUT:-"${BASE_COLLECTION_PATH}/node_core_dumps"}
 mkdir -p "${CORE_DUMP_PATH}"/
 
 function get_dump_off_node {
+	local node="$1"
 	local debugPod=""
+	local oc_debug_pid=""
+	local tmpfile
 
-	#Start Debug pod in background and capture output to get pod name
-	local tmpfile=$(mktemp)
-	oc debug --to-namespace="default" node/"$1" -- /bin/bash -c 'sleep 300' > "$tmpfile" 2>&1 &
+	tmpfile=$(mktemp)
+	trap 'rm -f "$tmpfile"' RETURN
+
+	# Start Debug pod in background and capture output to get pod name
+	oc debug --to-namespace="default" node/"$node" -- /bin/bash -c 'sleep 300' > "$tmpfile" 2>&1 &
+	oc_debug_pid=$!
 
 	#Wait for the debug pod to be created and extract its name with exponential backoff
 	local max_attempts=10  # Fewer attempts needed with exponential backoff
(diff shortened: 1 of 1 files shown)


sferich888 (changes requested):
Thanks for tackling this @shivprakashmuley — the core approach of combining the two `oc debug` calls and polling with exponential backoff is solid. A few things need to be addressed before we can merge:

1. **Background `oc debug` PID not captured or cleaned up** — As @coderabbitai [flagged](https://github.com/openshift/must-gather/pull/531/files#r2912066013), the `oc debug ... &` starts a background process but `$!` is never saved. If the pod name is never found (all retries exhausted), that background process is orphaned. Capture `debug_pid=$!` immediately after the background call and `kill`/`wait` it on failure paths.

2. **`oc wait` failure not handled** — Also [flagged by CodeRabbit](https://github.com/openshift/must-gather/pull/531/files#r2912066024). If `oc wait --for=condition=Ready` times out, the script still proceeds to `oc cp` because the `if [ -n "$debugPod" ]` check passes (the pod name was found, it just isn't ready). Check the exit code of `oc wait` and skip the copy / clean up the pod if it fails.

3. **`tmpfile` cleanup on early exit** — If the function exits early or the script is killed between `mktemp` and `rm -f "$tmpfile"`, the temp file leaks. Consider adding a `trap` at the top of the function to ensure cleanup, e.g.:
   ```bash
   trap 'rm -f "$tmpfile"' RETURN
   ```
   This ensures the temp file is removed regardless of how the function exits.

Please address these and we should be good to go. The rest of the change looks great — the exponential backoff, graceful `oc cp` warning, and PID tracking in `gather_core_dump_data` are all improvements.

sferich888 (changes requested):

collection-scripts/gather_core_dumps
```diff
@@ -7,22 +7,44 @@ mkdir -p "${CORE_DUMP_PATH}"/
...
+		debugPod=$(sed -n 's/.*pod\/\([^ ]*\).*/\1/p' "$tmpfile" 2>/dev/null | head -1)
+		if [ -z "$debugPod" ]; then
+			# Calculate exponential backoff: base_delay * 2^attempt
+			local delay=$(awk -v base="$base_delay" -v exponent="$attempt" -v max="$max_delay" \
+				'BEGIN {d = base * (2 ^ exponent); print (d > max) ? max : d}')
+			sleep "$delay"
+			attempt=$((attempt + 1))
+		fi
+	done
+	rm -f "$tmpfile"
```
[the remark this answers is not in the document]
shivprakashmuley: Addressed.

shivprakashmuley:
Thanks @sferich888 for the comments. I have pushed the relevant changes. Please have a re-look.

shivprakashmuley:
Hi @sferich888, can you please review the latest changes.

shivprakashmuley:
Hi @sferich888 , @ingvagabund can you please review and approve if everything looks good.

collection-scripts/gather_core_dumps
```diff
@@ -5,34 +5,69 @@ CORE_DUMP_PATH=${OUT:-"${BASE_COLLECTION_PATH}/node_core_dumps"}
...
 
-	#Get debug pod's name
-	debugPod=$(oc debug --to-namespace="default" node/"$1" -o jsonpath='{.metadata.name}')
+	tmpfile=$(mktemp)
+	trap 'rm -f "$tmpfile"' RETURN
 
-	#Start Debug pod force it to stay up until removed in "default" namespace
-	oc debug --to-namespace="default" node/"$1" -- /bin/bash -c 'sleep 300' >/dev/null 2>&1 &
+	# Start Debug pod in background and capture output to get pod name
+	oc debug --to-namespace="default" node/"$node" -- /bin/bash -c 'sleep 300' > "$tmpfile" 2>&1 &
```
sferich888, line 17: We probably don't want this in the default namespace! We like want it in the must/hater name space that way when it's deleted these pods are cleaned up.

collection-scripts/gather_core_dumps
```diff
@@ -5,34 +5,69 @@ CORE_DUMP_PATH=${OUT:-"${BASE_COLLECTION_PATH}/node_core_dumps"}
...
 	if [ -z "$debugPod" ]; then
-		echo "Debug pod for node ""$1"" never activated"
-	else
-		#Copy Core Dumps out of Nodes suppress Stdout
-		echo "Copying core dumps on node ""$1"""
-		oc cp --loglevel 1 -n "default" "$debugPod":/host/var/lib/systemd/coredump "${CORE_DUMP_PATH}"/"$1"_core_dump >/dev/null 2>&1 && PIDS+=($!)
-
-		#clean up debug pod after we are done using them
-		oc delete pod "$debugPod" -n "default"
+		kill "${oc_debug_pid}" 2>/dev/null
```
sferich888, line 42: Why move away from deleting the pod?

sferich888 (commented):
Reviewing from my phone but some of these changes seem odd or may simple need to be explained

collection-scripts/gather_core_dumps
shivprakashmuley: Pod gets deleted here https://github.com/openshift/must-gather/pull/531/changes#diff-f7dfd1af20867af285734a2bf99bf98c1766f2a1a74074d45a75f28ee0ef3a82R61 @sferich888 .

collection-scripts/gather_core_dumps
```diff
@@ -5,34 +5,69 @@ CORE_DUMP_PATH=${OUT:-"${BASE_COLLECTION_PATH}/node_core_dumps"}
...
+	local max_attempts=10  # Fewer attempts needed with exponential backoff
+	local attempt=0
+	local base_delay=0.1  # Starting delay in seconds
+	local max_delay=2.0   # Cap the maximum delay
+
+	# Initial delay to allow pod creation to start
+	sleep 0.5
+
+	while [ -z "$debugPod" ] && [ $attempt -lt $max_attempts ]; do
+		debugPod=$(sed -n 's/.*pod\/\([^ ]*\).*/\1/p' "$tmpfile" 2>/dev/null | head -1)
```
ingvagabund, line 30: I would not rely on whatever is printed in the logs as something referential. I don't think anyone will consider `"Starting pod/xyz-debug-abc..."` string as a part of the API.

ingvagabund, line 30: Instead of reading the pod name from the logs would it help to fetch the corresponding pod name via `oc get pod` while setting the right label selector (`debug.openshift.io/managed-by=oc-debug`) and checking for the right node name?

If this path does not work well (e.g. when two or more pods are still listed) `oc debug nodes/...` command could be extended with an extra option for adding custom labels.

shivprakashmuley: I have tried to check if we can leverage labels somehow to identify the corresponding debug pod

`$ oc get po smuley-20260407-520b9-879rp-master-0-debug-7rd2p -n default -o yaml
apiVersion: v1
kind: Pod
metadata:
  annotations:
    debug.openshift.io/source-container: container-00
    debug.openshift.io/source-resource: /v1, Resource=nodes/smuley-20260407-520b9-879rp-master-0
    openshift.io/required-scc: privileged
  creationTimestamp: "2026-04-10T07:44:50Z"
  generation: 1
  labels:
    debug.openshift.io/managed-by: oc-debug
  name: smuley-20260407-520b9-879rp-master-0-debug-7rd2p
  namespace: default
  resourceVersion: "1003576"
  uid: 3f7d2587-4025-4c6a-8f2f-d64f5898af55
spec:
  containers:
  - command:
    - /bin/bash
    - -c
    - sleep 300
    env:
    - name: TMOUT
      value: "900"
    - name: HOST
      value: /host
    image: quay.io/openshift-release-dev/ocp-v4.0-art-dev@sha256:1bf238da7a1100201223f287894d6cb90d7ac5eb6bcd3311382bcc6774cae8c0
    imagePullPolicy: IfNotPresent
  imagePullSecrets:
  - name: default-dockercfg-s4xcc
  nodeName: smuley-20260407-520b9-879rp-master-0`

Thought we can go with `Label + spec.nodeName`  but in case of 2 runs of `gather_core_dumps` on the same node - we won't be able to identify the pod corresponding to a particular run.

And from my understanding, we can not inject a custom label in `oc debug` command.

Thoughts? @ingvagabund .

ingvagabund: Would you have time to investigate the possibility of extending `oc debug nodes` with a new option for a list of custom labels?

shivprakashmuley: > Would you have time to investigate the possibility of extending `oc debug nodes` with a new option for a list of custom labels?

Hi @ingvagabund , I can try to investigate. That would require another story/ticket to track right? 
Can you please suggest on the process usually followed for `oc` changes. Do we need a enhancement proposal for the `oc debug nodes` changes? Also this means it will add more time in resolving https://github.com/openshift/must-gather/pull/531 please share your thoughts on if we are ok with that.

ingvagabund: Thank you. 

> That would require another story/ticket to track right?

That's up to you how you track the effort.

> Can you please suggest on the process usually followed for oc changes. Do we need a enhancement proposal for the oc debug nodes changes?

Given this is only about adding a new flag localized to oc debug node, no enhancement is needed. Only providing a meaningful description when opening a PR under openshift/oc.

> Also this means it will add more time in resolving https://github.com/openshift/must-gather/pull/531 please share your thoughts on if we are ok with that.

Not sure who "we" is in this context :). Also, it's up to you or your team how much time you have allocated for resolving the issue. I am not aware of any time constraints here.

shivprakashmuley: cc: @Prashanth684

PR #535, opened 2026-03-31
no-jira: build(make): default podman flags for ART DNF wrapper

simonstumpf:
Local podman image builds were failing because the build expected dependencies only present in CI. 
Set `IMAGE_BUILD_DEFAULT_FLAGS` so podman passes `--env ART_DNF_WRAPPER_POLICY=skip` in the Makefile. When AUTH_FILE is set, --authfile is appended with `+=` so it does not override that default.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -14,11 +14,12 @@ ifeq ($(BUILDER), podman)
 # Use podman to build the must gather image
 # Please ensure you have podman installed on your machine for this
 IMAGE_BUILD_BUILDER := podman build
-# Clear all build default flags
-IMAGE_BUILD_DEFAULT_FLAGS :=
+# ART_DNF_WRAPPER_POLICY=skip: skip the ART image-build DNF wrapper so local podman
+# builds use ordinary dnf; the wrapper applies in CI/product pipelines, not here.
+IMAGE_BUILD_DEFAULT_FLAGS := --env ART_DNF_WRAPPER_POLICY=skip
 # Set authfile if the user passes another location for the authentication file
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -14,11 +14,12 @@ ifeq ($(BUILDER), podman)
 # Use podman to build the must gather image
 # Please ensure you have podman installed on your machine for this
 IMAGE_BUILD_BUILDER := podman build
-# Clear all build default flags
-IMAGE_BUILD_DEFAULT_FLAGS :=
+# ART_DNF_WRAPPER_POLICY=skip: skip the ART image-build DNF wrapper so local podman
+# builds use ordinary dnf; the wrapper applies in CI/product pipelines, not here.
+IMAGE_BUILD_DEFAULT_FLAGS := --env ART_DNF_WRAPPER_POLICY=skip
```
ingvagabund, line 19: What about the case when the make is running within a CI? Will not this break the CI?

simonstumpf: In my point of view this only applies when BUILDER=podman. Default make image uses imagebuilder, so the new IMAGE_BUILD_DEFAULT_FLAGS are not used in the normal CI path. I’m still quite new to the project, so please double-check my logic.

joepvd: This is not a concern for anything that is building straight from Dockerfile, meaning: Production builds and CI builds. IMHO, this is safe.

ingvagabund:
/retitle no-jira: build(make): default podman flags for ART DNF wrapper

PR #562, opened 2026-09-09
OCPBUGS-122049: gather_ppc: collect per-node ethtool channel data

mrniranjan:
Earlier we used to collect ethtool data for only eth0 now extend this to collect data from all the hardware nics

AIA: Primarily AI, New content, Human-initiated, Reviewed, Codex GPT-5.6

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved per-node PPC diagnostics by collecting Ethernet feature and channel information from all hardware-backed network interfaces, rather than only a fixed interface.
  * Collector pods now use the host network, enabling more accurate inspection of the node’s physical network interfaces.
  * Diagnostics now identify the TuneD service running on each target node for more accurate per-node data.
  * Added a clear fallback message when no hardware-backed network interfaces are available.


MarSik:
I see one issue in the logic. Maybe it is theoretical, but I wonder if it is wise to depend on existing running containers to exec into and run commands. It seems dirty and unreliable to me.

Instead, why can't you just use the already mounted /host/sys in the daemonset to find the NICs?

MarSik (commented):
/lgtm

